PR #3, opened 2018-07-04
Main documents

Sinclert:
I have created the same common files as any other REANA module.

Some of them are still empty (_userguide.rst_, _developerguide.rst_, and some other) because there is still no code to reference.

Additionally, I have name 'reana-ui' everything that in other modules is called 'reana-server' or 'reana-cluster' even if it does not exist yet (example: travis).

Finally, I have added me to the list of contributors, in alphabetical order.

**VIP: I have consider that the version would start at the 0.3.0, but if it shouldn't let me know.**


docs/requirements.txt
```diff
@@ -0,0 +1,21 @@
...
+#
+# You should have received a copy of the GNU General Public License along with
+# REANA; if not, write to the Free Software Foundation, Inc., 59 Temple Place,
+# Suite 330, Boston, MA 02111-1307, USA.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization or
+# submit itself to any jurisdiction.
+
+-e .[all]
```
diegodelemos: Since we don't have `setup.py` we can remove `-e .[all]`and we can add documentation requirements here, `Sphinx>=1.7.5` and `sphinx-rtd-theme >=0.4.0`. Also, in `run-tests.sh` we can build the docs, see for example [workflow-controller](https://github.com/reanahub/reana-workflow-controller/blob/master/run-tests.sh), so it would fail if there is any error in the docs. You could also plug Travis CI, see [workflow-controller .travis.yml](https://github.com/reanahub/reana-workflow-controller/blob/master/.travis.yml) (things like docker and sudo true are not needed for this package, and `pip install -e .[all]` becomes `pip install -r requirements.txt`).

requirements-dev.txt
```diff
@@ -0,0 +1,23 @@
...
+#
+# You should have received a copy of the GNU General Public License along with
+# REANA; if not, write to the Free Software Foundation, Inc., 59 Temple Place,
+# Suite 330, Boston, MA 02111-1307, USA.
+#
+# In applying this license, CERN does not waive the privileges and immunities
+# granted to it by virtue of its status as an Intergovernmental Organization or
+# submit itself to any jurisdiction.
+
+-e git+git://github.com/reanahub/reana-commons.git#egg=reana-commons
```
diegodelemos: I think we don't need this file here :)

docs/conf.py
```diff
@@ -0,0 +1,213 @@
...
+#
+# source_suffix = ['.rst', '.md']
+source_suffix = '.rst'
+
+# The master toctree document.
+master_doc = 'index'
+
+# General information about the project.
+project = 'reana'
+copyright = '018, info@reana.io'
```
diegodelemos: 2018

docs/conf.py
```diff
@@ -0,0 +1,213 @@
...
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+
+# Get the version string. Cannot be done with import!
+g = {}
+with open(os.path.join('..', 'reana-ui', 'package.json'), 'rt') as fp:
+    data = json.load(fp)
+    version = data["version"]
```
diegodelemos: 👍

docs/conf.py
Sinclert: Done ✅

PR #9, opened 2018-07-09
Login grommet

Sinclert:
Work in progress:

- Login page.
- Common components such as Header.


Sinclert:
Pull request closed because we are choosing a different UI framework instead of Grommet.

Framework decision: [Semantic-UI](https://semantic-ui.com)

Issue #10, opened 2018-07-19
implementation: workflow-list


Sinclert:
Starting workflow-list implementation

Sinclert:
We thought it was interesting to introduced a new column (not showed in the mockups), which is the `duration` of the executions. I was thinking about following the format `HH h MM m SS s`.

Example: 
![screen shot 2018-07-24 at 13 56 12](https://user-images.githubusercontent.com/6978377/43136540-595f832e-8f49-11e8-859e-6eae302c72e0.png)

dinosk:
:+1: the duration column looks great

PR #11, opened 2018-07-19
Workflow list

Sinclert:
First steps of the workflow-list implementation
Connect #10

### DIFF
--- a/reana-ui/public/index.html
+++ b/reana-ui/public/index.html
@@ -1,0 +1,40 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+    <meta name="theme-color" content="#000000">
+    <!--
+      manifest.json provides metadata used when your web app is added to the
+      homescreen on Android. See https://developers.google.com/web/fundamentals/engage-and-retain/web-app-manifest/
+    -->
+    <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
+    <link rel="shortcut icon" href="%PUBLIC_URL%/favicon.ico">
+    <!--
+      Notice the use of %PUBLIC_URL% in the tags above.
+      It will be replaced with the URL of the `public` folder during the build.
+      Only files inside the `public` folder can be referenced from the HTML.
+
+      Unlike "/favicon.ico" or "favicon.ico", "%PUBLIC_URL%/favicon.ico" will
+      work correctly both with client-side routing and a non-root public URL.
+      Learn how to configure a non-root public URL by running `npm run build`.
+    -->
+    <title>REANA</title>
+  </head>
+  <body>
+    <noscript>
+      You need to enable JavaScript to run this app.
+    </noscript>
+    <div id="root"></div>
+    <!--
+      This HTML file is a template.
+      If you open it directly in the browser, you will see an empty page.
+
+      You can add webfonts, meta tags, or analytics to this file.
+      The build step will place the bundled scripts into the <body> tag.
+
+      To begin the development, run `npm start` or `yarn start`.
+      To create a production bundle, use `npm run build` or `yarn build`.
+    -->
+  </body>
+</html>
--- a/reana-ui/public/manifest.json
+++ b/reana-ui/public/manifest.json
@@ -1,0 +1,15 @@
+{
+  "short_name": "REANA",
+  "name": "REANA",
+  "icons": [
+    {
+      "src": "favicon.ico",
+      "sizes": "64x64 32x32 24x24 16x16",
+      "type": "image/x-icon"
+    }
+  ],
+  "start_url": "./index.html",
+  "display": "standalone",
+  "theme_color": "#000000",
+  "background_color": "#ffffff"
+}
--- a/reana-ui/src/components/Header.js
+++ b/reana-ui/src/components/Header.js
@@ -1,0 +1,50 @@
+/*
+	-*- coding: utf-8 -*-
+
+	This file is part of REANA.
+	Copyright (C) 2018 CERN.
+
+	REANA is free software; you can redistribute it and/or modify it under the
+	terms of the GNU General Public License as published by the Free Software
+	Foundation; either version 2 of the License, or (at your option) any later
+	version.
+
+	REANA is distributed in the hope that it will be useful, but WITHOUT ANY
+	WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+	A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with REANA; if not, see <http://www.gnu.org/licenses>.
+
+	In applying this license, CERN does not waive the privileges and immunities
+	granted to it by virtue of its status as an Intergovernmental Organization or
+	submit itself to any jurisdiction.
+*/
+
+import React, { Component } from "react";
+import { Segment, Image, Button, Menu, Icon } from "semantic-ui-react";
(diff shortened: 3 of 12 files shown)


diegodelemos:
@reanahub/developers I think we should impose some format rules, I have found https://standardjs.com/ to be really useful after researching a bit about JavaScript formatters. It is not only the absence of configuration but also how easy it is to use from the CLI and integration with editors ([emacs](https://standardjs.com/#emacs), [VS Code](https://standardjs.com/#visual-studio-code)...).

tiborsimko:
What about ESLint? It comes with AirBnB, Google, React styles, so that one could opt for a more popular style (AirBnB over Standard) or even use the same style rules that the React project uses. Here are some pages to read and ponder:

- https://hackernoon.com/what-javascript-code-style-is-the-most-popular-5a3f5bec1f6f
- https://medium.com/@uistephen/style-guides-for-linting-ecmascript-2015-eslint-common-google-airbnb-6c25fd3dff0 
- https://blog.sideci.com/how-the-react-developer-team-uses-eslint-2828564814da
- https://npmcompare.com/compare/eslint-config-airbnb,eslint-config-google,eslint-config-react,standard

ioannistsanaktsidis:
@reanahub/developers in CAP we are using this one https://prettier.io/ . We have set it up the moment you commit the files. It is working really good in case you want to check!

diegodelemos:
I would keep it simple and use `standard` (which we are [already using in Invenio](https://github.com/prettier/prettier/issues/40#issue-199895633) and is close to Python style) with no configuration if possible, only if really necessary (the only thing that I am not personally really happy with in `standard` defaults it is that is does not allow trailing commas, which in my opinion are necessary to simplify diffs). My second pick would be https://prettier.io/

I kind of agree with https://github.com/prettier/prettier/issues/40#issue-199895633.

Sinclert:
I have been testing `standard` and I have notice 2 things:

1. It is not native compatible with ES2017+ 'experimental features'. Even if this sounds exotic, it is something very common to have. For example:
screen shot 2018-07-26 at 14 25 22
It is easy to make it compatible with the new syntax. A package called `babel-eslint` can be installed and added to the parsing option to make it compatible:
`standard --parser babel-eslint`

2. I have found some parsing inconsistencies while parsing the JSX tags inside a `render()` functions. However, it doesn't happen in all the cases.

Sinclert:
I have been testing [Prettier](https://prettier.io/) and I have noticed the following:

1. It has some configuration level over CLI ([source](https://prettier.io/docs/en/options.html)). I was gonna suggest to use the `--use-tabs` flag, for instance.

2. Haven't encounter any problems with neither ES2017+ 'experimental syntax' nor JSX code inside a `render()` function.

So far, my opinion goes closer to [Prettier](https://prettier.io/).

Sinclert:
Finally, I have tested [ESlint](https://eslint.org) with the [AirBnb configuration](https://www.npmjs.com/package/eslint-config-airbnb), encountering the following problems:

1. Exactly as `Standard`, ES2017+ syntax is not natively supported. It can be supported by adding the `experimentalObjectRestSpread` option in the config file ([source](https://eslint.org/docs/2.0.0/user-guide/configuring)).

2. The default Vanilla / AirBnb config present some problems with JSX and React ([commented here](https://github.com/yannickcr/eslint-plugin-react/issues/447). They are not major, but requires to modify and built our own config file.

tiborsimko:
Looks like Prettier may be the most advantageous; let's see IRL tomorrow?

reana-ui/public/index.html
```diff
@@ -0,0 +1,40 @@
...
+    <!--
+      Notice the use of %PUBLIC_URL% in the tags above.
+      It will be replaced with the URL of the `public` folder during the build.
+      Only files inside the `public` folder can be referenced from the HTML.
+
+      Unlike "/favicon.ico" or "favicon.ico", "%PUBLIC_URL%/favicon.ico" will
+      work correctly both with client-side routing and a non-root public URL.
+      Learn how to configure a non-root public URL by running `npm run build`.
+    -->
+    <title>React App</title>
```
diegodelemos: `REANA` instead of `React App`.

reana-ui/public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
+  "short_name": "Reana-UI",
```
diegodelemos: Capital letters REANA here and everywhere else

Sinclert: I guess the name of the package still needs to be `reana-ui`, right?

diegodelemos: yes! you are right https://docs.npmjs.com/files/package.json#name

reana-ui/src/components/Table.js
```diff
@@ -0,0 +1,221 @@
...
+    data: [],
+    direction: null,
+    interval: null
+  };
+
+  /**
+   * Transforms millisecond into a 'HH MM SS' string format
+   */
+  static msToTime(millis) {
+    let seconds = Math.floor((millis / 1000) % 60);
```
diegodelemos: what about using the Date object? something like
```JavaScript
let string_date = "2018-07-30T15:52:44";
let date_obj = new Date(string_date)
let seconds = date_obj.getSeconds()
let minutes = date_obj.getMinutes()
let hours = date_obj.getHours()
```

Sinclert: That way of operating is used to plot the creation date. However, for the `Duration` field, we should compare every date entity (secs, mins, hours, days, months, year).

Because of this reason, I transform the date into a UNIX stamp number, and compute the difference between `Date.now()` and that number.

run-tests.sh
```diff
@@ -21,4 +21,4 @@
 # submit itself to any jurisdiction.
 
 sphinx-build -qnN docs docs/_build/html
-prettier reana-ui/src/**/*.js --list-different
+#prettier reana-ui/src/**/*.js --list-different
```
diegodelemos: Can you uncomment so we see if Travis is happy?

reana-ui/src/registerServiceWorker.js
```diff
@@ -0,0 +1,117 @@
+// In production, we register a service worker to serve assets from local cache.
```
diegodelemos: I would remove this, we are not going to use Service Workers now, let see what happens after first implementation.

reana-ui/src/components/Header.js
```diff
@@ -0,0 +1,50 @@
...
+	along with REANA; if not, see <http://www.gnu.org/licenses>.
+
+	In applying this license, CERN does not waive the privileges and immunities
+	granted to it by virtue of its status as an Intergovernmental Organization or
+	submit itself to any jurisdiction.
+*/
+
+import React, { Component } from "react";
+import { Segment, Image, Button, Menu, Icon } from "semantic-ui-react";
+import LogoImg from "../images/logo.svg";
```
tiborsimko: It may be good to call the logo `logo-reana.svg` as we have in other repositories... Would be more comfortable for finding and updating files.

reana-ui/src/components/Table.js
```diff
@@ -0,0 +1,222 @@
...
+	submit itself to any jurisdiction.
+*/
+
+import React, { Component } from "react";
+import { Icon, Menu, Segment, Table } from "semantic-ui-react";
+import TableProgress from "./TableProgress";
+import TableActions from "./TableActions";
+import _ from "lodash";
+
+const URL = "http://reana-dev.cern.ch/api/workflows?";
```
tiborsimko: Do we need to have this hardcoded? If so, perhaps create `config.js` in the toplevel?

reana-ui/src/registerServiceWorker.js
Sinclert: Do you mean removing the whole file? or just that line?

PR #13, opened 2018-07-24
docs: basic developers guide

Sinclert:
Basic developer guide in order to do a local deployment and test the UI

### DIFF
--- a/docs/developerguide.rst
+++ b/docs/developerguide.rst
@@ -2,3 +2,35 @@
 
 Developer guide
 ===============
+
+This guide explains how to locally-start REANA-UI, with reloading activated:
+
+1. Clone reana-ui Github project into your computer
+
+   .. code-block:: console
+
+      $ git clone https://github.com/reanahub/reana-ui.git
+
+2. Install all the project dependencies:
+
+   .. code-block:: console
+
+      $ cd reana-ui/reana-ui
+      $ npm install
+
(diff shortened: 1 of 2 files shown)


docs/developerguide.rst
```diff
@@ -2,3 +2,40 @@
 
 Developer guide
 ===============
+
+This tutorial explains how to quick-start with REANA-UI.
```
diegodelemos: Nice, super simple guide! What do you think about simplifying even a bit more? Merging this line with the line after the subsection `Deploy locally` to something like: `In this guide we will setup a development instance of the REANA-UI with live reloading.` and then start directly with the list of steps to follow, so we remove first subsection since we will not have a new one soon? Otherwise looks great to me!

Sinclert: Done ✅

docs/developerguide.rst
```diff
@@ -2,3 +2,35 @@
 
 Developer guide
 ===============
+
+This tutorial explains how to locally-start REANA-UI, with automatic refresh:
```
diegodelemos: Could we better use the word `guide` than `tutorial`? It is what we are using everywhere else, plus it is more appropriate. Also after googling a bit, I found live reloading to be the common term when it comes to talk about the automatic refresh, shall we switch 😜 ?

PR #14, opened 2018-07-30
docs: introducing Prettier

Sinclert:
[Prettier](https://prettier.io) has been chosen as the code styler package.

It is included in the "developpers guide" and in the `run-tests.sh`

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -21,3 +21,4 @@
 # submit itself to any jurisdiction.
 
 sphinx-build -qnN docs docs/_build/html
+prettier reana-ui/src/**/*.js --list-different
--- a/docs/developerguide.rst
+++ b/docs/developerguide.rst
@@ -3,6 +3,9 @@
 Developer guide
 ===============
 
+Quick start
+-----------
+
 This guide explains how to locally-start REANA-UI, with reloading activated:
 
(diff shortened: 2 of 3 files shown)


run-tests.sh
```diff
@@ -21,3 +21,4 @@
 # submit itself to any jurisdiction.
 
 sphinx-build -qnN docs docs/_build/html
+prettier **/*.js --list-different
```
diegodelemos: should we add a [`.prettifierignore`](https://prettier.io/docs/en/cli.html#ignore-path) for `docs/*`?
Otherwise:
```console
$ prettier '**/*.js' --list-different
docs/_build/html/_static/doctools.js
docs/_build/html/_static/documentation_options.js
docs/_build/html/_static/jquery-3.2.1.js
docs/_build/html/_static/jquery.js
docs/_build/html/_static/searchtools.js
docs/_build/html/_static/underscore-1.3.1.js
docs/_build/html/_static/underscore.js
docs/_build/html/_static/websupport.js
docs/_build/html/searchindex.js
$ echo $?
1
```

Sinclert: The command is meant to be executed inside `reana-ui/src`. I will make it more clear :)

Sinclert: Solved ✅

PR #18, opened 2018-07-31
Login page

Sinclert:
**NOT** work in progress any more

### DIFF
--- a/reana-ui/src/components/LoginForm.js
+++ b/reana-ui/src/components/LoginForm.js
@@ -1,0 +1,151 @@
+/*
+	-*- coding: utf-8 -*-
+
+	This file is part of REANA.
+	Copyright (C) 2018 CERN.
+
+	REANA is free software; you can redistribute it and/or modify it under the
+	terms of the GNU General Public License as published by the Free Software
+	Foundation; either version 2 of the License, or (at your option) any later
+	version.
+
+	REANA is distributed in the hope that it will be useful, but WITHOUT ANY
+	WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+	A PARTICULAR PURPOSE. See the GNU General Public License for more details.
+
+	You should have received a copy of the GNU General Public License
+	along with REANA; if not, see <http://www.gnu.org/licenses>.
+
+	In applying this license, CERN does not waive the privileges and immunities
+	granted to it by virtue of its status as an Intergovernmental Organization or
+	submit itself to any jurisdiction.
+*/
+
+import axios from "axios";
+import React, { Component } from "react";
+import ReactDOM from "react-dom";
+import {
+  Button,
+  Divider,
+  Form,
+  Grid,
+  Image,
+  Message,
+  Segment
+} from "semantic-ui-react";
+
+import LogoImg from "../images/logo-reana.svg";
+import WorkflowPage from "../pages/Workflows";
+import Config from "../config";
+
+export default class LoginPage extends Component {
+  /**
+   * Variables defining the state of the login
+   */
+  constructor(props) {
+    super(props);
+    this.state = {
+      email: "",
+      token: "",
+      show_message: false
+    };
+  }
+
+  /**
+   * Updates state variables given user input
+   */
+  handleChange = (e, { name, value }) => {
+    this.setState({ [name]: value });
+  };
+
+  /**
+   * Handling submit button
+   */
+  handleSubmit = () => {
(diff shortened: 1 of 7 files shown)


diegodelemos:
Depends on https://github.com/reanahub/reana-server/issues/73.

reana-ui/src/pages/Login.js
```diff
@@ -33,8 +33,32 @@ import {
...
+  };
+
+  handleChange = (e, { name, value }) => {
+    this.setState({ [name]: value });
+  };
+
+  handleSubmit = () => {
+    const { email, token } = this.state;
+
+    fetch(URL + "access_token=" + token)
```
diegodelemos: I think here starts the complication which I make reference in #19. To correctly handle errors we should resolve the promise and check if the status is the one we expect. `fetch` itself it is going to throw and exception only when there are network problems, but not when you for instance, get a 404. See the [docs](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Checking_that_the_fetch_was_successful) about correctly checking if `fetch` works. @reanahub/developers

reana-ui/src/components/WorkflowList.js
```diff
@@ -39,7 +38,8 @@ export default class WorkflowList extends Component {
     column: null,
     data: [],
     direction: null,
-    interval: null
+    interval: null,
+    token: this.props.token
```
ioannistsanaktsidis: Here you need to wrap the state in the constructor and call super(props) otherwise this.props will return undefined. Generally is a good practice to declare the component state inside a constructor with this.state =  {...} and not as a global.

reana-ui/src/components/WorkflowList.js
```diff
@@ -82,11 +82,11 @@ export default class WorkflowList extends Component {
    * Gets data from the specified API
    */
   getData() {
-    fetch(URL + "access_token=" + TOKEN)
+    const { column, direction, token } = this.state;
+
+    fetch(URL + "access_token=" + token)
```
ioannistsanaktsidis: here you can also use template literals, instead of concatenating the vars. Check this one https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

reana-ui/src/pages/Login.js
```diff
@@ -0,0 +1,121 @@
...
+
+import LogoImg from "../images/logo-reana.svg";
+import WorkflowPage from "./Workflows";
+
+const URL = "http://reana-dev.cern.ch/api/users?";
+
+export default class LoginPage extends Component {
+  state = {
+    email: "",
+    token: ""
```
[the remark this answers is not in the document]
Sinclert: Any particular reason? Error handling?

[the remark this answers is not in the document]
ioannistsanaktsidis: Yes exactly, I think it is easier to check this.state.email ? a:b , than string comparison 😄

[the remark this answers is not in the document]
ioannistsanaktsidis: Ok, maybe it is something with semantic-ui-forms that you are using. It  is not big deal, just pointed for matters of consistency. I guess it is ok to keep it as empty string. You can also keep it as null if you want, but you need to do this in the form
```
value={email || ''}
```

reana-ui/src/components/LoginForm.js
```diff
@@ -0,0 +1,150 @@
...
+                <Form.Input
+                  fluid
+                  icon="user"
+                  iconPosition="left"
+                  placeholder="E-mail"
+                  name="email"
+                  value={email}
+                  onChange={this.handleChange}
+                />
+                <Form.Input
```
diegodelemos: We need to format this as password otherwise it is clear text

reana-ui/src/pages/Workflows.js
```diff
@@ -30,7 +30,7 @@ export default class WorkflowsPage extends Component {
     return (
       <div>
         <Header />
-        <WorkflowList />
+        <WorkflowList token={this.props.token} />
```
diegodelemos: Is this way of passing parameters working? I am not getting any workflow in the list after logging in.

Sinclert: Yes, it is the way of doing it. It is currently not working because we required to work with JWT in all the endpoints. So, the way I make it work is by locally changing WorkflowsList requests to access the `/workflows` endpoint in the old manner.

PR #30, opened 2019-08-07
docker: serve production build through nginx

diegodelemos:


Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM node as react-build
+COPY . /code
```
tiborsimko: Would it be possible to pre-fetch-and-cache some dep steps so that image layer caching would be used?  Otherwise each build takes more then four minutes for me even without changing anything in the source code...

Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+WORKDIR /code
+RUN cd reana-ui && \
+    yarn && \
+    yarn build
+
+FROM nginx
+COPY --from=react-build /code/reana-ui/build /usr/share/nginx/html
+EXPOSE 80
```
tiborsimko: Is this already exposed to the local host? I can access the ui web page only after `kubectl describe service` and friends... Thinking of nodeport, clusterip, rafeik and firiends whether they are ready? 

 But I can get there fine :+1: BTW the login page could say instead of "New to us? Ask for a token" something more neutral such as "New user? Request a token"

PR #41, opened 2019-11-07
login: redux state management and login handling

### LINKED ISSUE #36: login: show UI depending of user being logged in or not
Whether the user is logged in or not should be part of the state of the React application. Currently, we just add links to the pages, for example to [`my GitLab projects`](https://github.com/reanahub/reana-ui/blob/12e92dd45ea067eff0bb844b79674fc72a33e6ad/reana-ui/src/pages/login/components/LoginForm.js#L47), so regardless of the user being logged in or not the link is there and they can navigate (of course the REST calls behind the scenes will fail because there is no session cookie) which is confusing.

User interface state should show according to the user status, logged in, anonymous etc..

mvidalgarcia:
Closes #36
Related to https://github.com/reanahub/reana-server/pull/183

##### `TODO:`
- [x] Improve redirects when user is logged in.
- [x] Log out (actions, reducers, etc.)


reana-ui/src/actions.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+    dispatch({ type: USER_RECEIVED, ...data });
+    return resp;
+  };
+}
+
+export function userLogout() {
+  return async dispatch => {
+    dispatch({ type: USER_LOGOUT });
+    window.location.href = USER_LOGOUT_URL;
```
diegodelemos: Are we ok here assuming that there won't be any problem/error while logging out?

reana-ui/src/components/App.js
```diff
@@ -0,0 +1,51 @@
...
+import GitLabProjects from "../pages/gitlabProjects/GitLabProjects";
+
+function ProtectedRoute(props) {
+  const loggedIn = useSelector(isLoggedIn);
+  const { component: Component, render, ...restProps } = props;
+  const renderContent = render ? render() : <Component {...restProps} />;
+  return (
+    <Route
+      {...restProps}
+      render={() => (loggedIn ? renderContent : <Redirect to="/" />)}
```
diegodelemos: It seems that this is always triggering a redirect to `/` for any path but `/projects`.

reana-ui/src/actions.js
mvidalgarcia: Probably not, but it's the only way I found to remove the session cookie.

reana-ui/src/components/App.js
diegodelemos: After checking IRL with @mvidalgarcia, this will be fixed in https://github.com/reanahub/reana-ui/issues/42.

PR #48, opened 2019-12-04
projects: use toggle to indicate integration

### LINKED ISSUE #38: gitlab: show if the GitLab projects are integrated or not
We should show whether the projects are integrated with REANA or not, with for example a toggle button as Travis does, so the projects which have the integration should be shown toggled and the ones which don’t should appear un-toggled.

image

Currently, it looks like this:

Screen Shot 2019-08-16 at 1 47 27 PM

mvidalgarcia:
closes #38 
depends on https://github.com/reanahub/reana-server/pull/189

#### `TODO`
- [x] Handle webhook deletion
- [x] Make UI sexier 💅 

Current look:
image

### DIFF
--- a/reana-ui/src/pages/gitlabProjects/GitLabProjects.js
+++ b/reana-ui/src/pages/gitlabProjects/GitLabProjects.js
@@ -8,6 +8,7 @@
   under the terms of the MIT License; see LICENSE file for more details.
 */
 
+import _ from "lodash";
 import React, { useEffect, useState } from "react";
 import {
   Button,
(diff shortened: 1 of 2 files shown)


reana-ui/src/pages/gitlabProjects/GitLabProjects.js
```diff
@@ -91,7 +104,7 @@ export default function GitLabProjects() {
     return (
       <div>
         <TopHeader />
-        <Container text className="gitlab-container">
+        <Container text className="container gitlab-container">
```
diegodelemos: `gitlab-container` class is deleted, right? (note also `no-projects-container` at the end of the file)

mvidalgarcia: Indeed, it's a leftover from previous changes, thanks! 👍

PR #54, opened 2019-12-19
release: v0.6.0

mvidalgarcia:


CHANGES.rst
```diff
@@ -1,6 +1,17 @@
...
 =======
 
+Version 0.6.0 (2019-12-19)
+--------------------------
+
+- Adds GitLab projects integration.
+- Adds CERN SSO integration and login/logout handling.
+- Adds state management with Redux.
+- Improves UX in projects page.
+- Allows enabling/disabling GitLab project integration easily.
```
roksys: What do you think about ``Allows enabling/disabling GitLab project integration.``? Is there a need to put emphasis on `easily`?

CHANGES.rst
```diff
@@ -1,6 +1,17 @@
...
 
+Version 0.6.0 (2019-12-19)
+--------------------------
+
+- Adds GitLab projects integration.
+- Adds CERN SSO integration and login/logout handling.
+- Adds state management with Redux.
+- Improves UX in projects page.
+- Allows enabling/disabling GitLab project integration.
+- Includes SASS and CSS-modules support.
```
tiborsimko: Can you group items roughly as follows:

- first basic login/user page i.e. SSO login plus "temporary token page"  which could be better said "simple user page showing access token" for example 
- then all GitLab ones together related to integration and project management
- end with all technology items, i.e. sass and css modules

CHANGES.rst
```diff
@@ -1,6 +1,17 @@
 Changes
 =======
 
+Version 0.6.0 (2019-12-19)
```
tiborsimko: 2019-12-20

docs/conf.py
```diff
@@ -50,7 +50,7 @@
 
 # General information about the project.
 project = 'reana'
-copyright = '2018, info@reana.io'
+copyright = '2019, info@reana.io'
```
tiborsimko: We had this alpha working in 2018 too, so 2018-2019 is better

reana-ui/src/config.js
```diff
@@ -10,9 +10,8 @@
 
 const config = {
   name: "CERN REANA UI",
-  sign_up_email: "info@reana.io",
-  api: "http://reana-dev.cern.ch",
-  pooling_secs: 5,
+  api: "https://reana-dev.cern.ch",
```
tiborsimko: Hardcoding reana-dev ?

diegodelemos: I guess we will just leave `reana.cern.ch` for now? (UI only works at CERN for now)

In a more general aspect, this is a problem we need to solve for the future because this is something we can only change before build time, so if one sees REANA-UI on Docker-hub there are two identical tags one for DEV and another for QA which only differ on this field.

tiborsimko: Yeah, we can leave it easy for now, and the PROD seems better than DEV regarding outside effect.

BTW please also change the waffle bit, see `reana-commons` for example:

```diff
diff --git a/CONTRIBUTING.rst b/CONTRIBUTING.rst
index 2b45294..2dfa681 100644
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -28,6 +28,6 @@ Each pull request should preserve or increase code coverage.
 Kanban
 ------
 
-We are using Kanban technique for keeping track of ongoing tasks. Please see our
-`Kanban board <https://waffle.io/reanahub/reana>`_ and look for issues that are
-labelled as "ready for work".
+We are using Kanban technique for keeping track of ongoing tasks. Please see
+our `Kanban boards <https://github.com/orgs/reanahub/projects>`_ and look for
+issues that are "ready for work".
```

We can use the same blob of text in every PR since Kanban boards are now organisation-wide.

diegodelemos: Done, thanks!

PR #56, opened 2020-01-10
docs: remove reana-cluster references

diegodelemos:
* REANA-Cluster is being deprecated so we are removing all references
  to it (closes reanahub/reana#238).

* Updates developer instructions to comply with latest Helm procedures.

### DIFF
--- a/docs/developerguide.rst
+++ b/docs/developerguide.rst
@@ -93,16 +93,36 @@ REANA-UI uses Single Sign-On (SSO) for Authentication.
 
   * ``api``: REANA-Server hostname, e.g. ``https://reana-johndoe.cern.ch``.
 
-3. Configure SSO-related REANA-Cluster environment variables. You can do this in two ways:
+3. Configure SSO-related environment variables. You can do this in two ways:
 
    * Setting the variables before deploying on
-     `reana-cluster.yaml <https://github.com/reanahub/reana-cluster/blob/master/reana_cluster/configurations/reana-cluster-minikube-dev.yaml>`_.
+     `reanahub/reana`'s `helm/reana-dev/values.yaml <https://github.com/reanahub/reana/blob/master/helm/reana/values.yaml>`_.
+
+    .. code-block:: diff
+
+      + reana_url: <hostname> # e.g. "reana-johndoe.cern.ch"
+      debug:
+        enabled: false
+      ui:
+        enabled: false
+      ...
+      secrets:
+        ...
+        cern:
+          sso:
+      -      CERN_CONSUMER_KEY: <CHANGEME>
+      +      CERN_CONSUMER_KEY: <real_cern_consumer_client_id>
+      -      CERN_CONSUMER_SECRET: <CHANGEME>
+      +      CERN_CONSUMER_SECRET: <real_cern_consumer_secret>
+      ...
+
+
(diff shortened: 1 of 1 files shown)


docs/developerguide.rst
```diff
@@ -93,16 +93,37 @@ REANA-UI uses Single Sign-On (SSO) for Authentication.
...
+        enabled: false
+      ...
+      secrets:
+        ...
+        cern:
+          sso:
+      -      CERN_CONSUMER_KEY: reana
+      +      CERN_CONSUMER_KEY: <real_cern_consumer_client_id>
+      -      # CERN_CONSUMER_SECRET: <cern_consumer_secret>
+      +      CERN_CONSUMER_SECRET: <real-cern_consumer_secret>
```
mvidalgarcia: For the sake of consistency:
```suggestion
      +      CERN_CONSUMER_SECRET: <real_cern_consumer_secret>
```

docs/developerguide.rst
```diff
@@ -127,38 +148,27 @@ To setup the GitLab integration locally you need to follow these steps:
...
-      mountpoints:
-        - type: hostPath
-        name: reana-server-code
-        path: "/code/reana-server:/code"
-      environment:
+      secrets:
+        ...
+        gitlab:
+      -    REANA_GITLAB_OAUTH_APP_ID: reana
+      +    REANA_GITLAB_OAUTH_APP_ID: reana
```
mvidalgarcia: Same value as above, what about `<real_gitlab_oauth_app_id>`?

docs/developerguide.rst
```diff
@@ -127,38 +148,27 @@ To setup the GitLab integration locally you need to follow these steps:
...
+      -    # REANA_GITLAB_OAUTH_APP_SECRET: <gitlab_oauth_app_secret>
+      +    REANA_GITLAB_OAUTH_APP_SECRET: "<real-GitLab-Secret>"
```
mvidalgarcia: `<real_gitlab_oauth_app_secret>`?

reana-ui/src/config.js
```diff
@@ -10,7 +10,7 @@
 
 const config = {
   name: "CERN REANA UI",
-  api: "https://reana.cern.ch",
+  api: "https://mb-rodrigdi.cern.ch",
```
mvidalgarcia: :trollface:

docs/developerguide.rst
tiborsimko: :+1: One thing I always like is having a certain prefix for certain things to facilitate finding similar stuff via `git grep real_`.  

For example, we use `FIXME` in the code, so that `pylint` and friends will report these lines, but also so that developers can do anytime `git grep FIXME` to find all the places where further work is being tagged.  Having one standardised string is preferable to having `FIX ME`, `TODO`, and a plethora of other such similar tags.  

In the similar manner, we can invent a common prefix such as `real_...` and use it everywhere, as then we could say in the docs that "admins can look for this string to customise stuff".  

BTW note that elsewhere we used `CHANGEME` prefix for these things, it so could be also `CHANGEME_real_gitlab_oauth_app_secret`... but we also used `CHANGE_ME`, which is not cool, because `git grep` stops working. 

Let's harmonise the "customize this" prefix on some value and use it the same everywhere...

PR #61, opened 2020-01-29
ui: skeleton

### LINKED ISSUE #60: ui: skeleton
image

mvidalgarcia:
closes #60 

image

### DIFF
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -2,7 +2,7 @@
 	-*- coding: utf-8 -*-
 
 	This file is part of REANA.
-	Copyright (C) 2019 CERN.
+	Copyright (C) 2020 CERN.
 
   REANA is free software; you can redistribute it and/or modify it
   under the terms of the MIT License; see LICENSE file for more details.
--- a/reana-ui/src/components/TopHeader.js
+++ b/reana-ui/src/components/TopHeader.js
@@ -9,59 +9,54 @@
 */
 
 import React from "react";
-import { Segment, Image, Menu, Icon } from "semantic-ui-react";
+import { Image, Icon, Popup, List } from "semantic-ui-react";
 import { useDispatch } from "react-redux";
 import { Link } from "react-router-dom";
 import LogoImg from "../images/logo-reana.svg";
-import Cookies from "universal-cookie";
 import { userLogout } from "../actions";
+import config from "../config";
 
 import styles from "./TopHeader.module.scss";
 
-const cookies = new Cookies();
-
-const REANA_DOCS_URL = "https://reana.readthedocs.io/en/latest/";
-const REANA_SITE_URL = "http://www.reana.io";
-
 export default function TopHeader() {
   const dispatch = useDispatch();
-
-  /**
-   * Logs out the current session
-   */
   const logOut = () => {
-    cookies.remove("user_token");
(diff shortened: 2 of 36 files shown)


reana-ui/src/config.js
```diff
@@ -2,16 +2,19 @@
...
 
   REANA is free software; you can redistribute it and/or modify it
   under the terms of the MIT License; see LICENSE file for more details.
 */
 
 const config = {
   name: "CERN REANA UI",
-  api: "https://reana.cern.ch",
-  pooling_secs: 5
+  api: "https://cachopo.cern.ch",
```
diegodelemos: :trollface:

reana-ui/src/pages/ReanaPage.js
```diff
@@ -0,0 +1,25 @@
+/*
```
diegodelemos: Would `MainPage.js/BasePage.js/[insert your suggestion]` instead of `ReanaPage.js` work as well? in case we want to keep it, then it would be `REANAPage.js`.

mvidalgarcia: `BasePage.js` sounds good IMO.

reana-ui/src/pages/login/components/LoginForm.js
```diff
@@ -2,109 +2,45 @@
...
+    >
+      <Grid.Column className={styles["login-column"]}>
+        <Image
+          centered
+          spaced
+          src={LogoImg}
+          size="medium"
+          className={styles["reana-logo"]}
+        />
+        <Segment className={styles["signin"]}>
```
diegodelemos: I don't see `signin` class defined anywhere unless some magic is happening 🧙‍♂️.

mvidalgarcia: Good catch, leftover.

reana-ui/src/components/TopHeader.js
```diff
@@ -2,66 +2,61 @@
...
-        </Menu.Item>
-        <Menu.Item onClick={logOut} className={styles["logout-button"]}>
-          <Icon inverted name="user" />
-          Log out
-        </Menu.Item>
-      </Menu>
-    </Segment>
+      <section>
+        <a href={config.docs_url} target="_blank" rel="noopener noreferrer">
+          <Icon name="book"></Icon> Help
```
diegodelemos: The only thing I don't like is that this icon is the same as the docs one on the footer, could we use something else?

Making reference to the community, icon `group`:
Screenshot 2020-01-31 at 13 46 59
Classic question mark, icon `question circle outline`:
Screenshot 2020-01-31 at 13 45 41

Plus, the link to help, should it rather go to the forum instead of the docs? I would expect a place where to (1) check if someone had the same idea/problem and if not (2) ask, WDYT?

reana-ui/src/actions.js
```diff
@@ -2,7 +2,7 @@
 	-*- coding: utf-8 -*-
 
 	This file is part of REANA.
-	Copyright (C) 2019 CERN.
+	Copyright (C) 2020 CERN.
```
tiborsimko: Actually it should be `2019, 2020` since we keep previous years in the copyright messages.

(No need for action now, we can amend that programmatically one day...)

PR #71, opened 2020-02-12
ui: minor fixes

mvidalgarcia:
* Reduces polling frequency
* Displays "refreshed at" in workflow list
* Updates docs links

### DIFF
--- a/reana-ui/src/config.js
+++ b/reana-ui/src/config.js
@@ -11,8 +11,8 @@
 const config = {
   name: "CERN REANA UI",
   api: "https://reana.cern.ch",
-  poolingSecs: 5,
-  docsURL: "https://reana.readthedocs.io/en/latest/",
+  poolingSecs: 15,
+  docsURL: "http://docs.reana.io/",
   forumURL: "https://forum.reana.io/",
   mattermostURL: "https://mattermost.web.cern.ch/it-dep/channels/reana"
 };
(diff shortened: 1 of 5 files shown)


reana-ui/src/config.js
```diff
@@ -11,8 +11,8 @@
 const config = {
   name: "CERN REANA UI",
   api: "https://reana.cern.ch",
-  poolingSecs: 5,
-  docsURL: "https://reana.readthedocs.io/en/latest/",
+  poolingSecs: 15,
+  docsURL: "https://docs.reana.io/",
```
tiborsimko: Better use HTTP only for `docs.reana.io` still, until we have proper SSL certificate

reana-ui/src/pages/workflowList/components/WorkflowList.js
```diff
@@ -65,6 +66,7 @@ export default function WorkflowList() {
     if (!interval.current) {
       interval.current = setInterval(() => {
         getWorkflows();
+        setRefreshedAt(moment().format("LTS"));
```
tiborsimko: Usually I prefer 24/24 format, which does not need `PM` at the end, which gives us an option to display time zone?

``17:20:49 UTC`` instead of ``5:20:49 PM``

reana-ui/src/styles/palette.scss
```diff
@@ -9,7 +9,7 @@
 */
 
 $gray: #767676;
-$light-gray: #e9e9e9;
+$light-gray: #ccc;
```
tiborsimko: ``#cccccc`` six letters for full colour spec fullness?

PR #81, opened 2020-02-27
wf-files: add mime-type based whitelisting

### LINKED ISSUE #79: preview: better whitelist/blacklist
Currently, the Workspace file previewer doesn't allow to preview `.root` files, but it allows previewing `.exe` files, which is obviously not functional and better be avoided.

Let's improve whitelisting/blacklisting formats which are good for preview and which are forbidden from being previewed.

Pay attention to security, XSS and friends.

I'd say let's rather blacklist everything by default and allow previewing only good formats (`.py`, `.C`, `.png`, etc) rather than whitelisting everything by default (leading to the above `.exe` file problem.)

P.S. In the future, we could implement proper MIME type detection and such; this is about fixing current status quickly.

mvidalgarcia:
closes #79

### DIFF
--- a/reana-ui/src/util.js
+++ b/reana-ui/src/util.js
@@ -9,6 +9,7 @@
 */
 
 import moment from "moment";
+import mime from "mime";
 
 /**
  * Mapping between workflow statuses and colors and icons.
@@ -111,4 +112,17 @@ export function parseFiles(files) {
   });
 
   return files;
+}
+
+/**
+ * Returns mime-type of a given file name.
+ * @param {String} fileName File name
+ */
+export function getMimeType(fileName) {
+  // Formats not considered by mime npm package
+  const WHITELIST = [".py"];
+  if (WHITELIST.find(ext => fileName.endsWith(ext))) {
+    return "text/plain";
+  }
+  return mime.getType(fileName);
 }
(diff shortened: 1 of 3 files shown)


reana-ui/src/util.js
```diff
@@ -112,3 +113,16 @@ export function parseFiles(files) {
...
   return files;
 }
+
+/**
+ * Returns mime-type of a given file name.
+ * @param {String} fileName File name
+ */
+export function getMimeType(fileName) {
+  // Formats not considered by mime npm package
+  const WHITELIST = [".py", ".ipynb"];
```
tiborsimko: I'm not sure ipynb would be text/plain?  We might need special previewer for notebook files... kind of like for PDF files...

mvidalgarcia: Yeah definitely, something for the future perhaps? This PR is a temporary solution to avoid no plain text files/images to be previewed. IMO everything should be handled on server-side.

tiborsimko: Yes, but maybe take ipynb out of whitelist, so that people can only download it, because it currently looks like this: 

2020-02-29-214306_749x633_scrot


Better not to allow preview for ipynb files for now then...

mvidalgarcia: Updated.

PR #85, opened 2020-02-28
wf-details: fix non-existing workflow display

### LINKED ISSUE #82: wf-details: show message when accesing unauthorized workflow details
**Current behavior**

Loader spinning forever.

**Expected behavior**

Message saying something like "Sorry you are not authorized to see this workflow"

mvidalgarcia:
fixes #82

### DIFF
--- a/reana-ui/src/pages/workflowDetails/WorkflowDetails.js
+++ b/reana-ui/src/pages/workflowDetails/WorkflowDetails.js
@@ -45,7 +45,7 @@ function WorkflowDetails() {
     dispatch(fetchWorkflow(workflowId));
   }, [dispatch, workflowId]);
 
-  if (loading || !workflow) {
+  if (loading) {
     return (
       <Dimmer active>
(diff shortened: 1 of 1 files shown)


reana-ui/src/pages/workflowDetails/WorkflowDetails.js
```diff
@@ -45,20 +45,20 @@ function WorkflowDetails() {
...
   }
 
-  if (_.isEmpty(workflow)) {
+  if (!workflow) {
     return (
       <Container text className={styles.warning}>
         <Message
           icon="warning sign"
-          header="Workflow does not exist."
+          header="Sorry you are not authorised to see this workflow or does not exist."
```
tiborsimko: "or it does not..."

Perhaps better said in the inverse order, "Sorry, this workflow either does not exist or you are not authorised to see it." I'll amend the PR accordingly.

P.S. In the future we should split these two cases, as it might cause confusion...

PR #94, opened 2020-04-22
ui: request access token

### LINKED ISSUE #93: ui: user token creation through request/approval process
### ui

- [x] Create some fast mockups to discuss before starting with the actual implementation.
- [x] Prevent displaying the user token in the welcome page for social logged in users. Instead, display some message saying that social login requires verification and the token is in a "waiting" state and that administrator will contact him/her back once it's approved.
- [x] Show waiting/approved status in profile page
- [x] Handle 'revoked' case

### server
- [x] Create a new endpoint to be called from the ui to create a user token and mark it as `requested`.
  1. Use [`User.request_access_token()`](https://github.com/reanahub/reana-db/blob/master/reana_db/models.py#L102-L115)
  2. Send email to reana administrators

Tightly related to https://github.com/reanahub/reana-server/issues/235

mvidalgarcia:
closes #93

I went for a super simple solution for the moment.
<details>
  <summary>Welcome page</summary>
image
image
</details> 

<details>
  <summary>Profile page</summary>
image
image
</details>


tiborsimko:
Do you have an easy access to the date and time when the request was made?  We could print it in the information, it might give people more confidence.

"Token requested on 2020-04-30 13:35:30 CEST"

"Token requested three days ago"

(as we do for workflow runs)

mvidalgarcia:
> Do you have an easy access to the date and time when the request was made? We could print it in the information, it might give people more confidence.
> 
> "Token requested on 2020-04-30 13:35:30 CEST"
> 
> "Token requested three days ago"
> 
> (as we do for workflow runs)

This is not done. Ideally, we should pass a token object in the [`/me` endpoint](https://github.com/reanahub/reana-server/blob/master/reana_server/rest/users.py#L318) and plug more values like `created` and `updated` from [`Timestamp` class](https://github.com/reanahub/reana-db/blob/master/reana_db/models.py#L149). I'm not sure how [`updated` works](https://github.com/kvesteri/sqlalchemy-utils/blob/master/sqlalchemy_utils/models.py#L27) but it might be enough to display to the user if it is updated when the token is requested. We can create a ticket for later. WDYT?

PR #97, opened 2020-05-20
global: show user email in profile popup

### LINKED ISSUE #96: profile: display user information & revoked token access
- [x] a) Display user information

  At the moment we're displaying zero information about the logged-in user. We can show email, full name, username, etc.

  Take into account the lightweight account have strange usernames and full names, e.g.
  ```
  email: user@example.org
  full_name: user@example.org
  username: 5112b1a36ed11cdfde8e
  ```
  So perhaps only displaying email is the most appropriate for now.

- [x] b) Prevent user with a revoked token from access the UI.

mvidalgarcia:
closes #96

### DIFF
--- a/reana-ui/src/components/Error.js
+++ b/reana-ui/src/components/Error.js
@@ -1,0 +1,37 @@
+/*
+  -*- coding: utf-8 -*-
+
+  This file is part of REANA.
+  Copyright (C) 2020 CERN.
+
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+import React from "react";
+import { Container, Image } from "semantic-ui-react";
+import PropTypes from "prop-types";
+
+import LogoImg from "../images/logo-reana.svg";
+
+import styles from "./Error.module.scss";
+
+export default function Error({ message, title }) {
+  return (
+    <Container textAlign="center" className={styles.container}>
+      <Image centered spaced src={LogoImg} size="small" />
+      <h3>{title}</h3>
+      <div>{message}</div>
+    </Container>
+  );
+}
+
+Error.propTypes = {
+  message: PropTypes.string,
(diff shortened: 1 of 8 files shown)


reana-ui/src/components/Error.js
```diff
@@ -0,0 +1,35 @@
...
+
+import LogoImg from "../images/logo-reana.svg";
+
+import styles from "./Error.module.scss";
+
+export default function Error({ message }) {
+  return (
+    <Container textAlign="center" className={styles.container}>
+      <Image centered spaced src={LogoImg} size="small" />
+      <h1>Error</h1>
```
diegodelemos, line 23: Everything works nicely, I have just two small comments which might not be relevant, I leave it for your interpretation and @tiborsimko's:
image
_Screenshot to illustrate_
- The word error is very big compared to REANA logo (an impression which might be caused by the second point)
- Logo and "Error" (negative) word are very near to each other

mvidalgarcia: This was intended to be a first minimal version of error pages, there's a lot of room for improvement. I just wanted to avoid returning a plain text "Forbidden".

tiborsimko: Yeah, it's not really an "internal error" kind of situation, more an "information passing" kind of situation, which the  current mock-up may not evoke very well graphically. Two suggestions: instead of "Error" you could say "Access denied"; the informative message could say something more verbose like "Sorry, the user access token aaaa-bbbb-1111-2222 was revoked on YYYY-MM-DD hh:mm:ss and cannot be used anymore."  Otherwise fully agree, we can improve UI later for all the various error/warning/info kind of situations we might have.

mvidalgarcia: I've changed the title and made it smaller for now.

image

I'll create a ticket to further improve error pages. We might get more inspiration when we have more cases.

PR #101, opened 2020-06-02
ui: add announcement component

### LINKED ISSUE #83: ui: optional announcement
This can be useful for reana-dev and reana-qa instances or in case we want to announce e.g. an intervention with expected downtime.

Quick draft idea:

image

mvidalgarcia:
closes #83

image

### DIFF
--- a/reana-ui/src/components/Announcement.js
+++ b/reana-ui/src/components/Announcement.js
@@ -1,0 +1,47 @@
+/*
+  -*- coding: utf-8 -*-
+
+  This file is part of REANA.
+  Copyright (C) 2020 CERN.
+
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+import React, { useState } from "react";
+import { Icon } from "semantic-ui-react";
+import { useSelector } from "react-redux";
+
+import { getAnnouncement } from "../selectors";
+
+import styles from "./Announcement.module.scss";
+
+export default function Announcement() {
+  const hiddenAnnouncement = window.localStorage.getItem("hideAnnouncement");
+  const announcement = useSelector(getAnnouncement);
+  const [hidden, setHidden] = useState(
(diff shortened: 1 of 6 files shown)


reana-ui/src/components/Announcement.js
```diff
@@ -0,0 +1,47 @@
...
+import { useSelector } from "react-redux";
+
+import { getAnnouncement } from "../selectors";
+
+import styles from "./Announcement.module.scss";
+
+export default function Announcement() {
+  const hiddenAnnouncement = window.localStorage.getItem("hideAnnouncement");
+  const announcement = useSelector(getAnnouncement);
+  const [visible, setVisible] = useState(
```
diegodelemos, line 22: Shouldn't it be `hidden` and `setHidden` instead of `visible` and `setVisible`?

reana-ui/src/components/Announcement.js
```diff
@@ -0,0 +1,47 @@
...
+    window.localStorage.setItem("hideAnnouncement", announcement);
+    setVisible(window.localStorage.getItem("hideAnnouncement"));
+  }
+
+  return (
+    announcement && (
+      <div className={styles.bar} hidden={visible}>
+        <span className={styles.message}>
+          <Icon name="warning circle" />
+          {announcement}
```
diegodelemos, line 36: Tested the whole workflow, all works fine ✅. The only thing I see we could improve is that we cannot put links, if it is easy we can do it now, otherwise, we open an issue.

mvidalgarcia: Yes, I thought about it. In that case, I think we would need to use [`dangerouslySetInnerHTML`](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) which I don't like at all as implies security issues.
I don't know if there's any other way, we cam investigate it.

reana-ui/src/components/Announcement.js
mvidalgarcia: Indeed, I don't know what I was thinking... 🤦

PR #104, opened 2020-06-15
global: support local users

### LINKED ISSUE #107: auth: display sign in/up validation errors for local users
**Current behavior**

When local users sign in/up, the input submitted values are validated on server-side (Invenio-Accounts), however, when not complying with validation rules e.g. too short password, email address already associated to an account, etc. we're not displaying any error in the UI, even though we're getting a detailed response from the server containing the validation errors.

image

**Desired behavior**

Whenever these validation errors occur, we should display some feedback for the users.
For this, [accesssible labels](https://react.semantic-ui.com/collections/form/#shorthand-field-control-id) might come handy.

mvidalgarcia:
closes reanahub/reana-server#256
closes #107


mvidalgarcia:
The build should work now.

PR #111, opened 2020-07-16
ui: pagination

mvidalgarcia:
closes reanahub/reana-server#180

pagination

### DIFF
--- a/reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
+++ b/reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
@@ -41,7 +41,7 @@ const PREVIEW_MIME_PREFIX_WHITELIST = {
   "application/json": { serverPreviewable: false }
 };
 const SIZE_LIMIT = 5 * 1024 ** 2; // 5MB
-const PAGE_SIZE = 5; // TODO: Move to ui-configmap?
+const PAGE_SIZE = 15;
 
 export default function WorkflowFiles({ id }) {
   const dispatch = useDispatch();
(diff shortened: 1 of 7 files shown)


reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
```diff
@@ -36,21 +41,24 @@ const PREVIEW_MIME_PREFIX_WHITELIST = {
...
   const loading = useSelector(loadingDetails);
   const _files = useSelector(getWorkflowFiles(id));
+  const filesCount = useSelector(getWorkflowFilesCount(id));
 
   const [files, setFiles] = useState(null);
   const [modalContent, setModalContent] = useState(null);
   const [column, setColumn] = useState(null);
   const [direction, setDirection] = useState(null);
   const [isServerPreviewable, setIsServerPreviewable] = useState(false);
+  const [pagination, setPagination] = useState({ page: 1, size: PAGE_SIZE });
```
diegodelemos, line 57: There is some sort of incompatibility between the pagination and the sorting. Maybe we can move it to a different issue.

The problem:
Screen Recording 2020-07-21 at 12 03 31

mvidalgarcia: Thanks for spotting 👍 , it's fixed now.
Of course this sorting is far from perfect as it only sorts per page which is not very useful.
In order to fix this we probably need to allow a query param in server/w-controller `&sort=asc/desc` so clicking on the sorting button would trigger a new query. This can be a totally separate issue in case we need it.

mvidalgarcia: https://github.com/reanahub/reana-workflow-controller/issues/322

PR #112, opened 2020-07-29
docs: single-page RTFD outline

mvidalgarcia:
- Formats with Black

addresses reanahub/reana#335


docs/conf.py
```diff
@@ -92,52 +92,53 @@
...
+        "REANA@Web": "http://www.reana.io",
+    },
+    "nosidebar": True,
 }
 
 # Add any paths that contain custom _static files (such as style sheets) here,
 # relative to this directory. They are copied after the builtin _static files,
 # so a file named "default.css" will overwrite the builtin "default.css".
-html_static_path = ['_static']
+html_static_path = ["_static"]
```
diegodelemos, line 125:
```suggestion
# html_static_path = ["_static"]
```

README.rst
```diff
@@ -14,25 +14,49 @@
...
 Features
---------
+========
+
+- profile page containing REANA access token
+- list of personal workflows
+- workflow details page containing logs, files, specification
+- GitLab integration to load your workflow repositories
+
+Quick start
```
diegodelemos, line 35: I feel this `Quick start` before the `Usage` might be confusing, what do you think about adding it after and calling it something like `Development server`/`Development` or similar?

PR #117, opened 2020-09-09
ui: made welcome page more generic

### LINKED ISSUE #115: ui: make welcome page more generic and customizable
**Current look**

image

**Desired look**
- This page shouldn't have CERN-specific terms by default, note LXPLUS, `/afs/...`, Mattermost channel.. (can be replaced by `pip install reana-client` probably).
- This text should be customizable so each instance can have personalized instructions.

audrium:
closes #115


reana-ui/src/pages/workflowList/components/Welcome.js
```diff
@@ -64,12 +57,64 @@ export default function Welcome() {
...
+function WelcomeRegular() {
   return (
-    <Container text className={styles["container"]}>
-      <Title as="h2">Welcome to REANA!</Title>
-      {reanaToken ? tokenContent : <WelcomeNoTokenMsg />}
-    </Container>
+    <>
+      <p>
+        It seems that you are using REANA for the first time. Would you like to
+        try out a small example? Install the REANA command-line client
```
tiborsimko, line 67: We cannot say "Install the ..." here, because the shown example does more than that, i.e. it installs the client, sets up environment, clones example, and runs it.

I'd suggest to simply say:

```
It seems that you are using REANA for the first time. Would you 
like to try out a small example? Please proceed as follows:
```

tiborsimko (approved):

Issue #118, opened 2020-09-11
local deployment: token request page missing request date information


tiborsimko:
**(1)** When connecting first time to REANA Web UI, the user can ask for a token:

![2020-09-11-095908_723x198_scrot](https://user-images.githubusercontent.com/517546/92890374-5bcfed80-f417-11ea-9f43-1ce8e7a3e4bb.png)

After user asks for a token, we add a confirmation that the token was requested:

![2020-09-11-094801_714x177_scrot](https://user-images.githubusercontent.com/517546/92890456-6e4a2700-f417-11ea-98ea-28bffd4004fe.png)

However this happens only on the CERN instance.  For local deployments, the page does not change.

We should add this.

**(2)** While we are at this, elsewhere on REANA Web UI we used the following date and time formats:

![2020-09-11-101551_545x104_scrot](https://user-images.githubusercontent.com/517546/92890853-ca14b000-f417-11ea-8568-6e57f2e92438.png)

Let's use `YYYY-MM-DDZhh:mm:ss` consistently also here instead of `11th Sep...`.

audrium:
We've found out that the issue appears because "api/token" request returns 500 error since the SMTP envars are not set in the reana-server. So to fix this issue couple of more things needs to be improved:

- Add generic error notification component to inform users when something goes wrong (React Error Boundaries could be used)
- Improve redux actions to catch the errors

tiborsimko:
In addition: 

- There was no asked-for-request item in the DB tables as a result of email failure. I think we should treat the email notification as just that, that is a notification.  So, if the token request can be well processed on the backend side (because the DB is up and running nicely), then let's process it and let's treat the email notification as another secondary failure (becase the email server is down or something). In this way, the core functionality won't be disturbed by the notification functionality. This is because admins could be alerted by several means, e.g. via email notification, e.g. via a web page where these token requests will be listed, etc. So let's consider the email notification as not mandatory, and let's treat it as a "continuable error", not as a "fatal error", on the backend side.

- There was no entry in `audit_log` table about this failure. We may want to also log failed requests somewhere, since token business is kind of critical. What about logging also failed attempts in the same table with another status? We may want to look into improving error logging, so that cases like these would be available for easier peruse. (Something for later; in this issue we can improve (i) front-end error reporting so that users see what's happening and could ask for help; (ii) back-end handling so that if an email server is down, it wouldn't prevent the token request from being asked for.)

audrium:
[...]

Issue #99, opened 2020-05-25
ui: better error pages


mvidalgarcia:
Introduce better error pages for different HTTP status responses.
For now a tiny component `<Error />` was introduced. It permits to customize title and message.
![image](https://user-images.githubusercontent.com/8863238/82815378-56b35680-9e99-11ea-840e-be8f2361dc30.png)

A [good place](https://www.google.com/search?q=error+pages+ui) for inspiration.

mvidalgarcia:
Evaluate if a sign out link in these pages makes sense. At the moment a revoked user cannot sign out and try to sign in with a different account.

audrium:
We should also think about unifying the error handling across the application. Some disscusion happened in this PR: https://github.com/reanahub/reana-ui/pull/120#discussion_r487907511 with a suggestion to use a similar approach to one described [here](https://alexandrempsantos.com/sane-error-handling-react-redux/#after---how-to-solve-these-problems).

mvidalgarcia:
When accessing `reana.cern.ch/foo`, one gets a blank page. No errors HTTP occur involved as this is not foreseen.

![image](https://user-images.githubusercontent.com/8863238/100134751-fa051880-2e88-11eb-974f-9173833b8fa2.png)

We should redirect to an error page. This can be done [via React routes](https://github.com/reanahub/reana-ui/blob/28ec51bdd26705fd0cf3b3ec5efc4a22cdeace87/reana-ui/src/components/App.js#L53-L73). We probably need a new `Route` that matches all (`*`?) and is evaluated the last one, to redirect to a 404 page.

Issue #46, opened 2019-12-03
ui: centralize all API calls in a client file


mvidalgarcia:
We need to decide whether to use Web API's [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) or [axios](https://github.com/axios/axios) to perform the calls.

mvidalgarcia:
Related https://github.com/reanahub/reana-ui/issues/19

audrium:
I've refactored some of the `actions` which were using `fetch` to `axios` in this PR: https://github.com/reanahub/reana-ui/pull/120

Here is one of the functions which were refactored:

- `fetch`:
```
export function requestToken() {
  return async dispatch => {
    let resp, data;
    try {
      dispatch({ type: USER_REQUEST_TOKEN });
      resp = await fetch(USER_REQUEST_TOKEN_URL, {
        method: "PUT",
        credentials: "include"
      });
    } catch (err) {
      throw new Error(USER_INFO_URL, 0, err);
    }
    if (resp.status === 401) {
      data = await resp.json();
      console.log(data.message);
    } else if (resp.ok) {
      data = await resp.json();
    }
    dispatch({ type: USER_TOKEN_REQUESTED, ...data });
    return resp;
  };
}
```

- `axios`:
```
export const requestToken = () => async (dispatch) => {
  dispatch({ type: USER_REQUEST_TOKEN });
  return await axios.put(USER_REQUEST_TOKEN_URL, null, { withCredentials: true })
    .then(resp => dispatch({ type: USER_TOKEN_REQUESTED, ...resp.data }))
    .catch(err => {
      dispatch(errorActionCreator(err, USER_INFO_URL));
      dispatch({ type: USER_TOKEN_ERROR });
    });
};
```

IMO `axios` biggest advantage is that it allows to write less verbose code. Also if the response is not 200, it is easy to catch it as an error while with `fetch` you have to check if response status not `ok` and throw an error by yourself which leads to verbose code.

From Mozzila Fetch API docs: 
"An accurate check for a successful fetch() would include checking that the promise resolved, then checking that the Response.ok property has a value of true."
Link: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

I'd vote to centralize API calls using `axios`

mvidalgarcia:
>  @audrium: I guess we need to migrate the remaining `actions` to `axios` and plug the `errorActionCreator` to notify the users when some issue appears

[...]

PR #120, opened 2020-09-14
ui: added generic error notification component

audrium:
addresses #118

### DIFF
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -1,14 +1,15 @@
 /*
-	-*- coding: utf-8 -*-
+  -*- coding: utf-8 -*-
 
-	This file is part of REANA.
-	Copyright (C) 2020 CERN.
+  This file is part of REANA.
+  Copyright (C) 2020 CERN.
 
   REANA is free software; you can redistribute it and/or modify it
   under the terms of the MIT License; see LICENSE file for more details.
 */
 
 import _ from "lodash";
+import axios from "axios";
 
 import { api } from "./config";
 import { parseWorkflows, parseLogs, parseFiles } from "./util";
@@ -18,8 +19,12 @@ import {
   getWorkflowSpecification
 } from "./selectors";
 
+export const ERROR = "Error";
+export const CLEAR_ERROR = "Clear error";
+
 export const CONFIG_FETCH = "Fetch app config info";
 export const CONFIG_RECEIVED = "App config info received";
+export const CONFIG_ERROR = "Fetch app config error";
 
 export const USER_FETCH = "Fetch user authentication info";
 export const USER_RECEIVED = "User info received";
@@ -33,9 +38,11 @@ export const USER_SIGN_ERROR = "User sign in/up error";
 export const USER_SIGNEDOUT = "User signed out";
 export const USER_REQUEST_TOKEN = "Request user token";
 export const USER_TOKEN_REQUESTED = "User token requested";
+export const USER_TOKEN_ERROR = "User token error";
 
 export const WORKFLOWS_FETCH = "Fetch workflows info";
 export const WORKFLOWS_RECEIVED = "Workflows info received";
+export const WORKFLOWS_FETCH_ERROR = "Workflows fetch error";
 export const WORKFLOW_LOGS_FETCH = "Fetch workflow logs";
 export const WORKFLOW_LOGS_RECEIVED = "Workflow logs received";
 export const WORKFLOW_FILES_FETCH = "Fetch workflow files";
@@ -68,20 +75,23 @@ const WORKFLOW_FILES_URL = (id, { page = 1, size }) => {
   return url;
 };
 
+function errorActionCreator(error, name) {
+  const { status, data } = error?.response;
+  const { message } = data;
+  return { type: ERROR, name, status, message };
+};
+
+export const clearError = { type: CLEAR_ERROR };
+
 export function loadConfig() {
   return async dispatch => {
-    let resp, data;
-    try {
-      dispatch({ type: CONFIG_FETCH });
-      resp = await fetch(CONFIG_URL, { credentials: "include" });
-    } catch (err) {
-      throw new Error(CONFIG_URL, 0, err);
-    }
-    if (resp.ok) {
-      data = await resp.json();
-    }
-    dispatch({ type: CONFIG_RECEIVED, ...data });
-    return resp;
+    dispatch({ type: CONFIG_FETCH });
+    return await axios.get(CONFIG_URL, { withCredentials: true })
+      .then(resp => dispatch({ type: CONFIG_RECEIVED, ...resp.data }))
+      .catch(err => {
+        dispatch(errorActionCreator(err, CONFIG_URL));
+        dispatch({ type: CONFIG_ERROR });
+      });
   };
 }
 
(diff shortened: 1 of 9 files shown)


reana-ui/src/components/Notification.module.scss
```diff
@@ -0,0 +1,15 @@
...
+
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+@import '../styles/palette';
+
+.container {
+  margin-top: 2em;
+}
```
tiborsimko, line 15: Works nicely :+1: Two possible improvements:

- Please add "X" close icon so that users can close the notification window.
- Consider fading out the notification window after e.g. 10 seconds?  Hmm, perhaps it is good to keep it up "until closed by the user" for now...

audrium: Added "X" close button as well as autoclose feature which will fade the message away after 16 seconds.

Also, refactored some more user "actions" to catch and display error messages in the top. So now it will work with following user requests:
- load config
- request token
- fetch workflows

tiborsimko: Thanks, fading out and closure both work nicely.  

One cosmetic comment about styling:

- on detailed record page, we have:

2020-09-14-140613_727x95_scrot


- on the token request page, we have:

2020-09-14-140643_722x105_scrot

Shall we perhaps harmonise the colouring and closure icons?

audrium: Changed the workflow details page to use the new notification component with the same colors. Also, introduced "closable" property in the notification component to stay in the page as in workflow details page case.

reana-ui/src/reducers.js
```diff
@@ -65,6 +72,18 @@ const detailsInitialState = {
   loadingDetails: false
 };
 
+const error = (state = errorInitialState, action) => {
```
diegodelemos, line 75: I have just one doubt now, with this change we will have now error data split into `auth.error` and `error`. Also, we have to call the reducers twice, to handle the error in the original part of the state tree (e.g. `workflow`) and in the `error` part:
```javascript
...
      dispatch(getError(err, USER_INFO_URL));
      dispatch({ type: WORKFLOWS_ERROR });
...
```
I wonder if we could somehow centralize this behavior (I've found [this](https://alexandrempsantos.com/sane-error-handling-react-redux/#after---how-to-solve-these-problems) which might be useful). What do you think?

reana-ui/src/components/Notification.js
```diff
@@ -0,0 +1,57 @@
...
+import { useDispatch, useSelector } from "react-redux";
+import { Container, Message, Transition } from "semantic-ui-react";
+import PropTypes from "prop-types";
+
+import { clearError } from "../actions";
+import { getError } from "../selectors";
+
+import styles from "./Notification.module.scss";
+
+export default function Notification({ icon, message }) {
```
diegodelemos: Not to solve within this PR, but maybe to be ticketised if we agree:

Would it make sense to use the state for `message`(s) too? That way, the components that want to flash a message just push it to the state and `Notification` just shows it, as when an error occurs.

What do you think?

reana-ui/src/components/Notification.js
```diff
@@ -0,0 +1,57 @@
...
+    clearTimeout(timer.current);
+    timer.current = setTimeout(() => hide(), 16000);
+  }
+
+  return (
+    <Transition visible={visible} duration={300}>
+      <Container text className={styles.container}>
+        <Message
+          icon={icon}
+          header={message || error?.message}
```
diegodelemos, line 39: What do you think about changing `header` to be something shorter and generic such as "An error has occurred" (since it is thought as a title), and then inside `content` we could add `error.message` for details?

diegodelemos (commented):
Works great, a couple of minor comments and two that need some feedback :)

reana-ui/src/reducers.js
```diff
@@ -30,6 +35,8 @@ import {
 } from "./actions";
 import { USER_ERROR } from "./errors";
 
+const errorInitialState = null;
```
diegodelemos, line 38: One last comment which doesn't have to be addressed in this PR. Should we make the error field a list? Right now if I have two errors in a row in a short time, the second error "overwrites" the first and I can only see the latter.

reana-ui/src/actions.js
```diff
@@ -157,51 +165,29 @@ export function userSignout() {
...
-      data = await resp.json();
-      console.log(data.message);
-    } else if (resp.ok) {
-      data = await resp.json();
-    }
-    dispatch({ type: USER_TOKEN_REQUESTED, ...data });
-    return resp;
-  };
-}
+export const requestToken = () => async (dispatch) => {
```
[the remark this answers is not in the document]
audrium: Thanks a lot for the feedback! I was checking the project code and got an impression that there is a switch from using `fetch` and traditional functions to `axios` + arrow functions in the newer code. As an example I checked `GitLabProjects` component. So that's why I started using these two in the updated code. But I understand now that it's not like that :) So in this case I totally agree that we should keep the consistency. I can refactor `actions` to traditional functions

reana-ui/src/actions.js
```diff
@@ -68,22 +75,23 @@ const WORKFLOW_FILES_URL = (id, { page = 1, size }) => {
...
-      throw new Error(CONFIG_URL, 0, err);
-    }
-    if (resp.ok) {
-      data = await resp.json();
-    }
-    dispatch({ type: CONFIG_RECEIVED, ...data });
-    return resp;
-  };
-}
+const getError = (error, name) => {
```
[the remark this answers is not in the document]
audrium: Yeah, good catch. Will try to find a better name for the function

reana-ui/src/actions.js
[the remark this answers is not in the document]
audrium: At least for me arrow functions are more readable and sometimes can save some extra lines. Like in this case:

```
export function loadConfig() {
  return async dispatch => {
    dispatch({ type: CONFIG_FETCH });
    return await axios.get(CONFIG_URL, { withCredentials: true })
      .then(resp => dispatch({ type: CONFIG_RECEIVED, ...resp.data }))
      .catch(err => {
        dispatch(getError(err, CONFIG_URL));
        dispatch({ type: CONFIG_ERROR });
      });
  }
};
```

```
export const loadConfig = () => async (dispatch) => {
  dispatch({ type: CONFIG_FETCH });
  return await axios.get(CONFIG_URL, { withCredentials: true })
    .then(resp => dispatch({ type: CONFIG_RECEIVED, ...resp.data }))
    .catch(err => {
      dispatch(getError(err, CONFIG_URL));
      dispatch({ type: CONFIG_ERROR });
    });
};
```

reana-ui/src/reducers.js
audrium: Very good remarks, thanks. With the `error` reducer I wanted to introduce a single place to store errors for the whole app. As a benefit it is very easy to know if there is something wrong with the app and to show it in the notification bar to the users. I really like the idea described in the link to unify the error handling, but as I understand it, it does not provide a central place to store errors. Each reducer will have it's own error object (because of combined reducers). Maybe there is a way to achieve it with some kind of similar approach?

diegodelemos: I see, thanks for the elaborated response! Maybe what we have is a naming issue then: if for instance `WORKFLOWS_ERROR` would be named `WORKFLOWS_FETCH_FAILED` it would make more sense in my eyes, in the end, what it does is just stopping the loader. This way the previous snipped would read:
```diff
...
      dispatch(getError(err, USER_INFO_URL));
-     dispatch({ type: WORKFLOWS_ERROR });
+     dispatch({ type: WORKFLOWS_FETCH_FAILED });
...
```

The problem is less evident with for instance [`USER_FETCH_ERROR`](https://github.com/reanahub/reana-ui/blob/97e0f77dcf44f0213c2b64ffdce2f349669bc9de/reana-ui/src/reducers.js#L108-L113) (or similar) since it handles the message too. Maybe the best would be to create an issue documenting this centralized approach you have started, so we can discuss it with @mvidalgarcia and standardize error management and notifications.

reana-ui/src/actions.js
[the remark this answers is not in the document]
diegodelemos: I agree, maybe what we can do is keep it as it mainly is right now and discuss this later with @mvidalgarcia, as long as we keep consistency I am fine with both.

reana-ui/src/actions.js
```diff
@@ -157,51 +165,29 @@ export function userSignout() {
...
-    } else if (resp.ok) {
-      data = await resp.json();
-    }
-    dispatch({ type: USER_TOKEN_REQUESTED, ...data });
-    return resp;
-  };
-}
+export const requestToken = () => async (dispatch) => {
+  dispatch({ type: USER_REQUEST_TOKEN });
+  return await axios.put(USER_REQUEST_TOKEN_URL, null, { withCredentials: true })
```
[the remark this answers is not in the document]
audrium: Left a comment there together with my observations about the two

reana-ui/src/actions.js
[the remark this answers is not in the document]
audrium: Renamed it to `errorActionCreator` which I took it from the blog post you mentioned :)

reana-ui/src/reducers.js
audrium: Your suggestion makes mores sense to me as well, changed it accordingly. Also, wrote a comment in https://github.com/reanahub/reana-ui/issues/99 to come back to this discussion in the future.

reana-ui/src/components/Notification.js
audrium: Created a new issue to address it: https://github.com/reanahub/reana-ui/issues/123

diegodelemos (approved):

tiborsimko (approved):

reana-ui/src/actions.js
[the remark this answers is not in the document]
mvidalgarcia: - WRT arrow functions vs regular functions: I don't have a strong preference. I think it's ok to use arrow functions as long as we don't deal with `this`. But I would need to read more about it.
- WRT `axios` vs `fetch`: I think `axios` is definitely more powerful. OTOH `fetch` comes with the Web API so we save a dependency and our app is lighter. The idea was finding pros and cons in [this issue](https://github.com/reanahub/reana-ui/issues/46) and centralize all the calls to the REST API in a single file (possibly `client.js`). This way we can encapsulate all the complexity in this file (no matter if we use `fetch` or `axios`!), perhaps create a class and perform the calls something like `Client.getWorkflows()`. What do you think?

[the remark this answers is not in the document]
audrium: I think it's a good idea to separate api calls from redux actions, but having it in a single file could bloat it very quickly in the future. Perhaps it could be divided into some logical units (maybe similar how reana-server is divided - workflows, users, gitlab) and placed in `/api` folder?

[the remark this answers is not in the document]
mvidalgarcia: Definitely, we can split it as it grows.

PR #121, opened 2020-09-14
ui: fixed firefox issue in workflow details page

### LINKED ISSUE #116: detailed workflow page: improve Firefox experience
One Web UI observation regarding Chromium vs Firefox. 

 - When I reload a workflow detailed page in Chromium, everything looks fine: the page gets white, the REANA logo gets "foggy", the "Loading workflow..." wheel appears, and then the new page content. All smooth and nice.

 - However, when using Firefox, I also see the access error message box "Sorry this workflow either does not exist..." (etc) flashing briefly in the beginning of this process.  One cannot see it for long, so it is kind of OK, but a bit disturbing... It doesn't happen on every reload, but at least half the time, so people will get to notice this, I think. 

Is there something we could do to improve this for Firefox users?

P.S. React lifecycle methods?

audrium:
closes #116

### DIFF
--- a/reana-ui/src/pages/workflowDetails/WorkflowDetails.js
+++ b/reana-ui/src/pages/workflowDetails/WorkflowDetails.js
@@ -14,7 +14,7 @@ import { useParams } from "react-router-dom";
 import { Container, Dimmer, Loader, Message, Tab } from "semantic-ui-react";
 
 import { fetchWorkflow } from "../../actions";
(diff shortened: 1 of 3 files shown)


audrium:
Workflow details page flashed "no workflows error" because the loader was not initialized yet. Only after sending the request to fetch all the workflows the loader state was set to "true". There was a fraction of a second between these states and Chrome decided not to render it :)

diegodelemos (approved):

PR #126, opened 2020-09-17
ui: improved the looks of pagination

### LINKED ISSUE #124: ui: improve the looks of pagination
Icon styling could be improved by specifying the icon names like in this [example](https://react.semantic-ui.com/addons/pagination/#usage-shorthand)

audrium:
closes #124

### DIFF
--- a/reana-ui/src/pages/workflowList/WorkflowList.js
+++ b/reana-ui/src/pages/workflowList/WorkflowList.js
@@ -12,7 +12,7 @@ import _ from "lodash";
 import moment from "moment";
 import React, { useEffect, useRef, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
-import { Dimmer, Loader, Pagination } from "semantic-ui-react";
+import { Dimmer, Loader } from "semantic-ui-react";
 
 import { fetchWorkflows } from "../../actions";
 import {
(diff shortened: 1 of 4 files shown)


audrium:
Workflow List (Old):


Workflow List (New):


Workflow Files (Old):


Workflow Files (New):

reana-ui/src/pages/workflowList/WorkflowList.js
```diff
@@ -92,8 +93,6 @@ function Workflows() {
               interval.current = null;
```
mvidalgarcia, line 93: I guess you can also remove the `defaultActivePage` here (4 lines above).

mvidalgarcia (approved):
LGTM 👍

reana-ui/src/pages/workflowList/WorkflowList.js
audrium: Yep, I tried that, it seems that `semantic-ui-react` does not set this value by default, so removing it results in an error:
```Received NaN for the `value` attribute. If this is expected, cast the value to a string.```.
So it ether have to be `defaultActivePage` or `activePage` prop set for this to work correctly.

mvidalgarcia: Ah ok, I see 👍

PR #134, opened 2020-09-28
ui: display user quota on profile page 

### LINKED ISSUE #133: ui: display "Your quota" in the profile page
image

mvidalgarcia:
closes #133

image

### DIFF
--- a/reana-ui/src/pages/profile/components/Quota.js
+++ b/reana-ui/src/pages/profile/components/Quota.js
@@ -1,0 +1,85 @@
+/*
+  -*- coding: utf-8 -*-
+
+  This file is part of REANA.
+  Copyright (C) 2020 CERN.
+
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+import React from "react";
+import { useSelector } from "react-redux";
+import { PieChart } from "react-minimal-pie-chart";
+import { Grid, Label } from "semantic-ui-react";
+
+import { getUserQuota } from "../../../selectors";
+
+import styles from "./Quota.module.scss";
+
+const SEPIA_COLOR = "#f5ecec";
+const DARK_SEPIA_COLOR = "#b68181";
+
+export default function Quota() {
+  const quota = useSelector(getUserQuota);
+
+  function renderPieChart(title, unit, quota) {
+    const { usage, limit, health } = quota;
+    const percentage = Math.round((usage / limit) * 100);
+    const hasLimit = limit > 0;
+    const quotaHealthMapping = {
+      healthy: "green",
+      warning: "orange",
+      critical: "red",
+    };
+
+    return (
+      <Grid.Column className={styles.column}>
+        <PieChart
+          data={[
+            {
+              title: title,
+              value: hasLimit ? usage : 0,
+              color: DARK_SEPIA_COLOR,
+            },
+          ]}
+          lineWidth={30}
+          background={SEPIA_COLOR}
+          totalValue={limit}
+          startAngle={270}
+          style={{ height: "150px", flexBasis: "200px" }}
+          label={() => title}
+          labelStyle={{
+            fontSize: "14px",
+            fill: DARK_SEPIA_COLOR,
+          }}
+          labelPosition={0}
+        />
+        <div className={styles["quota-details"]}>
+          <div className={styles.usage}>
+            <h3>
+              {usage} {unit}
+            </h3>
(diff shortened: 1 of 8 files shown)


audrium (approved):

reana-ui/src/pages/profile/components/Quota.js
```diff
@@ -0,0 +1,68 @@
...
+import { Grid, Label } from "semantic-ui-react";
+
+import styles from "./Quota.module.scss";
+
+const SEPIA_COLOR = "#f5ecec";
+const DARK_SEPIA_COLOR = "#b68181";
+
+export default function Quota() {
+  const dummy = [
+    { title: "CPU", value: 71, limit: 120, unit: "min" },
```
diegodelemos, line 22: Some remarks on these fields:
- `title`: missing from current REST API, should we add title? This comment is closely related to https://github.com/reanahub/reana-server/pull/288#discussion_r498788299 so maybe we can merge leaving "CPU" hardcoded and improve this with the issue that will be created from this comment (implementation clarification: this is [not the title of a Resource](https://github.com/reanahub/reana-db/blob/63dc70464cf05e5337cb0c29826807bc2d4895bf/reana_db/models.py#L523) but actually the [name of a ResourceType](https://github.com/reanahub/reana-db/blob/63dc70464cf05e5337cb0c29826807bc2d4895bf/reana_db/models.py#L498)).
- `value`: for now in canonical units
- `limit`: to be removed until we implement limits in https://github.com/reanahub/reana-server/issues/282
- `unit`: I am not sure this should come from the server. Until now I assumed we would always return in canonical units (keep REST API as raw as possible) and let the clients figure out because in the end this is just a presentational problem (which unit to display in).
    - Server-side conversions:
       - Pros
           - Lighter/simpler client(s)
           - Centralised logic
           - Less code overall
       - Cons
           - More complex REST API
           - REST API deals with presentational duties

    What is your opinion on this?

reana-ui/src/pages/profile/components/Quota.js
```diff
@@ -0,0 +1,68 @@
...
+    { title: "HDD", value: 83232, limit: 123456, unit: "KiB" },
+  ];
+  return (
+    <Grid columns={2}>
+      {dummy.map(({ title, value, limit, unit }) => {
+        const percentage = ((value / limit) * 100).toFixed(2);
+        let percentageColor = "green";
+        // TODO: To be consistent with `reana-client quota-show`
```
diegodelemos, line 30: Maybe we could return a `quota_health` field from the server and do this logic there? If you think it makes sense we could add it as part of https://github.com/reanahub/reana-server/issues/289.

Note: this kind of goes a bit against the comment of doing the unit conversion client side... maybe I am leaning towards doing everything server side... Looking forward to reading your feedback.

mvidalgarcia: I agree, if it's going to be used both in `r-ui` and `r-client`, it makes sense to serve it from `r-server`.

reana-ui/src/pages/profile/components/Quota.js
mvidalgarcia: I don't know.. even if we always return the canonical units, if I was to consume the REST API looking for CPU quota and I got a raw number, I wouldn't have guessed that it's in milliseconds. We even have a `ResourceUnit` enum for it, if we follow that logic, we wouldn't need it at all.

reana-ui/src/pages/profile/components/Quota.js
```diff
@@ -0,0 +1,68 @@
...
+          } else {
+            percentageColor = "orange";
+          }
+        }
+        return (
+          <Grid.Column className={styles.column}>
+            <PieChart
+              data={[{ title: title, value: value, color: DARK_SEPIA_COLOR }]}
+              background={SEPIA_COLOR}
+              totalValue={limit}
```
mvidalgarcia, line 43: Now that we have implemented that `quota_limit = 0` means no limit. How can we represent this with pie charts? I'm afraid they don't make sense in the case.

mvidalgarcia:
Some ideas for the "no limit" representation:

a)
image

b)
image

c)
image

ParthS007:
I like the first one. 1️⃣

diegodelemos:
I like `c)` followed by `b)`. Regarding `a)` although the most visually pleasant for me, infinite might convey the wrong message "use really as much as you want", which despite the fact of being the reality, might trigger certain unwanted user behaviours.

audrium:
I also vote for `c)` :1st_place_medal:

mvidalgarcia:
Added version `d)` base on @tiborsimko feedback.

PR #136, opened 2020-09-30
ui: introduce workflow deletion

### LINKED ISSUE #129: quotas: allow users to recover quota
- Provide `reana-client` instructions on how to delete workflow runs (MVP)
- Offer buttons to delete workflow runs
- Show quota consumption per workflow

mvidalgarcia:
closes #129 
image

image

### DIFF
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -25,7 +25,8 @@ import {
 } from "./selectors";
 
 export const ERROR = "Error";
-export const CLEAR_ERROR = "Clear error";
+export const NOTIFICATION = "Notification";
+export const CLEAR_NOTIFICATION = "Clear notification";
 
 export const CONFIG_FETCH = "Fetch app config info";
 export const CONFIG_RECEIVED = "App config info received";
@@ -78,10 +79,20 @@ const WORKFLOW_SET_STATUS_URL = (id, status) =>
 function errorActionCreator(error, name) {
   const { status, data } = error?.response;
   const { message } = data;
-  return { type: ERROR, name, status, message };
+  return {
+    type: ERROR,
+    name,
+    status,
+    message,
+    header: "An error has occurred",
+  };
 }
 
-export const clearError = { type: CLEAR_ERROR };
+export function triggerNotification(header, message) {
+  return { type: NOTIFICATION, header, message };
+}
+
+export const clearNotification = { type: CLEAR_NOTIFICATION };
 
 export function loadConfig() {
   return async (dispatch) => {
@@ -316,7 +327,7 @@ export function deleteWorkflow(id, workspace = false) {
       )
       .then((resp) => {
         dispatch({ type: WORKFLOW_DELETED, ...resp.data });
-        return resp.data;
+        dispatch(triggerNotification("Success!", resp.data.message));
       })
       .catch((err) => {
         dispatch(
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -321,7 +321,7 @@ export function deleteWorkflow(id, workspace = false) {
     dispatch({ type: WORKFLOW_DELETE_INIT });
     return await axios
       .put(
-        WORKFLOW_SET_STATUS_URL(id, "deleted"),
+        WORKFLOW_SET_STATUS_URL(id, { status: "deleted" }),
         { workspace },
         { withCredentials: true }
       )
@@ -331,7 +331,10 @@ export function deleteWorkflow(id, workspace = false) {
       })
       .catch((err) => {
         dispatch(
-          errorActionCreator(err, WORKFLOW_SET_STATUS_URL(id, "deleted"))
+          errorActionCreator(
+            err,
+            WORKFLOW_SET_STATUS_URL(id, { status: "deleted" })
+          )
         );
       });
   };
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -55,6 +55,8 @@ export const WORKFLOW_FILES_RECEIVED = "Workflow files received";
 export const WORKFLOW_SPECIFICATION_FETCH = "Fetch workflow specification";
 export const WORKFLOW_SPECIFICATION_RECEIVED =
   "Workflow specification received";
+export const WORKFLOW_DELETE_INIT = "Initialize workflow deletion";
+export const WORKFLOW_DELETED = "Workflow deleted";
 
 const CONFIG_URL = `${api}/api/config`;
 const USER_INFO_URL = `${api}/api/you`;
@@ -63,12 +65,15 @@ const USER_SIGNIN_URL = `${api}/api/login`;
 const USER_SIGNOUT_URL = `${api}/api/logout`;
 const USER_REQUEST_TOKEN_URL = `${api}/api/token`;
 const WORKFLOWS_URL = (params) =>
-  `${api}/api/workflows?${stringifyQueryParams(params)}`;
+  `${api}/api/workflows?verbose=true&${stringifyQueryParams(params)}`;
 const WORKFLOW_LOGS_URL = (id) => `${api}/api/workflows/${id}/logs`;
 const WORKFLOW_SPECIFICATION_URL = (id) =>
   `${api}/api/workflows/${id}/specification`;
 const WORKFLOW_FILES_URL = (id, pagination) =>
   `${api}/api/workflows/${id}/workspace?${stringifyQueryParams(pagination)}`;
+
+const WORKFLOW_SET_STATUS_URL = (id, status) =>
+  `${api}/api/workflows/${id}/status?${stringifyQueryParams(status)}`;
 
 function errorActionCreator(error, name) {
   const { status, data } = error?.response;
@@ -297,5 +302,26 @@ export function fetchWorkflowSpecification(id) {
       parameters: data.parameters,
     });
     return resp;
+  };
+}
+
+export function deleteWorkflow(id, workspace = false) {
+  return async (dispatch) => {
+    dispatch({ type: WORKFLOW_DELETE_INIT });
+    return await axios
+      .put(
+        WORKFLOW_SET_STATUS_URL(id, "deleted"),
+        { workspace },
+        { withCredentials: true }
+      )
+      .then((resp) => {
+        dispatch({ type: WORKFLOW_DELETED, ...resp.data });
+        return resp.data;
+      })
+      .catch((err) => {
+        dispatch(
+          errorActionCreator(err, WORKFLOW_SET_STATUS_URL(id, "deleted"))
+        );
+      });
   };
 }
--- a/reana-ui/src/pages/workflowList/components/WorkflowActionsPopup.js
+++ b/reana-ui/src/pages/workflowList/components/WorkflowActionsPopup.js
@@ -9,11 +9,14 @@
 */
 
 import PropTypes from "prop-types";
+import { workflowShape } from "../../../props";
 import React, { useState } from "react";
 import { Icon, Menu, Popup } from "semantic-ui-react";
 import { useDispatch } from "react-redux";
 
 import { deleteWorkflow } from "../../../actions";
+
+import styles from "./WorkflowActionsPopup.module.scss";
 
 export default function WorkflowActionsPopup({
   workflow,
@@ -52,15 +55,18 @@ export default function WorkflowActionsPopup({
   }
   return (
     <Popup
+      basic
       trigger={
         <Icon
           name="ellipsis vertical"
+          className={styles.icon}
           onClick={(e) => {
             setOpen(true);
             e.stopPropagation();
           }}
         />
       }
+      position="bottom left"
       on="click"
       open={open}
       onClose={() => setOpen(false)}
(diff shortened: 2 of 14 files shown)


reana-ui/src/actions.js
```diff
@@ -293,3 +308,24 @@ export function fetchWorkflowSpecification(id) {
     return resp;
   };
 }
+
+export function deleteWorkflow(id, workspace = false) {
+  return async (dispatch) => {
+    dispatch({ type: WORKFLOW_DELETE_INIT });
+    return await axios
+      .put(
```
ParthS007, line 316: Just wondering which will be best here, `PUT` or `PATCH` since we are just changing the status of workflow and not actually deleting it. Please correct me if I am wrong.

reana-ui/src/pages/workflowList/components/WorkflowActionsPopup.js
```diff
@@ -0,0 +1,77 @@
...
+        setOpen(false);
+        e.stopPropagation();
+      },
+    });
+  }
+  return (
+    <Popup
+      trigger={
+        <Icon
+          name="ellipsis vertical"
```
ParthS007, line 57: What do you think if we have a `horizontal ellipsis` in the top right corner with the `Menu` opening downside instead of upside?

reana-ui/src/pages/workflowList/components/WorkflowActionsPopup.js
```diff
@@ -0,0 +1,77 @@
...
+          onClick={(e) => {
+            setOpen(true);
+            e.stopPropagation();
+          }}
+        />
+      }
+      on="click"
```
ParthS007, line 64: what does this `on` prop do? I don't see `on` prop in [react-semantic-ui `Icon`](https://react.semantic-ui.com/elements/icon/#groups-icon-group) !

reana-ui/src/pages/workflowList/components/WorkflowActionsPopup.js
```diff
@@ -0,0 +1,77 @@
...
+      open={open}
+      onClose={() => setOpen(false)}
+    >
+      <Menu items={menuItems} secondary vertical />
+    </Popup>
+  );
+}
+
+WorkflowActionsPopup.propTypes = {
+  workflow: PropTypes.object.isRequired,
```
ParthS007, line 74: Can you also expand object proptype as well, which means we also have an idea of what to expect inside the `workflow` object?

reana-ui/src/pages/workflowList/components/WorkflowDeleteModal.js
```diff
@@ -0,0 +1,75 @@
...
+        </Button>
+        <Button onClick={onCloseModal}>Cancel</Button>
+      </Modal.Actions>
+    </Modal>
+  );
+}
+
+WorkflowDeleteModal.propTypes = {
+  open: PropTypes.bool.isRequired,
+  workflow: PropTypes.object.isRequired,
```
ParthS007, line 72: same.

reana-ui/src/pages/workflowList/components/WorkflowList.js
```diff
@@ -42,22 +45,32 @@ export default function WorkflowList({ workflows, refreshedAt }) {
...
+        return (
           <div
             key={id}
             onClick={() => history.push(`/details/${id}`)}
             className={`${styles["workflow"]} ${
               status === "deleted" ? styles["deleted"] : ""
             }`}
           >
-            <div>
+            <div className={styles["details-box"]}>
```
ParthS007, line 59: Somewhere we are following `dot` representation(`x.y`) and at other places, we are referring like `key` in the dict(`x["y"]`). am I missing something?

```suggestion
            <div className={styles.details-box}>
```

ParthS007 (commented):
Works fine, tested locally! 🚀 

Some UI suggestions and questions 😃

reana-ui/src/pages/workflowList/components/WorkflowList.js
```diff
@@ -42,22 +45,32 @@ export default function WorkflowList({ workflows, refreshedAt }) {
...
                 name={statusMapping[status].icon}
                 color={statusMapping[status].color}
               />{" "}
-              <span className={styles["name"]}>{name}</span>
-              <span className={styles["run"]}>#{run}</span>
+              <span className={styles.name}>{name}</span>
+              <span className={styles.run}>#{run}</span>
+              <span
+                className={`${styles.size} ${
+                  status === "deleted" && size !== "0K" ? styles.highlight : ""
```
audrium, line 68: we have many checks for `status === "deleted" ` while rendering the component. I would propose to create a new variable (something like `statusDeleted`) with this check in line 50 and use it while rendering

audrium: we could even have another one for `status === "deleted" && size !== "0K"` just to give it a name and to be more declerative

reana-ui/src/pages/workflowList/components/WorkflowActionsPopup.js
```diff
@@ -0,0 +1,77 @@
...
+export default function WorkflowActionsPopup({
+  workflow,
+  setOpenDeleteModal,
+  setSelectedWorkflow,
+}) {
+  const dispatch = useDispatch();
+  const [open, setOpen] = useState(false);
+  const { id, name, run, size, status } = workflow;
+  let menuItems = [];
+  if (status === "deleted") {
```
audrium, line 27: Some crazy idea - maybe we need to introduce another status for deleted workflows with deleted workspaces?
If it's not needed anywhere else (in the backend), we could simply set it in the redux reducers. It would simplify workflow list rendering as well

reana-ui/src/pages/workflowList/components/WorkflowList.js
```diff
@@ -79,18 +92,35 @@ export default function WorkflowList({ workflows, refreshedAt }) {
...
+              {statusMapping[status].preposition}{" "}
+              {status !== "deleted" && duration}
               <div>
                 step {completed}/{total}
               </div>
             </div>
+            <div className={styles.actions}>
+              {(status !== "deleted" ||
+                (status === "deleted" && size !== "0K")) && (
+                <WorkflowActionsPopup
```
audrium, line 108: I noticed that it's very easy to miss the button when you are trying to press it. Then and you are being redirected to `workflow details` page. So I wonder if we could increase the margin of the button, so it could have a bit bigger area

reana-ui/src/pages/workflowList/components/WorkflowActionsPopup.js
audrium: It triggering the popup: https://react.semantic-ui.com/modules/popup/#variations-size

reana-ui/src/actions.js
mvidalgarcia: I think you're right. If it were complete deletion we should use `DELETE` though. In this case, as we're only sending the new `status` and not the full workflow object it should be a [`PATCH`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH).

I left the API untouched as it was, I don't know if it's worth changing it at the moment as it'd affect other places. Opinions? @tiborsimko @diegodelemos @audrium

reana-ui/src/pages/workflowList/components/WorkflowDeleteModal.js
```diff
@@ -0,0 +1,75 @@
...
+    setSelectedWorkflow({});
+    setDeleteWorkspace(true);
+  };
+
+  const { id, name, run, size } = workflow;
+  return (
+    <Modal open={open} onClose={onCloseModal} closeIcon>
+      <Modal.Header>Delete workflow</Modal.Header>
+      <Modal.Content>
+        {
```
audrium, line 38: Correct me if I'm wrong, but I think curly brackets are not needed here

reana-ui/src/pages/workflowList/components/WorkflowDeleteModal.js
```diff
@@ -0,0 +1,75 @@
...
+          <>
+            <p>
+              Are you sure you want to delete workflow "{name} #{run}"?
+            </p>
+            <Checkbox
+              checked={deleteWorkspace}
+              onChange={(_, data) => {
+                setDeleteWorkspace(data.checked);
```
audrium, line 46: Super minor detail - this could be a nice one liner

reana-ui/src/actions.js
audrium: +1 for changing it, but maybe we could create another task to change it sometime in the future as it's quite minor thing IMO

reana-ui/src/pages/workflowList/components/WorkflowActionsPopup.js
mvidalgarcia: Do you mean something like this?

image

vs

image

I agree on the opening downside, looks better IMO 🙏 Not sure about the horizontal dots.. Opinions @tiborsimko @diegodelemos @audrium

audrium: I'd give my vote for the vertical option :1st_place_medal:

reana-ui/src/pages/workflowList/components/WorkflowList.js
mvidalgarcia: Hyphens are not allowed in dot representation, that's why we have some mixture.. which I agree it's not very nice...

reana-ui/src/pages/workflowList/components/WorkflowActionsPopup.js
mvidalgarcia: I'm not sure, while it might help in these specific cases you mentioned, it can be tricky for other (if not implemented server-side) such as filtering. Let's see what others think @tiborsimko @diegodelemos @ParthS007

reana-ui/src/pages/workflowList/components/WorkflowDeleteModal.js
mvidalgarcia: Indeed, leftover from an older version when I was only using prop shorthands.

reana-ui/src/pages/workflowList/components/WorkflowActionsPopup.js
ParthS007: The reason for horizontal dots is that the `Menu` almost takes the space within that workflow box only and in the vertical, it is over the second workflow box as well, it will not create any problem but UI wise the horizontal one looks cleaner. Maybe, we can remove the `pointing` from menu since it is not aligning with the `dots` in both horizontal and vertical.

reana-ui/src/pages/workflowList/components/WorkflowList.js
ParthS007: okay, We can create another ticket to address this mixture to use only one of them. 👍

reana-ui/src/pages/workflowList/components/WorkflowActionsPopup.js
ParthS007: Yeah, I think different `status` for `workflow` and `workspace` sounds good. As we develop further, it will be useful for having different `status` for both. But we have to add in the server-side too if we decide to separate both.

reana-ui/src/actions.js
diegodelemos: > +1 for changing it, but maybe we could create another task to change it sometime in the future as it's quite minor thing IMO

+1

This is a good example to use our [Triage board](https://github.com/orgs/reanahub/projects/7). Creating an issue stemming from this comment, tagging it as refactor, and adding it to `Dev next/future` depending on the urgency. @ParthS007 do you want to take care of it? Not to forget to look for existing issues that might be related such as https://github.com/reanahub/reana-workflow-controller/issues/143.

reana-ui/src/pages/workflowList/components/WorkflowActionsPopup.js
diegodelemos: I think we can live with this for the time being as the logic is not so complex... introducing an "inconsistency" with the backend might bite us in the future. We could make it more idiomatic if we have a new property on worfklows which can be `isWorkspaceEmpty` (or some other better name), at the end that is going to do the same check as we have now `size == "0K"`....

reana-ui/src/actions.js
ParthS007: Yes, I can have a look at it. Not sure how many other places will be affected by this.

reana-ui/src/pages/workflowList/components/WorkflowList.js
mvidalgarcia: Addressed.

audrium (approved):

reana-ui/src/pages/workflowList/components/WorkflowActionsPopup.js
diegodelemos: For me it shows very detached from the button:
image

If we remove the `pointing` I would expect the box to be closer to the button, what do you think?

diegodelemos (approved):

reana-ui/src/pages/workflowList/components/WorkflowActionsPopup.js
mvidalgarcia: There's an [`offset`](https://react.semantic-ui.com/modules/popup/#usage-offset) prop to address this, but I cannot manage to make it work. If you manage please let me know.

PR #137, opened 2020-09-30
ui: implemented filtering in workflow list page

### LINKED ISSUE #132: ui: Implement filtering in workflow list page
Namely basic search, filter by workflow state and sorting by oldest/newest.

image

audrium:
closes #132

### DIFF
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -57,11 +57,13 @@ const USER_SIGNUP_URL = `${api}/api/register`;
 const USER_SIGNIN_URL = `${api}/api/login`;
 const USER_SIGNOUT_URL = `${api}/api/logout`;
 const USER_REQUEST_TOKEN_URL = `${api}/api/token`;
-const WORKFLOWS_URL = ({ page = 1, size }) => {
-  let url = `${api}/api/workflows`;
-  if (size) {
-    url += `?page=${page}&size=${size}`;
-  }
+const WORKFLOWS_URL = ({ page = 1, size, search, status, sortDir }) => {
+  //TODO: use query-string to format query params
+  let url = `${api}/api/workflows?`;
+  if (size) url += `page=${page}&size=${size}&`;
+  if (search) url += `search=${search}&`;
+  if (sortDir) url += `sort=${sortDir}&`;
+  if (!_.isEmpty(status)) url += `status=${status}`;
   return url;
 };
 const WORKFLOW_LOGS_URL = (id) => `${api}/api/workflows/${id}/logs`;
@@ -181,16 +183,19 @@ export function requestToken() {
   };
 }
 
-export function fetchWorkflows(pagination) {
+export function fetchWorkflows(pagination, search, status, sortDir) {
   return async (dispatch) => {
     dispatch({ type: WORKFLOWS_FETCH });
     return await axios
-      .get(WORKFLOWS_URL({ ...pagination }), { withCredentials: true })
+      .get(WORKFLOWS_URL({ ...pagination, search, status, sortDir }), {
+        withCredentials: true,
+      })
       .then((resp) =>
         dispatch({
           type: WORKFLOWS_RECEIVED,
           workflows: parseWorkflows(resp.data.items),
           total: resp.data.total,
+          userHasWorkflows: resp.data.user_has_workflows,
         })
       )
       .catch((err) => {
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -176,11 +176,19 @@ export function requestToken() {
   };
 }
 
-export function fetchWorkflows(pagination, search, status, sortDir) {
+export function fetchWorkflows(
+  pagination,
+  search,
+  status,
+  sort,
+  showLoader = true
+) {
   return async (dispatch) => {
-    dispatch({ type: WORKFLOWS_FETCH });
+    if (showLoader) {
+      dispatch({ type: WORKFLOWS_FETCH });
+    }
     return await axios
-      .get(WORKFLOWS_URL({ ...pagination, search, status, sortDir }), {
+      .get(WORKFLOWS_URL({ ...pagination, search, status, sort }), {
         withCredentials: true,
       })
       .then((resp) =>
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -12,7 +12,12 @@ import _ from "lodash";
 import axios from "axios";
 
 import { api } from "./config";
-import { parseWorkflows, parseLogs, parseFiles } from "./util";
+import {
+  parseWorkflows,
+  parseLogs,
+  parseFiles,
+  stringifyQueryParams,
+} from "./util";
 import {
   getWorkflow,
   getWorkflowLogs,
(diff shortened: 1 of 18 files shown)


mvidalgarcia:
One observation at first glance. I don't get the welcome page anymore when there no workflows.

<details>
<summary>See screenshot</summary>
image
</details>

audrium:
> One observation at first glance. I don't get the welcome page anymore when there are no workflows.
> See screenshot
> image

Good catch. It is fixed now

reana-ui/src/actions.js
```diff
@@ -57,11 +57,13 @@ const USER_SIGNUP_URL = `${api}/api/register`;
 const USER_SIGNIN_URL = `${api}/api/login`;
 const USER_SIGNOUT_URL = `${api}/api/logout`;
 const USER_REQUEST_TOKEN_URL = `${api}/api/token`;
-const WORKFLOWS_URL = ({ page = 1, size }) => {
-  let url = `${api}/api/workflows`;
-  if (size) {
-    url += `?page=${page}&size=${size}`;
-  }
+const WORKFLOWS_URL = ({ page = 1, size, search, status, sortDir }) => {
+  //TODO: use query-string to format query params
```
mvidalgarcia, line 61: Maybe we can include this dependency already? It shouldn't be a lot of work, should it?
If it's too much we can ticketize and leave it for later.

audrium: Yeah, I think we can do it now. I left it as a todo item just to make a leaner PR. I will implement it asap

mvidalgarcia: If it's a separate commit it should be fine.

reana-ui/src/pages/workflowList/components/WorkflowFilters.js
```diff
@@ -0,0 +1,40 @@
...
+import WorkflowStatusFilter from "./WorkflowStatusFilter";
+import WorkflowSorting from "./WorkflowSorting";
+
+import styles from "./WorkflowFilters.module.scss";
+
+export default function WorkflowFilters({
+  statusFilter,
+  setStatusFilter,
+  sortDir,
+  setSortDir,
```
mvidalgarcia, line 23: Can we have `PropTypes` for these new components?

reana-ui/src/pages/workflowList/components/WorkflowSearch.js
```diff
@@ -0,0 +1,28 @@
...
+const TYPING_DELAY = 1000;
+
+export default function WorkflowSearch({ search }) {
+  const handleChange = debounce(search, TYPING_DELAY);
+  return (
+    <Input
+      fluid
+      icon="search"
+      placeholder="Search..."
+      onChange={(event, data) => handleChange(data.value)}
```
mvidalgarcia, line 25:
```suggestion
      onChange={(_, data) => handleChange(data.value)}
```

reana-ui/src/pages/workflowList/components/WorkflowStatusFilter.js
```diff
@@ -0,0 +1,35 @@
...
+
+export default function WorkflowStatusFilters({ valueList, filter }) {
+  return (
+    <Dropdown
+      text="Status"
+      closeOnChange
+      selection
+      multiple
+      options={statusOptions}
+      onChange={(event, data) => filter(data.value)}
```
mvidalgarcia, line 31:
```suggestion
      onChange={(_, data) => filter(data.value)}
```

mvidalgarcia (commented):
Great job! 👌 Looks good overall. Some remarks:

- We've introduced the loader indicator when fetching workflows, this looks good when applying filters and performing searches. However, it's also triggered when the polling happens every 15 secs, which looks kind of weird.. so I'm not sure about having the loader anymore (This will be solved when we have our greatly desired WebSockets as a reload would be new data), WDTY?

- When loading the workflow list for the first time, without applying any filter or search, the polling works as usual. However, after setting some filters, the polling stops working. This might be related to multiple `setTimeout`s in place conflicting (I think I faced I similar issue when implementing the pagination). [Check this](https://github.com/reanahub/reana-ui/blob/master/reana-ui/src/pages/workflowList/WorkflowList.js#L91-L94), it may help.

reana-ui/src/actions.js
```diff
@@ -181,16 +176,19 @@ export function requestToken() {
   };
 }
 
-export function fetchWorkflows(pagination) {
+export function fetchWorkflows(pagination, search, status, sortDir) {
   return async (dispatch) => {
     dispatch({ type: WORKFLOWS_FETCH });
     return await axios
-      .get(WORKFLOWS_URL({ ...pagination }), { withCredentials: true })
+      .get(WORKFLOWS_URL({ ...pagination, search, status, sortDir }), {
```
mvidalgarcia, line 183: ⚠️ the query param in the server is called `sort` instead of `sortDir`!

reana-ui/src/pages/workflowList/WorkflowList.js
```diff
@@ -43,28 +47,51 @@ function Workflows() {
...
+    reanaToken,
+    searchFilter,
+    statusFilter,
+    sortDir,
+  ]);
+
+  const applyFilter = (filter) => (value) => {
+    unstable_batchedUpdates(() => {
+      filter(value);
+      setPagination({ ...pagination, page: 1 });
```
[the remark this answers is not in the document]
audrium: Yes, exactly, I think we should always go back to page 1 after some filters are applied. Because there could be too few filtered results to paginate. Before I implemented this change it was very weird to use it from the UX perspective.

reana-ui/src/pages/workflowList/components/WorkflowSearch.js
audrium: I'm not sure that using underscore for unused parameters are the best idea. I mean it's nice convention to indicate it, but in the same time we loose the information about what data it is (in case we will need it in the future). Also, it might be confusing when there are multiple unused parameters. What do you think about naming it as `_event`?

audrium:
>     * We've introduced the loader indicator when fetching workflows, this looks good when applying filters and performing searches. However, it's also triggered when the polling happens every 15 secs, which looks kind of weird.. so I'm not sure about having the loader anymore (This will be solved when we have our greatly desired WebSockets as a reload would be new data), WDTY?

Yep, I agree that it's nice to have loading indicator after filtering, but for polling it's not that great. Maybe we can disable dispatching of `WORKFLOWS_FETCH` action when polling - this will not show the loading spinner in that case. It could be a temp solution until WebSockets will be implemented

reana-ui/src/pages/workflowList/components/WorkflowSearch.js
mvidalgarcia: In these kind of events it's kind of a convention that `event` is always the first arg. In any case, don't worry too much, it's not very important, it's more about avoiding the eslint warning of "Unused variable" more than other thing 😅

audrium:
>     * When loading the workflow list for the first time, without applying any filter or search, the polling works as usual. However, after setting some filters, the polling stops working. This might be related to multiple `setTimeout`s in place conflicting (I think I faced I similar issue when implementing the pagination). [Check this](https://github.com/reanahub/reana-ui/blob/master/reana-ui/src/pages/workflowList/WorkflowList.js#L91-L94), it may help.

It was indeed the case, nice that you spotted. `interval.current` was not set to null on polling useEffect: [fixed here](https://github.com/reanahub/reana-ui/blob/8f9f4895094e9be2873588c05d294e519dcc34aa/reana-ui/src/pages/workflowList/WorkflowList.js#L95)

reana-ui/src/pages/workflowList/WorkflowList.js
```diff
@@ -66,16 +66,20 @@ function Workflows() {
...
       interval.current = setInterval(() => {
+        const showLoader = false;
         dispatch(
-          fetchWorkflows({ ...pagination }, searchFilter, statusFilter, sortDir)
+          fetchWorkflows(
+            { ...pagination },
+            searchFilter,
+            statusFilter,
+            sortDir,
+            showLoader
```
mvidalgarcia, line 76: I think we can pass directly `showLoader=false` here.

mvidalgarcia (approved):
> It was indeed the case, nice that you spotted. interval.current was not set to null on polling useEffect: fixed here

Happy to read that, it's working fine now 🤟

reana-ui/src/pages/workflowList/WorkflowList.js
audrium: Yeah, had some doubts about this.. Initially I passed it directly, but it kind of looks weird when you read the code. False value does not say much about what is happening, that's why added the variable. Don't know, we can remove it

mvidalgarcia: Yeah definitely, what I mean is that you can pass the name of the variable as well. Like this:

```js
fetchWorkflows(
  { ...pagination },
  searchFilter,
  statusFilter,
  sortDir,
  showLoader=false
)
```

Correct me if I'm wrong, but I think it's allowed.

audrium: Hm, it's not allowed (yet ;D). Maybe you are confusing it with [default parameters](http://es6-features.org/#DefaultParameterValues)?. Or another option would be to wrap everything into object and pass to `fetchWorkflows`.. Maybe let's leave it for now

mvidalgarcia: 🤔 Maybe it's a new feature...

image

Nevermind, it's not important.

PR #141, opened 2020-10-13
global: fix fypo

mvidalgarcia:
closes reanahub/reana#422


tiborsimko (approved):
Works well; `r-commons` tests to be fixed before merge

PR #146, opened 2020-10-22
global: display sizes in human readable across UI

diegodelemos:
* Closes reanahub/reana-server#296.

### DIFF
--- a/reana-ui/src/components/WorkflowActionsPopup.js
+++ b/reana-ui/src/components/WorkflowActionsPopup.js
@@ -33,7 +33,7 @@ export default function WorkflowActionsPopup({ workflow, className }) {
   const [open, setOpen] = useState(false);
   const { id, size, status, session_status: sessionStatus } = workflow;
   const isDeleted = status === "deleted";
-  const isDeletedUsingWorkspace = isDeleted && size !== "0K";
+  const isDeletedUsingWorkspace = isDeleted && size.raw;
   const isSessionOpen = sessionStatus === "created";
 
   let menuItems = [];
--- a/reana-ui/src/pages/workflowList/components/WorkflowList.js
+++ b/reana-ui/src/pages/workflowList/components/WorkflowList.js
@@ -51,7 +51,8 @@ export default function WorkflowList({ workflows, loading }) {
           session_status: sessionStatus,
         } = workflow;
         const isDeleted = status === "deleted";
-        const isDeletedUsingWorkspace = isDeleted && size !== "0K";
+        const hasDiskUsage = size.raw > 0;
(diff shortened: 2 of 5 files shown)


reana-ui/src/components/WorkflowActionsPopup.js
```diff
@@ -33,7 +33,7 @@ export default function WorkflowActionsPopup({ workflow, className }) {
   const [open, setOpen] = useState(false);
   const { id, size, status, session_status: sessionStatus } = workflow;
   const isDeleted = status === "deleted";
-  const isDeletedUsingWorkspace = isDeleted && size !== "0K";
+  const isDeletedUsingWorkspace = isDeleted && size.raw !== 0;
```
mvidalgarcia, line 36: I guess something like this is enough:

```suggestion
  const isDeletedUsingWorkspace = isDeleted && size.raw;
```

reana-ui/src/pages/workflowList/components/WorkflowList.js
```diff
@@ -51,7 +51,7 @@ export default function WorkflowList({ workflows, loading }) {
           session_status: sessionStatus,
         } = workflow;
         const isDeleted = status === "deleted";
-        const isDeletedUsingWorkspace = isDeleted && size !== "0K";
+        const isDeletedUsingWorkspace = isDeleted && size.raw !== 0;
```
mvidalgarcia, line 54: Same

mvidalgarcia (approved):

reana-ui/src/pages/workflowList/components/WorkflowList.js
```diff
@@ -74,7 +74,7 @@ export default function WorkflowList({ workflows, loading }) {
                 }`}
               >
                 <Icon name="hdd" />
-                {size}
+                {size.human_readable}
```
mvidalgarcia, line 77: I see that while the workflow is running we still don't have its size but the icon shows anyway.

image

WDYT about hiding the icon if we don't have the human-readable value?

mvidalgarcia: Also, nor necessarily belonging to this PR, but we should display the size on the details page as well, we're not doing it at the moment. This comment can be converted into an issue if you find this appropriate.

diegodelemos: > I see that while the workflow is running we still don't have its size but the icon shows anyway.

Fixed, thanks!

> Also, not necessarily belonging to this PR, but we should display the size on the details page as well, we're not doing it at the moment. This comment can be converted into an issue if you find this appropriate.

Good catch! Let's do this in another issue, just to not delay this PR group more.

PR #149, opened 2020-10-27
ci: added github actions workflow

audrium:
Here is an example of GH actions workflow: https://github.com/audrium/reana-ui/actions/runs/331829733


ParthS007 (approved):
LGTM

PR #156, opened 2020-11-16
actions: improve error handling

### LINKED ISSUE #155: ui: better exception handling
> The `reana-ui issue` on `0.7.0devYYYYMMDD` deployed on PROD that I mentioned at the stand-up today was the following:

>  - `reana-server` was not returning a response for a reason or another;
> - `reana-client` reported correctly an error in CLI;
> - ... but `reana-ui` showed default homepage, very fast, as if the user did not launch any workflows at all so far.

> The database was OK and the system came back to life after killing `r-server` and `r-w-controller`.

> Verify `reana-ui` exception handling on `maint-0.7` and see what happens with the web interface when `r-server` misbehaves, e.g. when it is not contactable?  It would be better if the front page "keep spinning", or shows an error, rather than the default "welcome" page.

_Originally reported by @tiborsimko on Mattermost_

mvidalgarcia:
- Avoids infinite loop of `fetchWorkflows` requests if config endpoint misbehaves (pollingSecs).
- Avoids displaying welcome page when an exception happens on server.
- Gets rid of all `fetch()` calls, replacing them with `axios` calls.

partially addresses #46
closes #155

### DIFF
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -98,48 +98,34 @@ export function loadConfig() {
 
 export function loadUser() {
   return async (dispatch) => {
-    let resp, data;
-    try {
-      dispatch({ type: USER_FETCH });
-      resp = await fetch(USER_INFO_URL, { credentials: "include" });
-    } catch (err) {
-      throw new Error(USER_INFO_URL, 0, err);
-    }
-    if (resp.status === 403) {
-      data = await resp.json();
-      dispatch({ type: USER_FETCH_ERROR, ...data });
-    } else if (resp.ok) {
-      data = await resp.json();
-    }
-    dispatch({ type: USER_RECEIVED, ...data });
-    return resp;
+    dispatch({ type: USER_FETCH });
+    return await axios
+      .get(USER_INFO_URL, { withCredentials: true })
+      .then((resp) => dispatch({ type: USER_RECEIVED, ...resp.data }))
+      .catch((err) => {
+        dispatch(errorActionCreator(err, USER_INFO_URL));
+        dispatch({ type: USER_FETCH_ERROR });
+      });
   };
 }
 
 function userSignFactory(initAction, succeedAction, actionURL, body) {
   return async (dispatch) => {
-    let resp, data;
-    try {
-      dispatch({ type: initAction });
-      resp = await fetch(actionURL, {
-        method: "POST",
-        headers: {
-          "Content-Type": "application/json",
-        },
-        body: JSON.stringify(body),
-        credentials: "include",
+    dispatch({ type: initAction });
+    return await axios
+      .post(actionURL, body, {
+        withCredentials: true,
+        headers: { "Content-Type": "application/json" },
+      })
+      .then((resp) => {
+        dispatch({ type: succeedAction });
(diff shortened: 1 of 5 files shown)


reana-ui/src/actions.js
```diff
@@ -151,20 +137,18 @@ export const userSignin = (formData) =>
...
 export function userSignout() {
   return async (dispatch) => {
-    let resp;
-    try {
-      dispatch({ type: USER_SIGNOUT });
-      resp = await fetch(USER_SIGNOUT_URL, {
-        method: "POST",
-        credentials: "include",
+    dispatch({ type: USER_SIGNOUT });
+
```
audrium, line 141: Minor nitpick: this empty line could be removed

reana-ui/src/pages/workflowList/WorkflowList.js
```diff
@@ -66,7 +66,7 @@ function Workflows() {
     };
   }, [config.pollingSecs, dispatch, pagination, reanaToken]);
 
-  if (!workflows) {
+  if (!workflows || !config.pollingSecs) {
```
audrium, line 69: Minor suggestions: this logic could be moved to `selectors` just to be more declarative

audrium (approved):
Left couple of minor comments, but in overall it looks good to me! Much much better now compared with plain `fetch`!

reana-ui/src/pages/workflowList/WorkflowList.js
mvidalgarcia: Do you mean something like this?

```js
export const getWorkflows = (state) =>
  !!state.config.pollingSecs && state.workflows.workflows;
```

If yes.. I'm not sure.. for now we're only using getWorkflows in a single place, but if we start using it somewhere else it will always depend on the existence of `pollingSecs`, which I'm not sure if it's correct.

audrium: Actually my initial thought was to move `!workflows || !config.pollingSecs` to selectors, but probably it's not the right place to keep it. Maybe we can create a new variable in the top of the component, something like `const showWelcomePage = !workflows || !config.pollingSecs;` and use this while rendering. It could increase readability a bit

mvidalgarcia: Ah ok, I get your point now. Unfortunately, the polling is making our lives harder and harder. I'll create a variable as you suggest.

mvidalgarcia: I've created a new selector called `isConfigLoaded` to be more accurate, instead of just checking `pollingSecs`. Let me know what you think. cc/ @audrium

audrium: I like how it's implemented now! Just a thought: we could modify `config` reducer to hold a flag `isLoaded` or `isFetched` and use that flag in the `isConfigLoaded` `selector`. What do you think?

mvidalgarcia: That's indeed better than comparing objects, updated!

audrium: LGTM!

PR #157, opened 2020-11-17
Merge branch 'maint-0.7'

mvidalgarcia:


reana-ui/src/actions.js
```diff
@@ -122,51 +122,45 @@ export function loadConfig() {
...
+    dispatch({ type: USER_FETCH });
+    dispatch({ type: QUOTA_FETCH });
+    return await axios
+      .get(USER_INFO_URL, { withCredentials: true })
+      .then((resp) => {
+        dispatch({ type: USER_RECEIVED, ...resp.data });
+        dispatch({ type: QUOTA_RECEIVED, ...resp.data });
+      })
+      .catch((err) => {
+        // 403 Forbidden, user token was revoked.
```
audrium, line 134: In `maint-0.7` we had a simple catch:
```
      .catch((err) => {
        dispatch(errorActionCreator(err, USER_INFO_URL));
        dispatch({ type: USER_FETCH_ERROR });
      });
```
Why do we want to fail silently on all other errors except `403`?

Wouldn't it be better to explicitly handle `401`? And treat all other errors in the same way?
Something like:
```
      .catch((err) => {
        // 403 Forbidden, user token was revoked.
        // 401 Unauthorized, user did not sign in, we fail silently.
        const { status, data } = err?.response;
        if (status !== 401) {
          dispatch(errorActionCreator(err, USER_INFO_URL));
          dispatch({ type: USER_FETCH_ERROR, ...data });
          dispatch({ type: QUOTA_FETCH_ERROR, ...data });
        }

      });
```

mvidalgarcia: The problem is that we still need to trigger the `USER_FETCH_ERROR` to stop the loading spinner (untested, talking from memory). In your example, it would fail silently on 401, but the spinner would remain.

https://github.com/reanahub/reana-ui/blob/42a71801147d0e6fca8e5356bced96ae438000b0/reana-ui/src/reducers.js#L153-L158

audrium: I get your point now. Maybe then something like this could be used:
```
      .catch((err) => {
        // 403 Forbidden, user token was revoked.
        // 401 Unauthorized, user did not sign in, we fail silently.
        const { status, data } = err?.response;
        if (status === 401) {
          dispatch({ type: USER_FETCH_ERROR });
          dispatch({ type: QUOTA_FETCH_ERROR });
          return err;
        }
        dispatch(errorActionCreator(err, USER_INFO_URL));
        dispatch({ type: USER_FETCH_ERROR, ...data });
        dispatch({ type: QUOTA_FETCH_ERROR, ...data });
      });
```

mvidalgarcia: Ok, this looks similar to what I have. But you're right and maybe the other errors (e.g 500) should trigger the `errorActionCreator`. What do you think about keeping what I have but changing the check to this?

```js
if (err.response.status !== 401)
```

audrium: That indeed would be a nice approach! In general I guess it's better to avoid `let` and mutations, but in this case it would save us from repetitive `dispatching` of the same action

audrium (approved):

PR #158, opened 2020-11-17
ci: add hadolint linter

audrium:
addresses https://github.com/reanahub/reana/issues/450

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ on: [push, pull_request]
 
 jobs:
   lint-shellcheck:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - name: Checkout
         uses: actions/checkout@v2
@@ -21,7 +21,7 @@ jobs:
           ./run-tests.sh --check-shellscript
 
   docs-sphinx:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     steps:
       - name: Checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -82,5 +82,8 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v2
 
+      - name: Lint Dockerfile
```
tiborsimko, line 85: (1) Any particular reason for doing Dockerfile linting as part of `docker-build` check?  IIRC elsewhere in `reana-env-root6` we had a separate `lint-docker` check... It would seem more coherent with Python linting (where it is separate) to have a separate test for Docker linting? Perhaps named  `lint-docker`?

(2) If we do this check only in Github CI, and not in `run-tests.sh`, then running `run-tests.sh` locally won't be running all the tests, breaking symmetry running-locally vs running-on-github. IOW, it could happen that tests pass locally but not on GitHub CI. What about adding a test to `run-tests.sh --check-docker`?  Assuming that all developers have `hadolint` installed locally, the symmetry would be back, and if tests pass locally, the developers can be assured that they will also pass on GitHub CI.

tiborsimko: Hmm, I did something similar while installing XRootD CLI for `cernopendata-client` image, so perhaps we could muse about and synchronise on this live? https://github.com/cernopendata/cernopendata-client/pull/95/files

audrium: Good points, I will update the PRs accordingly

tiborsimko (approved):

PR #168, opened 2020-12-15
ui: split Signin component into two separate components

audrium:
addresses https://github.com/reanahub/reana/issues/469

Split `Signin` component into two separate components:
- `Signin` 
- `Signup`

Added `SignWrapper` HOC which injects form submission logic
Added `SignContainer` which wraps both new components to render necessary data


reana-ui/src/pages/signin/SignWrapper.js
```diff
@@ -0,0 +1,43 @@
...
+
+import React from "react";
+import { useDispatch, useSelector } from "react-redux";
+import { useHistory, useLocation } from "react-router-dom";
+
+import { getConfig } from "../../selectors";
+import SignContainer from "./components/SignContainer";
+
+export default function SignWrapper(WrappedComponent) {
+  return function SignWrapperHoc() {
```
mvidalgarcia, line 19: Opinion: I'm not a big fan of HOC, although I understand that in this particular case can make sense as you're not injecting new props to the component, and I think that the wrapped components can also work standalone. More about this [here](https://medium.com/javascript-scene/do-react-hooks-replace-higher-order-components-hocs-7ae4a08b7b58). 

Have you considered replicating the same with [custom hooks](https://reactjs.org/docs/hooks-custom.html)?
I think that something like `useSign()` that returns `handleSubmit` function can work similarly.

This is not required at all, only in case you wanted to give it a try. I'd like to see some custom hooks in our app at some point.

mvidalgarcia (commented):
I still think both forms look pretty similar, what about adding a title on top of the form itself? e.g.

mvidalgarcia:
Ah, I forgot, could you please cherry-pick [this commit](https://github.com/reanahub/reana-ui/commit/da90de1ff8a105bcbca8fe3fb63bc8e37cacffef) so we avoid the "User not logged in" error notification?

reana-ui/src/pages/signin/SignWrapper.js
audrium: Thanks a lot for this suggestion! Indeed it looks much cleaner, no need to create unnecessary layers of components in this case. Custom hooks didn't came into my mind somehow. I've submitted my changes in additional commit so that it would be easier for you to see the changes. Later I can squash it

reana-ui/src/pages/signin/Signin.js
```diff
@@ -9,110 +9,66 @@
...
+      )}
+    </>
   );
 }
 
 Signin.propTypes = {
-  /** Whether to show signup instead of signin form. */
-  signup: PropTypes.bool,
+  handleSubmit: PropTypes.func.isRequired,
+  config: PropTypes.object.isRequired,
```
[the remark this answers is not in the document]
audrium: The reason was to save some lines of code while importing `getConfig, useSelector` and calling the selector to get the configs, but with custom hooks it was brought back to the components. Now it's is more explicit

reana-ui/src/pages/signin/SignWrapper.js
mvidalgarcia: Looks amazing, so happy to see our first custom hook! 😇
I wonder if it would make sense to have a dedicated file/folder called `hooks` to store them? WDYT?
I see them as helpers/utils that can be reused in different components so it might make sense. Not sure what's the common practice in this case.

mvidalgarcia (approved):
LGTM, we can address the aesthetic comments in other tasks!

reana-ui/src/pages/signin/SignWrapper.js
audrium: Totally agree, I had a hard time figuring out where to put it. I moved it to the top level of the project file called `hooks` for now. Later once we have lots of them we can refactor it easily into folder with files.
Actually this hook seems very specific for Sign in/up components, not sure if we will be able to reuse it somehow, but still it's quite nice to have all the hooks in one place!

PR #169, opened 2020-12-17
ui: control signup form visibility with configs

audrium:
addresses https://github.com/reanahub/reana/issues/469

### DIFF
--- a/reana-ui/src/pages/signin/Signup.js
+++ b/reana-ui/src/pages/signin/Signup.js
@@ -43,7 +43,7 @@ export default function Signup() {
       </Segment>
       {config.localUsers && (
         <p>
-          Already signed up? Go to <Link to="/signin">Sign In</Link>
+          Already signed up? Go to <Link to="/signin">Sign in</Link>
         </p>
       )}
     </SignContainer>
--- a/reana-ui/src/components/App.js
+++ b/reana-ui/src/components/App.js
@@ -14,7 +14,13 @@ import { Redirect, BrowserRouter, Route, Switch } from "react-router-dom";
 import { useSelector } from "react-redux";
 import { Dimmer, Loader } from "semantic-ui-react";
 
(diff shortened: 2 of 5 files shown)


reana-ui/src/pages/signin/Signup.js
```diff
@@ -31,21 +32,17 @@ export default function Signup() {
...
           />
-        )}
-      </Segment>
-      {config.localUsers && (
-        <p>
-          Already signed up? Go to <Link to="/signin">Sign In</Link>
-        </p>
+        </Segment>
       )}
+      <SignFormInfo config={config} signup />
```
mvidalgarcia, line 45: I agree that `SignFormInfo` component looks a bit convoluted and prints completely different things. What if we keep `Signup` component as it was and [disable the `/singup` route](https://github.com/reanahub/reana-ui/blob/a551744041dfb66a1863ded170e49ff96887502d/reana-ui/src/components/App.js#L64-L67) if `hideSignup` is true?
Then we can have the logic of `SingFormInfo` inside `Signin` component.

audrium: This approach looks cleaner, but there is one downside. Currently in the `App.js` [we are waiting for the /api/you response](https://github.com/reanahub/reana-ui/blob/master/reana-ui/src/components/App.js#L45) before rendering any pages. If we want to disable the route in this place we will have to wait for `/api/config` to arrive, since `hideSignup` is stored there. So +1 request to wait for the user before seeing some content. Not sure if this is a big problem, probably noticeable only on a very slow connection, but still something to think about. Because of this reason I took another approach which is not very clean.. Could be that sooner or later we will still have to depend on the configs to render or not some component or page.

audrium: After playing around a bit, I think it's still better to disable the route, so I changed the code according to your suggestions

diegodelemos (approved):

PR #182, opened 2021-05-04
footer: remove copyright and move ropo to left

### LINKED ISSUE #178: footer: dynamic copyright years
The footer has hardcoded copyright year:

```console
$ cat reana-ui/src/components/Footer.js
...
       <span>Copyright © 2020 CERN</span>
```

This should rather read 2017-2021, i.e. the beginning of the RENAA project is 2017, and ending with the current year.

(Similarly to CERN Open Data.)

Alternatively, we can consider removing displaying copyright in the footer, since the value is not clear: the site is mostly for CERN internal use, and many sites at CERN do not display copyright at the bottom. (See also Google, DuckDuckGo, etc.) So having copyright in the footer is less important for REANA than it is for CERN Open Data, say.

### LINKED ISSUE #180: footer: missing Mattermost chat link and RoPO position
At CERN deployment, the Mattermost chat link is missing in the footer:

chat

This can be amended via Helm deployment values regarding chat URL, but it also raises a question that it may be perhaps more appropriate to move the RoPO link next to copyright (see also #178), so that the left-hand-side footer section will become a "legal corner"  (displaying only copyright and privacy notice link), and the right-hand side footer section will become "user help" corner (displaying chat, forum and docs links).

mvidalgarcia:
closes #178
closes #180


audrium:
LGTM :rocket: 
Old footer vs new one:
Screenshot from 2021-05-04 13-33-13

audrium (approved):

PR #185, opened 2021-05-12
ui: add cluster health status page

### LINKED ISSUE #183: ui: prototype new page to show users REANA cluster status
In the context of the [Big multi-user workflows](https://github.com/orgs/reanahub/projects/37) sprint, we will come up with more dynamic ways of deciding whether a workflow should be started or not.

This will make users wonder why their workflows take long to be queued. This new page is intended to give users information about whether the cluster is busy, so they can relate to why their workflows take long to be scheduled.

Steps:
- Serve information to the UI (information should be provided by https://github.com/reanahub/reana/issues/501)
  - Node health: percentage of nodes up (unschedulable vs total). Pie chart.
  - Workflow health: percentage of running vs pending/queued workflows. Pie chart.
  - Job health: percentage of running vs pending/queued jobs. Pie chart.
- Mockup the new page
  - To take inspiration from other systems such as Pachyderm, Kubeflow etc...
- Implement mockup and connect to backend

Note:
- We already have a [`status.py` in RS](https://github.com/reanahub/reana-server/blob/master/reana_server/status.py) which could be extended for this case
- This should be easily extensible since new measures can come up while progressing in other tasks in the sprint.

mvidalgarcia:
closes #183

##### TODO:
- [ ] Sessions (Jupyter notebooks) icon instead of Pie Chart.
- [ ] Test multi-node cluster setup to force `unschedulable` cases.

image

### DIFF
--- a/reana-ui/src/pages/status/Status.js
+++ b/reana-ui/src/pages/status/Status.js
@@ -1,0 +1,135 @@
+/*
+  -*- coding: utf-8 -*-
+
+  This file is part of REANA.
+  Copyright (C) 2021 CERN.
+
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+import { useEffect, useState } from "react";
+import { useDispatch } from "react-redux";
+import { Container, Grid, Icon, Label, Loader } from "semantic-ui-react";
+import axios from "axios";
+
+import BasePage from "../BasePage";
+import { errorActionCreator } from "~/actions";
+import { Title, PieChart } from "~/components";
+import { api } from "~/config";
+import { healthMapping } from "~/util";
+
+import styles from "./Status.module.scss";
+
+export default function Status() {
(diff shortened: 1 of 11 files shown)


reana-ui/src/pages/status/Status.js
```diff
@@ -0,0 +1,135 @@
...
+  };
+
+  return (
+    <BasePage>
+      <Container text className={styles.container}>
+        <Title>
+          Cluster Health
+          <Icon name="heartbeat" />
+        </Title>
+        {loading || !status ? (
```
audrium, line 119: Minor: right now if something goes wrong with the request (500 for example), we still show that the cluster is loading. We could provide something went wrong message for the user, WDYT?

audrium: ah, I see you now dispatch `errorActionCreator` on error, nice. I got a bit outdated version locally..

audrium (approved):
Nicely done, works great!

PR #197, opened 2021-10-28
profile: hide quota section when disabled

### LINKED ISSUE #198: wf-details: error and infinite spinner when wf workspace is deleted
Steps to reproduce:
1. Submit a workflow
2. Delete it including workspace (either from UI or r-client)
3. Access workflow details
4. Click on workspace tab
image

The message is accurate but we shouldn't show an infinite spinner as if it was loading something, and probably it shouldn't be displayed as an error.

mvidalgarcia:
closes reanahub/reana#562
closes #198

### DIFF
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -312,7 +312,17 @@ export function fetchWorkflowFiles(id, pagination) {
         })
       )
       .catch((err) => {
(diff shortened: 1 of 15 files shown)


audrium (approved):
Works nicely! I haven't spotted any issues.
Great job with `client.js`, finally we have it!

PR #199, opened 2021-11-01
details: display message when workspace is deleted

### LINKED ISSUE #198: wf-details: error and infinite spinner when wf workspace is deleted

mvidalgarcia:
- Better identify when a workflow workspace was deleted by checking the HTTP code

closes #198

image

### DIFF
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -345,7 +345,17 @@ export function fetchWorkflowFiles(id, pagination) {
         })
       )
       .catch((err) => {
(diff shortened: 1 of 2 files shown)


mvidalgarcia:
Commit included as part of [this PR](https://github.com/reanahub/reana-ui/pull/197). Closing.

PR #212, opened 2022-01-12
workspace: add file search

### LINKED ISSUE #204: package: bump all module versions
Many vulnerabilities found: https://github.com/reanahub/reana-ui/security/dependabot

Two dependabot PRs
- https://github.com/reanahub/reana-ui/pull/202
- https://github.com/reanahub/reana-ui/pull/203

### LINKED ISSUE #211: ui: workspace file filtering
**Current behaviour**

Some workflows can produce many files, e.g. `reana-demo-cms-h4l`:

```console
$ reana-client ls -w h4l | wc -l
840
```

The CLI allows filtering only files I am interested in:

```console
$ reana-client ls -w h4l --filter name=results
NAME                                           SIZE    LAST-MODIFIED
results/Higgs4L1file.root                      59770   2021-12-07T01:08:44
results/DoubleMuParked2012C_10000_Higgs.root   54181   2021-12-07T01:08:54
results/scramdone.txt                          0       2021-12-07T01:05:44
results/mass4l_combine_userlvl3.pdf            18141   2021-12-07T01:11:04
```

The web UI currently does not, and offers 56 paginated pages:

reana-web-ui-h4l

Therefore it may take a lot of time to locate desired file in the web UI if one does not use CLI in thiese kinds of situations.

**Expected behaviour**

It'll be useful to offer a text box where a user could type part of the file name and the web interface would output only those files that match the input pattern.  Similarly to how list of workflows can be filtered:

reana-web-ui-search

Obviously, with an adapted UX design to fit the Workspace page style.

mvidalgarcia:
closes #211
closes #204

https://user-images.githubusercontent.com/8863238/149172123-d18f801d-8f11-4858-acd8-c8b019efdf14.mov

**Test:**
- To be tested thoroughly rebuilding the REANA-UI docker image as it involves a change in the Node.js version.

### DIFF
--- a/reana-ui/package.json
+++ b/reana-ui/package.json
@@ -3,7 +3,7 @@
   "version": "0.8.0",
   "private": true,
   "dependencies": {
-    "axios": "latest",
+    "axios": "^0.24.0",
     "lodash": "^4.17.15",
     "mime": "^2.4.4",
     "moment": "^2.24.0",
--- a/reana-ui/package.json
+++ b/reana-ui/package.json
@@ -7,7 +7,7 @@
     "lodash": "^4.17.15",
     "mime": "^2.4.4",
     "moment": "^2.24.0",
-    "node-sass": "^4.13.0",
+    "node-sass": "^6.0.0",
     "prop-types": "^15.7.2",
     "query-string": "^6.13.4",
     "react": "^17.0.2",
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -2,7 +2,7 @@
   -*- coding: utf-8 -*-
 
   This file is part of REANA.
-  Copyright (C) 2020, 2021 CERN.
+  Copyright (C) 2020, 2021, 2022 CERN.
 
   REANA is free software; you can redistribute it and/or modify it
   under the terms of the MIT License; see LICENSE file for more details.
@@ -21,7 +21,7 @@ import client, {
   INTERACTIVE_SESSIONS_OPEN_URL,
   INTERACTIVE_SESSIONS_CLOSE_URL,
 } from "~/client";
-import { parseWorkflows, parseLogs, parseFiles } from "~/util";
+import { parseWorkflows, parseLogs, parseFiles, formatSearch } from "~/util";
 import {
   getConfig,
   getWorkflow,
@@ -243,9 +243,8 @@ export function fetchWorkflows(
     if (showLoader) {
       dispatch({ type: WORKFLOWS_FETCH });
     }
-    const nameSearch = search ? JSON.stringify({ name: [search] }) : search;
     return await client
-      .getWorkflows(pagination, nameSearch, status, sort)
+      .getWorkflows(pagination, formatSearch(search), status, sort)
       .then((resp) =>
         dispatch({
           type: WORKFLOWS_RECEIVED,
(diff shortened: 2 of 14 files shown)


reana-ui/src/actions.js
```diff
@@ -298,11 +298,12 @@ export function fetchWorkflowLogs(id) {
   };
 }
 
-export function fetchWorkflowFiles(id, pagination) {
+export function fetchWorkflowFiles(id, pagination, search) {
   return async (dispatch) => {
     dispatch({ type: WORKFLOW_FILES_FETCH });
+    const nameSearch = search ? JSON.stringify({ name: [search] }) : search;
```
audrium, line 304: Very minor suggestion: since we have the same logic in line 246, we could extract it into a small function

reana-ui/src/components/Search.js
```diff
@@ -2,20 +2,20 @@
   -*- coding: utf-8 -*-
 
   This file is part of REANA.
-  Copyright (C) 2020 CERN.
+  Copyright (C) 2020, 2021, 2022 CERN.
```
audrium, line 5: Very minor: copyright year was not updated in most of the other touched files

reana-ui/package.json
```diff
@@ -3,11 +3,11 @@
   "version": "0.8.0",
   "private": true,
   "dependencies": {
-    "axios": "latest",
+    "axios": "^0.24.0",
```
audrium, line 6: Nice to see this! :+1:

audrium (approved):
Nice work, everything works as expected without any issues!

- I think we could update the `readme` with `yarn install` instead of `npm install`. Since we have `yarn.lock` and `package-lock.json` it could be confusing which one we should use?
- Search bar placeholder text is hard to read since the color is barely different from the background if not focused. If it's the same for you, we could create a mini issue and to fix it later?

reana-ui/src/components/Search.js
mvidalgarcia: Updated, thanks!

Did you manage to find a nice way to keep this updated in VSCode?

audrium: Not yet, is seems there are no extensions which would do exactly this task in VSCode unfortunately. However there are extensions like 'Run on Save' which could trigger custom task like this after saving a file, but I haven't tried to implement it.

audrium:
> Mmm it's true that is very light but I still can see it. This is the default color, check SUI-React Input. I've created a new commit making it a bit darker, let me know what you think.

That was quick! Thanks, it looks way better now I think

reana-ui/src/util.js
```diff
@@ -125,6 +125,15 @@ export function parseFiles(files) {
   return files;
 }
 
+/**
+ * Parses search input term.
+ * @param {String} term Search term
+ * @returns term format expected by the API.
+ */
+export function parseSearch(term) {
```
audrium, line 133: Minor question about the naming. Does this function really `parses` search input term? As I understand it produces a query param, so this function call could be misleading to someone. Perhaps `getSearchParam` or something like that would be more meaningful? WDYT?

mvidalgarcia: Yeah, I wasn't very sure either, I followed other function naming but the logic is not really the same. `getSearchParam` is a bit better but I'm still not convinced, maybe `encodeSearch`?

audrium: One of the hardest problems in CS..
`encodeSearch` sounds nice to me! Though it might be mixed with encoding of strings. Another solution would be simply `formatSearch`?

mvidalgarcia: Fair enough 👍

PR #213, opened 2022-01-25
package: upgrade to `react-scripts` 5.0.0

### LINKED ISSUE #204: package: bump all module versions

mvidalgarcia:
closes #204


run-tests.sh
```diff
@@ -20,8 +20,12 @@ check_sphinx () {
     sphinx-build -qnNW docs docs/_build/html
 }
 
+check_lint () {
+    cd reana-ui && yarn lint && cd ..
```
audrium, line 24: Minor: another option would be to use parentheses in commands. In this way we don't have to go back (`cd ..`). Command would look like this: `(cd reana-ui && yarn lint)`. Same could be applied for `check_prettier` and `check_js_tests`. WDYT?

run-tests.sh
```diff
@@ -55,6 +59,7 @@ do
     case $arg in
         --check-shellscript) check_script;;
         --check-sphinx) check_sphinx;;
+        --check-lint) check_lint;;
```
audrium, line 62: We should also add `check_lint` to `check_all()` and probably add a new GA CI job as well?

Minor: year was not updated in this file

reana-ui/package.json
```diff
@@ -16,19 +16,23 @@
...
     "start": "craco start",
     "build": "craco build",
     "test": "craco test --env=jsdom",
-    "eject": "craco eject"
+    "lint": "eslint .",
+    "prettier": "prettier --check . || (echo \"oops 🙀 pls run 'yarn fmt'\" && exit 1)",
+    "fmt": "prettier --write .",
+    "ci": "run-p lint prettier",
+    "eject": "craco eject",
+    "postinstall": "semantic-ui-css-patch"
```
audrium, line 35: Lots of new commands were added, it's not completely clear for me when one has to use it? For example `postinstall` command is triggered automatically or one has to do it at some point? Maybe it would be worth to have a short description in the readme?

audrium (approved):
Looks good and works fine for me.

I noticed that now we have both `yarn.lock` and `package-lock.json` which could be confusing. Since we switched completely to `yarn`, perhaps we can get rid of the `package-lock.json`? Otherwise it would be good to update it as well..
By the way, while installing I got this warning from `yarn`:
```
Warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json.
```

reana-ui/package.json
mvidalgarcia: https://docs.npmjs.com/cli/v8/using-npm/scripts#npm-install

mvidalgarcia:
> I noticed that now we have both `yarn.lock` and `package-lock.json` which could be confusing.

Where's the `package-lock.json`? I can't see it.

run-tests.sh
mvidalgarcia: Make sense, didn't know the subshell functionality 👍

audrium:
> > I noticed that now we have both `yarn.lock` and `package-lock.json` which could be confusing.
> 
> Where's the `package-lock.json`? I can't see it.

Indeed it seems it was generated by me locally and because of `.gitignore` I didn't spot it was not included in the repo.. Please ignore my comment, all is good :)

reana-ui/package.json
mvidalgarcia: Would something like this make sense?

```
Yarn scripts
============

- `start`: start a development server
- `build`: build a production-ready bundle in the `build` folder
- `test`: run unit tests
- `lint`: run linter
- `prettier`: check code formatting with prettier
- `fmt`: fix formatting problems with prettier
- `ci`: run both linter and format checkers, useful before committing changes
```
Most of them are very repetitive but it might help, not sure, let me know.

audrium: looks nice and clear in my opinion! thanks

run-tests.sh
audrium: What about GA CI job for linting? Don't we need it?

PR #225, opened 2022-02-23
ui: `LaunchOnReana` page

### LINKED ISSUE #221: design Web (and API) endpoints for running workflows from external sources
Design web endpoint for the functionality of running REANA workflows defined in external sources (GitHub, Zenodo, etc) without using `reana-client`.

Example A: It is URL that defines external source:

```
https://reana.cern.ch/run?url=https://github.com/cms-opendata-analyses/DimuonSpectrumNanoAODOutreachAnalysis/blob/2012/reana.yaml
```

Pros: unique scheme for all sources. Cons: Need to discover when to "git clone" everything vs when to "wget specs only" (e.g. when they are hosted on GitHub but the clone is not necessary).

Example B: It is "type" that defines external source:

```
https://reana.cern.ch/run?from=github;repo=foo;branch=bar;spec=reana-baz.yaml
https://reana.cern.ch/run?from=url;uri=github.com/cms-opendata-analyses/DimuonSpectrumNanoAODOutreachAnalysis/blob/2012/reana.yaml
```

Pros: Easier to distinguish when to fetch and when to clone. Cons: users need to "read docs" to understand URL creation.

Example C: Look at experiences with Binder, SWAN, etc.

Note: In either case, we may want to add other optional parameters, such as `...&name=myworkflow` that will create workflow of the desired name.  (And if `name` is not defined, it could be equal to the GitHub repository name, or equal to Zenodo record ID, etc. But that's the role for the backend later.)

Note: The Web endpoint will be the web page hosting the human-visible part of starting the workflow before redirecting:

```
- user starts at github.com/cms-opendata-analyses/DimuonSpectrumNanoAODOutreachAnalysis/blob/2012/reana.yaml
- badge brings user to https://reana.cern.ch/run?url=https://github.com/cms-opendata-analyses/DimuonSpectrumNanoAODOutreachAnalysis/blob/2012/reana.yaml
- when the workflow is created and started, the user is redirected to /details/uuid42
```

We may want to think about API endpoint that could offer this function programmatically without any user-facing part. The scenario would be then:

```
- system starts in ipython
- system calls https://reana.cern.ch/api/run?url=https://github.com/cms-opendata-analyses/DimuonSpectrumNanoAODOutreachAnalysis/blob/2012/reana.yaml
- the above page replies in JSON the uuid42 of the created workflow
- system concats other APIs to check status, see logs, etc
```

Note: We are focusing on the Web endpoint here in this sprint. But it's good to have the API endpoint optics in mind when designing the flow.

### LINKED ISSUE #223: ui: create intermediate spinner web page for running workflows from external sources
Once we have Web (and API) endspoints done, design a intermediate spinner-like web page that would display:

- creating workflow foo...
- fetching sources from zenodo...
- validating workflow specification...
- starting workflow execution...
- ... and redirect to `/details/uuid42` when all is done.

This page should typically run for 2-3 seconds only, because we are not
concerned with big file upload on the server-side. We are only uploading specs.

MVP: display one single spinner "Executing workflow..." before redirection. Do not display intermediate steps.

Beyond MVP: display various steps individually, especially if some step may take a lot of time.

Note that we probably don't want to go for the full socket.io real-time update user experience. It would be nice, but probably too much work for this sprint. Something we could revive as part of Live-Logs sprint later. An MVP of one single spinner with one single text would do for a start.

Note: Recall to nicely display any errors e.g. when fetching, when validating, etc. The reason should be easily understandable and "actionable" by the user.

mvidalgarcia:
closes #221
closes #223

**To test:**

- Boot up l[ocal REANA-UI](https://github.com/reanahub/reana-ui#readme)
- Access `http://localhost:3000/launch?url=https://zenodo.org/foo/bar/reana.yaml&name=myanalysis&parameters={%22events%22%3A%222000%22}&foo=bar` or similar.
- Edit `client.js`'s `launchWorkflow` function test multiple scenarios based on `Promises`

```js
const mockData = {
   data: {
     workflow_id: "cdcf48b1-c2f3-4693-8230-b066e088c6ac",
     workflow_name: "mytest.1",
     message: "The workflow has been successfully submitted.",
   },
 };

// Delayed Success `Promise` (Spinner)
function delay(t, v) {
  return new Promise(function (resolve) {
    setTimeout(resolve.bind(null, v), t);
  });
}
return delay(5000, mockData);

// Regular Success `Promise`
return Promise.resolve(mockData);
    
// Rejection `Promise` (Error)
return Promise.reject({
  response: { data: { message: "Something wrong happened! 💣" } },
});
```

### DIFF
--- a/reana-ui/src/hooks.js
+++ b/reana-ui/src/hooks.js
@@ -2,12 +2,13 @@
   -*- coding: utf-8 -*-
 
   This file is part of REANA.
-  Copyright (C) 2020 CERN.
+  Copyright (C) 2020, 2022 CERN.
 
   REANA is free software; you can redistribute it and/or modify it
   under the terms of the MIT License; see LICENSE file for more details.
 */
 
+import React from "react";
 import { useDispatch } from "react-redux";
 import { useHistory, useLocation } from "react-router-dom";
 
@@ -29,4 +30,13 @@ export function useSubmit(action) {
   };
 
   return handleSubmit;
+}
+
+/**
+ * React Hook to retrieve the current query string params.
+ * @returns URLSearchParams object
+ */
+export function useQuery() {
+  const { search } = useLocation();
+  return React.useMemo(() => new URLSearchParams(search), [search]);
 }
--- a/reana-ui/src/pages/launchOnReana/LaunchOnReana.js
+++ b/reana-ui/src/pages/launchOnReana/LaunchOnReana.js
@@ -9,7 +9,7 @@
 import { useState } from "react";
 import { useHistory } from "react-router-dom";
 import { useDispatch } from "react-redux";
-import { Button, Container, Loader } from "semantic-ui-react";
+import { Button, Container, Loader, Message } from "semantic-ui-react";
 
 import BasePage from "../BasePage";
 import { Title } from "~/components";
@@ -40,12 +40,39 @@ export default function LaunchOnReana() {
       .finally(() => setLoading(false));
   }
 
+  /**
+   * Filter out parameters not required by Launch on REANA logic.
+   * @param {URLSearchParams} query Standard URLSearchParams object.
+   */
   function filterParams(query) {
     for (const qsParamsKey of query.keys()) {
-      if (!LAUNCH_ON_REANA_PARAMS_WHITELIST.includes(qsParamsKey)) {
+      if (
+        !Object.keys(LAUNCH_ON_REANA_PARAMS_WHITELIST).includes(qsParamsKey)
+      ) {
         query.delete(qsParamsKey);
       }
     }
+  }
+
+  /**
+   * Filter out non-required Launch on REANA qs params.
+   * @returns {Array} List of required qs params.
+   */
+  function getRequiredQsParams() {
+    return Object.entries(LAUNCH_ON_REANA_PARAMS_WHITELIST)
+      .filter(([_, { required }]) => required)
+      .map(([param, _]) => param);
+  }
+
+  /**
+   *
+   * @param {URLSearchParams} query Search query string.
+   * @returns {boolean} Whether the search query string includes all the required parameters.
+   */
+  function isMissingRequiredParams(query) {
+    const requiredParams = getRequiredQsParams();
+    const qsKeys = [...query.keys()];
+    return !requiredParams.every((param) => qsKeys.includes(param));
   }
 
   return (
@@ -54,19 +81,28 @@ export default function LaunchOnReana() {
         <div>
           <Title>Launch on REANA</Title>
         </div>
-        <ul>
-          {[...query].map(([k, v]) => (
-            <li key={k}>
-              {k}: {v}
-            </li>
-          ))}
-        </ul>
+        {isMissingRequiredParams(query) ? (
+          <Message
+            icon="warning circle"
+            header="Missing required query string fields."
+            content={`The required fields: ${getRequiredQsParams()}`}
+            warning
+          />
+        ) : (
+          <ul>
+            {[...query].map(([k, v]) => (
+              <li key={k}>
+                {k}: {v}
+              </li>
+            ))}
+          </ul>
+        )}
         <Button
           primary
           icon="rocket"
           content="Launch"
           onClick={handleLaunch}
-          disabled={loading}
+          disabled={loading || isMissingRequiredParams(query)}
         />
         {loading && <Loader content="Executing workflow..." />}
       </Container>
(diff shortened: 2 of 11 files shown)


reana-ui/src/hooks.js
```diff
@@ -30,3 +31,12 @@ export function useSubmit(action) {
...
   return handleSubmit;
 }
+
+/**
+ * React Hook to retrieve the current query string params.
+ * @returns URLSearchParams object
+ */
+export function useQuery() {
+  const { search } = useLocation();
+  return React.useMemo(() => new URLSearchParams(search), [search]);
```
audrium, line 41: It's a pity that we have an old react router version, we could use [useSearchParams hook](https://reactrouter.com/docs/en/v6/api#usesearchparams) in v6.

Have you considered using `query-string` (we have it as a dependency) for parsing the query params instead of `URLSearchParams`? I guess both works in the same way..

reana-ui/src/pages/launchOnReana/LaunchOnReana.js
```diff
@@ -0,0 +1,75 @@
...
+        history.push(`/details/${workflowId}`);
+      })
+      .catch((err) => {
+        dispatch(errorActionCreator(err));
+      })
+      .finally(() => setLoading(false));
+  }
+
+  function filterParams(query) {
+    for (var qsParamsKey of query.keys()) {
```
audrium, line 44: Minor: could we use `let` here instead of `var`?

Also, if we would use `queryString.parse()` instead of `URLSearchParams()` inside `useQuery()` hook, we could use `filter` and `reduce` here to filter our the list.. perhaps worth switching?

reana-ui/src/pages/launchOnReana/__tests__/LaunchOnReana.test.js
```diff
@@ -0,0 +1,37 @@
...
+  expect(screen.getByRole("heading")).toHaveTextContent("Launch on REANA");
+  expect(screen.getByRole("button")).not.toBeDisabled();
+  expect(screen.getByText("url: https://example.org/reana.yaml"));
+  const invalidQsParams = screen.queryByText("foo: bar");
+  expect(invalidQsParams).toBeNull();
+
+  const loadingSpinner = screen.queryByText("Executing workflow...");
+  expect(loadingSpinner).toBeNull();
+  fireEvent.click(screen.getByText("Launch"));
+  expect(screen.getByText("Executing workflow..."));
```
audrium, line 36: Nice to see these tests in reana-ui!!  :rocket:

audrium (commented):
Nice work! I left couple of small comments. I think we could use `query-string` instead of `URLSearchParams` for parsing the query params. This will give us simple object which is easier to work with in `LaunchOnReana` page. 

Very minor: I think we could also display a warning or handle somehow the scenario when there are no query params in the page (or not all are present from `LAUNCH_ON_REANA_PARAMS_WHITELIST` list?) e.g. `http://localhost:3000/launch`

reana-ui/src/hooks.js
mvidalgarcia: > It's a pity that we have an old react router version, we could use [useSearchParams hook](https://reactrouter.com/docs/en/v6/api#usesearchparams) in v6.

Looks like there are pretty significant changes between v5 and v6. They even recommend waiting for a backwards compatibility package to be released.

> We recommend waiting for the backwards compatibility package to be released before upgrading apps that have more than a few routes.

> Have you considered using query-string (we have it as a dependency) for parsing the query params instead of URLSearchParams? I guess both works in the same way..

Yes. But I decided to stick with the standard library as they provide the functionality we want. This way is easier to get rid of 3rd party dependencies and make the app lighter, as well as having fewer dependency headaches related to upgrades, vulnerabilities, etc. I guess this would need further discussion and a consensual decision.

reana-ui/src/pages/launchOnReana/LaunchOnReana.js
mvidalgarcia: > Minor: could we use `let` here instead of `var`?

Got tricked by the [MDN example](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/keys#examples) 😅 

I think `const` would be enough.

mvidalgarcia:
> Very minor: I think we could also display a warning or handle somehow the scenario when there are no query params in the page (or not all are present from LAUNCH_ON_REANA_PARAMS_WHITELIST list?) e.g. http://localhost:3000/launch

Yeah, the UI is pretty minimal ATM and definitely not the final one. Even the qs params list is too simple, it's kind of a proof of concept. We would need to design something better taking into account that this could be not just read-only.

I'll add the warning.

audrium (approved):
works nicely!

PR #227, opened 2022-03-04
auth: redirect to requested page after sso login

### LINKED ISSUE #220: respect referer URLs in the logging chain
**Current behaviour**

Consider a workflow detailed page such as `https://reana.cern.ch/details/00000000-0000-0000-0000-000000000000`.

If a user is logged in, hitting this URL will open the workflow detailed page.

If a user is not logged in, hitting this URL behaves in the following way. The user is presented with a login page (SSO in case of CERN), and after successfully logging in, the user is redirected to the "Your workflows" page.

The original context of what user wanted to see is lost.

**Expected behaviour**

After logging in, the user should be automatically redirected to the "desired URL".

The "logging chain" may need a "referrer" argument to know where to land after logging in.

- /some/url -> need local login -> login form -> logged in -> redirect back to /some/url
- /some/url -> need CERN SSO -> CERN SSO login form -> logged in -> redirect back to /some/url

This will be useful when a user hits "Run-on-REANA" badge whilst not being logged in on the target REANA deployment.

mdonadoni:
Closes #220

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,7 @@ Version 0.9.0 (UNRELEASED)
 ---------------------------
 
 - Adds Launch on REANA page allowing the submission of workflows via badge-clicking.
+- Fixes redirection chain for non-signed-in CERN SSO users to access the desired target page after sign-in.
 
 Version 0.8.2 (2022-02-15)
 ---------------------------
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -5,6 +5,7 @@ Version 0.9.0 (UNRELEASED)
 ---------------------------
 
 - Adds Launch on REANA page allowing the submission of workflows via badge-clicking.
+- Adds redirection to the page requested by the user after logging-in using SSO.
```
tiborsimko, line 8: Since this is basically a bug fix, I would suggest something like:

- Fixes redirection chain for non-signed-in CERN SSO users to access the desired target page after sign-in.

tiborsimko (commented):
Works well :+1: 

Left a cosmetic comment for the change log message.

tiborsimko (approved):

Issue #230, opened 2022-03-08
detailed workflow page: confusing temporary message "workspace was deleted"


tiborsimko:
Take a workflow run that was ran a long time ago.

Go to the detailed workflow run page, the "Logs" tab.

Now switch to the "Workspace" tab. In my case, I'm seeing the following "workspace was deleted message":

![workspace1](https://user-images.githubusercontent.com/517546/157277233-f449ef1b-8d95-419d-849b-70d9473e4572.png)

Then, after about two seconds, the page refreshes and the "correct" workflow's workspace content is properly displayed.

It may be confusing to users to see a message saying "workspace was deleted", even if it is only temporary.

We should rather display the usual "textless spinner image" when the page is loading, as when users switch tabs by clicking between "Logs", "Workspace", and "Specification" tabs on the page later.

(IOW the "workspace was deleted" message seems to appear only the first time when the tab is switched.)

Issue #179, opened 2021-03-11
ui: SSO sign-in accompanied by non-SSO text


tiborsimko:
**Current behaviour**

For CERN deployments, one is greeted with:

![sso](https://user-images.githubusercontent.com/517546/110764596-86c6bd00-8253-11eb-8a7e-8a48bd8efcb6.png)

This may be providing contradictory information, because at CERN users should only use SSO, without having to contact REANA administrators.

**Expected behaviour**

Display only the SSO login button, named more precisely as _"Sign in with CERN SSO"_ or perhaps even _"Sign in with CERN Single Sign-On"_.

The phrase _"If you don't have an account..."_ should be displayed in non-SSO, non-local-sign-up, admins-create-accounts deployment conditions only.

tiborsimko:
Bumping this issue into the Run-on-REANA sprint, because when we start to publish badges on GitHub example repositories, the login page make become very visible to all sorts of CERN and non-CERN users.

We should amend the text as mentioned above, and also express who can connect upfront. For example, the phrase could read:

"Note that you need to hold an official CERN account in order to use `reana.cern.ch` service."

PR #233, opened 2022-03-09
ui: improve sso login text

audrium:
closes https://github.com/reanahub/reana-ui/issues/179
closes https://github.com/reanahub/reana-ui/issues/230

In order to test https://github.com/reanahub/reana-ui/issues/230:
use some network throttling (GPRS worked for me) and observe that the message is no longer present in Workspace tab.
Try to delete some workspace to see that it's displayed when needed.

In order to test https://github.com/reanahub/reana-ui/issues/179:
modify `cernSSO`, `localUsers`, `hideSignup` in `config reducer` [here](https://github.com/reanahub/reana-ui/blob/master/reana-ui/src/reducers.js#L129) 

Here is how the main use cases looks like after this PR:
- Production settings (`cernSSO: true, localUsers: false, hideSignup: true`)

- `cernSSO: false, localUsers: true, hideSignup: false`

- `cernSSO: false, localUsers: true, hideSignup: true`

- `cernSSO: true, localUsers: true, hideSignup: true`
Screenshot 2022-03-10 at 10-45-58 REANA

### DIFF
--- a/reana-ui/src/pages/signin/Signin.js
+++ b/reana-ui/src/pages/signin/Signin.js
@@ -59,7 +59,7 @@ export default function Signin() {
         {config.cernSSO && (
           <>
             <Button basic fluid size="large" onClick={handleClick}>
-              Sign in with SSO
+              Sign in with CERN Single Sign-On
             </Button>
             {config.localUsers && (
               <Divider section horizontal>
@@ -77,7 +77,13 @@ export default function Signin() {
           />
         )}
       </Segment>
-      {config.hideSignup && (
+      {config.hideSignup && !config.localUsers && config.cernSSO && (
+        <p>
+          Note that you need to hold an official CERN account in order to use
+          <Link to="/"> {window.location.hostname}</Link> service.
+        </p>
+      )}
+      {config.hideSignup && config.localUsers && (
         <p>
           If you do not have an account yet, please contact
           <a href={`mailto:${config.adminEmail}`}> REANA administrators</a>
--- a/reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
+++ b/reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
@@ -199,7 +199,7 @@ export default function WorkflowFiles({ id }) {
     />
   );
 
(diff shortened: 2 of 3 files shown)


reana-ui/src/pages/signin/Signin.js
```diff
@@ -77,7 +77,13 @@ export default function Signin() {
           />
         )}
       </Segment>
-      {config.hideSignup && (
+      {config.hideSignup && config.cernSSO && (
```
mdonadoni, line 80: When having both local accounts and users authenticated through SSO (that is `cernSSO: true`, `localUsers: true` and `hideSignup: true`), the message "_Note that you need to hold an official CERN account in order to use reana.cern.ch service_" is shown. Should we tell the users they can request an account by contacting the administrators, like we do when `cernSSO: false`, `localUsers: true` and `hideSignup: true`?

reana-ui/src/pages/signin/Signin.js
```diff
@@ -77,7 +77,13 @@ export default function Signin() {
...
+        <p>
+          Note that you need to hold an official CERN account in order to use
+          <Link to="/"> reana.cern.ch</Link> service.
```
mdonadoni, line 83: What do you think about using `window.location.hostname` instead of `reana.cern.ch`?

mdonadoni, line 83: What do you think about using `window.location.hostname` instead of `reana.cern.ch`?

reana-ui/src/pages/signin/Signin.js
audrium: Thanks for spotting this! Working with 3 flags is like solving a three body problem. I've fixed it and included an image how it looks like in this case

reana-ui/src/pages/signin/Signin.js
audrium: I agree that it's not too nice to hardcode it like this, my thinking was that for `cernSSO` case this value most likely will never change. Ideally we would have a config for this coming from the api, for now I've switched it to `window.location.hostname` as you suggested, thanks

Issue #229, opened 2022-03-08
performance: improve slow detailed workflow page refresh


tiborsimko:
A problem spotted on the CERN production site.  When a user has more workflow runs than the default web page limit (5), and views some detailed workflow page and its workspace i.e. when the user is on some "Workspace" tab, then upon hitting "Refresh" in the browser the web page spins for several seconds showing "Loading workflow..." before it lands on the refreshed page. (BTW this refreshed page is the "wrong" "Logs" tab, not the original "Workspace" tab, but that can be ticketised as another issue.)

The problem is that it takes perhaps 10 seconds to refresh that page, which is unnecessarily slow.

How to reproduce locally:

1. Launch many workflows e.g. via

```console
$ reana-dev kind-load-docker-image -c DEMO
$ reana-dev run-example -b kubernetes -c DEMO --submit-only
```

2. Open a certain detailed workflow web page e.g.

```console
$ firefox https://0.0.0.0:30443/details/a50ab2fd-66b1-4cc0-9575-022e170cbe4b
```

3. Watch REST API requests coming from r-ui on the r-server pod:

```console
$ kubectl logs deployment/reana-server rest-api -f
```

4. Hit reload in the browser whilst observing logs; they will say:

```
[pid: 16|app: 0|req: 182/338] 1.2.0.1 () {52 vars in 1267 bytes} [Tue Mar  8 13:11:08 2022] GET /api/config => generated 353 bytes in 10 msecs (HTTP/1.1 200) 15 headers in 583 bytes (1 switches on core 1)
[pid: 16|app: 0|req: 182/339] 1.2.0.1 () {52 vars in 1261 bytes} [Tue Mar  8 13:11:08 2022] GET /api/you => generated 490 bytes in 26 msecs (HTTP/1.1 200) 15 headers in 583 bytes (1 switches on core 3)
[pid: 16|app: 0|req: 183/340] 1.2.0.1 () {52 vars in 1300 bytes} [Tue Mar  8 13:11:08 2022] GET /api/workflows?verbose=true& => generated 28321 bytes in 101 msecs (HTTP/1.1 200) 15 headers in 585 bytes (1 switches on core 2)
[...]

PR #234, opened 2022-03-10
client: improve fetching workflow by id

audrium:
closes https://github.com/reanahub/reana-ui/issues/229

The problem was that when user refreshed detailed workflow page, [fetchWorkflow(id) action](https://github.com/reanahub/reana-ui/blob/master/reana-ui/src/actions.js#L271) was triggered which fetched all of the user workflows. For users who created lots of the workflows this operation was not very optimal.. This PR fixes this issue by fetching a specific workflow in this case.

To test go to detailed workflow page, do a hard refresh and inspect that only a specific workflow was fetched instead all of them.

### DIFF
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -232,19 +232,20 @@ export function confirmUserEmail(token) {
   };
 }
 
-export function fetchWorkflows(
+export function fetchWorkflows({
   pagination,
   search,
   status,
   sort,
-  showLoader = true
-) {
+  showLoader = true,
+  workflowId,
+}) {
   return async (dispatch) => {
     if (showLoader) {
       dispatch({ type: WORKFLOWS_FETCH });
     }
     return await client
-      .getWorkflows(pagination, formatSearch(search), status, sort)
+      .getWorkflows(pagination, formatSearch(search), status, sort, workflowId)
       .then((resp) =>
         dispatch({
           type: WORKFLOWS_RECEIVED,
@@ -268,7 +269,13 @@ export function fetchWorkflow(id) {
     if (workflow) {
       return workflow;
     } else {
-      dispatch(fetchWorkflows());
+      dispatch(
+        fetchWorkflows({
+          workflowId: {
+            workflow_id_or_name: id,
+          },
+        })
+      );
     }
   };
 }
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,7 @@ Version 0.9.0 (UNRELEASED)
 - Adds Launch on REANA page allowing the submission of workflows via badge-clicking.
 - Adds 404 Not Found error page.
 - Fixes redirection chain for non-signed-in CERN SSO users to access the desired target page after sign-in.
+- Fixes ``fetchWorkflow`` action to fetch a specific workflow instead of the entire user workflow list.
 
 Version 0.8.2 (2022-02-15)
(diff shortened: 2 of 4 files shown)


reana-ui/src/actions.js
```diff
@@ -268,7 +269,11 @@ export function fetchWorkflow(id) {
     if (workflow) {
       return workflow;
     } else {
-      dispatch(fetchWorkflows());
+      dispatch(
+        fetchWorkflows(...Array(5), {
```
mvidalgarcia, line 273: Phew! It took me a while to understand the `Array(5)` part. I think that for these cases we can make use of destructuring objects as function args. We can make `fetchWorkflows` to expect an object instead of multiple args.

After playing a little bit with it, I came up with this solution, let me know what you think:

```diff
diff --git a/reana-ui/src/actions.js b/reana-ui/src/actions.js
index 3d4ac03..aaa950b 100644
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -232,14 +232,14 @@ export function confirmUserEmail(token) {
   };
 }
 
-export function fetchWorkflows(
+export function fetchWorkflows({
   pagination,
   search,
   status,
   sort,
   showLoader = true,
-  workflowId = {}
-) {
+  workflowId,
+}) {
   return async (dispatch) => {
     if (showLoader) {
       dispatch({ type: WORKFLOWS_FETCH });
@@ -270,8 +270,10 @@ export function fetchWorkflow(id) {
       return workflow;
     } else {
       dispatch(
-        fetchWorkflows(...Array(5), {
-          workflow_id_or_name: id,
+        fetchWorkflows({
+          workflowId: {
+            workflow_id_or_name: id,
+          },
         })
       );
     }
diff --git a/reana-ui/src/pages/workflowList/WorkflowList.js b/reana-ui/src/pages/workflowList/WorkflowList.js
index 98d057b..7e4b08a 100644
--- a/reana-ui/src/pages/workflowList/WorkflowList.js
+++ b/reana-ui/src/pages/workflowList/WorkflowList.js
@@ -71,20 +71,25 @@ function Workflows() {
 
   useEffect(() => {
     dispatch(
-      fetchWorkflows({ ...pagination }, searchFilter, statusFilter, sortDir)
+      fetchWorkflows({
+        pagination: { ...pagination },
+        search: searchFilter,
+        status: statusFilter,
+        sort: sortDir,
+      })
     );
 
     if (!interval.current && reanaToken && pollingSecs) {
       interval.current = setInterval(() => {
         const showLoader = false;
         dispatch(
-          fetchWorkflows(
-            { ...pagination },
-            searchFilter,
-            statusFilter,
-            sortDir,
-            showLoader
-          )
+          fetchWorkflows({
+            pagination: { ...pagination },
+            search: searchFilter,
+            status: statusFilter,
+            sort: sortDir,
+            showLoader,
+          })
         );
         setRefreshedAt(currentUTCTime());
       }, pollingSecs * 1000);

```

Perhaps it's too much of a change, but it looks cleaner to me. I haven't tested all the affected parts.

audrium: Thanks for a nice suggestion! Indeed looks much cleaner and works as expected

mvidalgarcia (approved):
WDYT about adding a changelog? Looks like a relevant performance improvement.

Same for the [RWC PR](https://github.com/reanahub/reana-workflow-controller/pull/434).

PR #243, opened 2022-03-16
LaunchOnReana: clear notifications after workflow launch

### LINKED ISSUE #239: launcher: clean error notifications on successful workflow launch
Issue: error notification is carried to the workflow details page after successful workflow launch.

How to reproduce:
After implementing rate limit for the launch endpoint user can quickly press the launch button couple of times during the 5 seconds window. In this case the error notification is shown which is carried to the workflow details page after successful workflow launch.
Result:
Screenshot 2022-03-14 at 16-20-16 REANA

We should clear notifications on successful workflow launch

mdonadoni:
Closes #239 

How to test: Try to reproduce the issue
Expected result: The error notification is not shown after successfully launching a workflow


VMois (approved):
Works as expected. Nice!

Issue #245, opened 2022-03-17
ui: minor, status and running time fo workflow overflows when name is too long


VMois:
![Screenshot 2022-03-17 at 08 50 48](https://user-images.githubusercontent.com/15694232/158762331-4b46e204-5bbb-4be3-b390-8faf3500abb2.png)

Screen resolution: 1920x1200

For testing, you can use the following workflows:

- https://github.com/cms-opendata-analyses/DimuonSpectrumNanoAODOutreachAnalysis/tree/4eb69e073d3b2567c49b2525c2a53341a16dc1ab

- https://github.com/cms-opendata-analyses/DimuonSpectrumNanoAODOutreachAnalysis/tree/v1.3

Launch them with the Launch-on-REANA badge for simplicity.

VMois:
![Screenshot 2022-03-17 at 08 59 21](https://user-images.githubusercontent.com/15694232/158763322-31c98e57-0615-4f27-85c6-9be1c0248c29.png)

mvidalgarcia:
I can't help but think about this.

![image](https://user-images.githubusercontent.com/8863238/158785774-1034415b-798b-412e-ac9e-5243d88ad24d.png)

Issue #244, opened 2022-03-16
details: harmonise look-and-feel of "refresh" and "hamburger" buttons


tiborsimko:
When a workflow is running, then on the detailed workflow page, when the mouse is far away from the area, one sees:

![r1](https://user-images.githubusercontent.com/517546/158650489-34991cd8-6f97-4181-a5b4-e20be8031d70.png)

When mouse hovers over "refresh" button, one sees it getting darker:

![r2](https://user-images.githubusercontent.com/517546/158650561-8d12e9e4-6e48-472b-924d-62a66331d113.png)

When mouse hovers over "hamburger" button, one sees it getting lighter (and colourful):

![r3](https://user-images.githubusercontent.com/517546/158650735-943dd486-a5db-4ae0-aff8-b76d218f5a65.png)

We should harmonise the style so that when both are "off" they are displayed in the same colour (e.g. dark gray), and when some is "on", it is displayed in another colour (e.g. light grey or light brown).

P.S. Besides cosmetics, using very light grey for "refresh" makes the "refresh" button not to be very visible, which may be a usability problem, because the details page is not auto-refreshing. So we may take the harmonisation occasion to make it nicely visible on pages. Especially useful for Launch-on-REANA functionality.

mvidalgarcia:
> We should harmonise the style so that when both are "off" they are displayed in the same colour (e.g. dark gray), and when some is "on", it is displayed in another colour (e.g. light grey or light brown).

+1 on harmonizing.

All across the UI we have lighter colors when "off" and slightly darker colors on hover, so I would suggest to be consistent with this.

PS. Reminder to also change the kebab menu colors on the workflow list page.

PR #246, opened 2022-03-17
global: minor UI fixes

### LINKED ISSUE #249: long workflow name breaks "details" top view
For example, when using the launcher, in some cases the workflow name can be very long and break the` details` page layout. The maximum allowed length of workflow name is 255 characters.

Screenshot 2022-03-21 at 13 38 40

_Originally posted by @VMois in https://github.com/reanahub/reana-server/issues/464#issuecomment-1073860831_

mvidalgarcia:
closes https://github.com/reanahub/reana-ui/issues/244
closes https://github.com/reanahub/reana-ui/issues/245
closes #249

image

### DIFF
--- a/reana-ui/src/components/WorkflowActionsPopup.module.scss
+++ b/reana-ui/src/components/WorkflowActionsPopup.module.scss
@@ -2,7 +2,7 @@
   -*- coding: utf-8 -*-
 
   This file is part of REANA.
-  Copyright (C) 2020 CERN.
+  Copyright (C) 2020, 2022 CERN.
 
   REANA is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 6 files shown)


tiborsimko (approved):
Looks nice.

One side thing I wonder is whether we shouldn't "darken" a little bit the text "Refreshed at 08:00:14 UTC" appearing on the workflow list page: currently it's too light perhaps. Darkening it would play nicely with the darker "refresh" button introduced by this PR. And we could also envisage placing a small "refresh" button just before this text, so that impatient users have a visual clue about refresh right there. (This could also allow us to decrease the polling frequency if we want in the future.)

audrium (approved):
Nicely done! Everything works as expected!

PR #253, opened 2022-03-23
ui: add polling to the workflow details page

audrium:


reana-ui/src/pages/workflowDetails/WorkflowDetails.js
```diff
@@ -32,20 +38,47 @@ export default function WorkflowDetailsPage() {
...
   }, [dispatch, workflowId]);
 
+  useEffect(() => {
+    if (workflow && FINISHED_STATUSES.includes(workflow.status)) {
+      cleanPolling();
+    }
+  }, [workflow]);
+
+  const setPolling = () => {
+    if (!interval.current && reanaToken && pollingSecs) {
```
mvidalgarcia, line 70: Do we need the reana token here? I understand this was added to the workflow list page to prevent the polling when the token was still not granted but I don't think anyone can get to the workflow details page with no token.

mvidalgarcia (commented):
It's doing the job but I have some observations:

- The polling only fetches general information about the workflow, no information about the logs and steps so while the workflow might show as `finished`, with no manual refresh from the user the logs aren't refreshed, and the steps are still `running`.

image

- When the polling happens, I can see some blinking. I guess this happens to display a loader. In the case of polling, we should disable the loader as we do for the workflow list page.

- I see the logic and functionality are very similar to the one on the workflow list page. Would it make sense to encapsulate the polling functionality into a custom hook `usePolling` to reuse it in both places?  It probably doesn't belong to this task, something to take into account for the future.

audrium:
>     * The polling only fetches general information about the workflow, no information about the logs and steps so while the workflow might show as `finished`, with no manual refresh from the user the logs aren't refreshed, and the steps are still `running`.

Thanks for your nice suggestions! I've changed the code accordingly. Funny enough, the logs were actually updating to me before I've disabled the loader. Apparently it was causing a rerender of `WorkflowLogs` component which refetched the data as a side effect? :) In any case, it's fixed now, the logs and also the step labels are changing as expected.

CHANGES.rst
```diff
@@ -7,6 +7,7 @@ Version 0.9.0 (UNRELEASED)
 - Adds a label link to the workflow launcher URL remote origin on the workflow details page.
 - Adds Launch on REANA page allowing the submission of workflows via badge-clicking.
 - Adds 404 Not Found error page.
+- Adds refresh polling to the workflow details page.
```
mdonadoni, line 10: minor: What do you think about `Adds periodic refresh of the workflow details page`?

mdonadoni (approved):
Works as expected! I left only a minor comment

PR #255, opened 2022-03-29
ui: simplify SSO sign-in message

tiborsimko:
The SSO sign-in message now reads:

2022-03-29-184023_502x329_scrot

The link to `reana-qa.cern.ch` is not user friendly, because it leads to the same page... The user does not learn anything new by following it.

This commit improves the UX by simply removing the link.


audrium (approved):
Works as expected
Screenshot 2022-03-30 at 09-52-29 REANA

PR #257, opened 2022-04-05
actions: add missing default parameter to `triggerNotification`

mdonadoni:
This PR fixes an issue that caused some notifications not to be correctly shown to the user, for example when deleting workflows or opening Jupyter notebooks.

How to test:
Delete a workflow from the REANA homepage and check that notifications are correctly displayed at the top of the page.


audrium (approved):
Good catch, works nicely!

PR #258, opened 2022-04-07
auth: migrate to the new SSO

audrium:
closes https://github.com/reanahub/reana-server/issues/480

### DIFF
--- a/reana-ui/src/pages/signin/OAuthSignin.js
+++ b/reana-ui/src/pages/signin/OAuthSignin.js
@@ -1,0 +1,35 @@
+/*
+  -*- coding: utf-8 -*-
+
+  This file is part of REANA.
+  Copyright (C) 2022 CERN.
+
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+import { Redirect } from "react-router-dom";
+import { useDispatch } from "react-redux";
+
+import { triggerNotification } from "~/actions";
+import { useQuery } from "~/hooks";
+
+export default function OAuthSignin() {
+  const dispatch = useDispatch();
+  const query = useQuery();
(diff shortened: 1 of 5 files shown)


reana-ui/src/pages/workflowList/WorkflowList.js
```diff
@@ -111,6 +115,10 @@ function Workflows() {
     interval.current = null;
   };
 
+  if ("next_url" in queryPrams) {
+    return <Redirect to={queryPrams.next_url} />;
+  }
+
```
mdonadoni, line 121: What do you think about moving the redirection logic somewhere else?

Two possible options:
1. Move it to `WorkflowListPage`, so that we do not add more complexity to `Workflows`.
2. Create a new route and modify reana-server's configuration so that, after logging in, the user is redirected there. We can achieve this by modifying [these options](https://github.com/inveniosoftware/invenio-oauthclient/blob/v1.5.4/invenio_oauthclient/contrib/cern_openid.py#L148-L151) of `invenio-oauthclient`.

We should also handle login errors, for example by showing some notifications to the user, but this can be done in another PR.

reana-ui/src/pages/signin/OAuthSignin.js
```diff
@@ -0,0 +1,35 @@
...
+import { Redirect } from "react-router-dom";
+import { useDispatch } from "react-redux";
+
+import { triggerNotification } from "~/actions";
+import { useQuery } from "~/hooks";
+
+export default function OAuthSignin() {
+  const dispatch = useDispatch();
+  const query = useQuery();
+  const queryPrams = Object.fromEntries(query.entries());
```
mdonadoni, line 20: minor: `queryPrams` should be `queryParams`

mdonadoni (approved):
Works as expected, there is just one small typo to fix :tada:

PR #262, opened 2022-04-14
wf-details: refresh after workflow deletion

### LINKED ISSUE #259: workflow deletion UI loop and freeing disk troubles
**Current behaviour**

When a user is on the "Your workflows page" and would like to delete one workflow using the hamburger menu in the regular web interface, the user is brought to a new page that says:

delete

Clicking on the "Delete" button does a job, sort of, because the workflow is deleted behind the scenes, but the page blinks white and the user is redirected to the very same page offering the Delete or Cancel options.

**Expected behaviour**

When the deletion is successful, the user would expect to see the confirmation and would expect to be brought back to the workflow list page from where the user started. 

**Note**

Whilst working on fixing this, it may be good to check up the "Free up disk" functionality from the hamburger menu on the detailed workflow page. It seems it is not working when trying to activate it after the above scenario.

delete2

mdonadoni:
Closes #259


audrium (approved):
LGMT, Great work!

PR #263, opened 2022-04-27
profile: disable radio button when toggling gitlab integration

### LINKED ISSUE #261: gitlab: disable toggle button while it's waiting for the response
Right now when someone tries to integrate a Gitlab project to REANA and toggles a specific project twice quickly (easy to do while using a slow internet connection), it will create two gitlab webhooks. It means REANA will start two workflows on every change in a Gitlab project.

How to reproduce:
1. Use some connection throttling (e.g. I've used `Good 3G` to slow down the connection)
2. Navigate to `profile` page and press quickly to toggle a gitlab poject. It should result in two `POST` request like here:
image
3. Do some changes in a Gitlab project
4. REANA will start two workflows

We should fix this issue by:
- Introducing `loading` state for gitlab toggle button in `reana-ui` to disable it until there is a response from the server
- Preventing from duplicate webhooks creation for the same project in `reana-server`

mdonadoni:
Closes #261

How to test:
1. Patch `GitLabProjects.js`:
```diff
@@ -26,8 +26,24 @@ export default function GitLabProjects() {
      */
     const getProjects = () => {
       setFetchingProjects(true);
-      client
-        .getGitlabProjects()
+      //   client
+      //     .getGitlabProjects()
+      Promise.resolve({
+        data: {
+          123: {
+            hook_id: null,
+            name: "abc",
+            path: "johndoe/abc",
+            url: "https://gitlab.cern.ch/johndoe/abc",
+          },
+          456: {
+            hook_id: null,
+            name: "xyz",
+            path: "johndoe/xyz",
+            url: "https://gitlab.cern.ch/johndoe/xyz",
+          },
+        },
+      })
         .then((res) => {
           let data = res.data;
           for (let projectId in data) {
@@ -70,8 +86,18 @@ export default function GitLabProjects() {
       expectedStatus = 204;
     }
 
-    client
-      .toggleGitlabProject(method, data)
+    // client
+    //   .toggleGitlabProject(method, data)
+    new Promise((resolve) =>
+      setTimeout(() => {
+        resolve({
+          status: expectedStatus,
+          data: {
+            id: 1,
+          },
+        });
+      }, 1000)
+    )
       .then((res) => {
         if (res.status === expectedStatus) {
           setProjects((currentProjects) => ({
```
2. Toggle the switches to turn on and off the GitLab integration

Expected result: The switches are disabled after being toggled


reana-ui/src/pages/profile/components/GitLabProjects.js
```diff
@@ -29,7 +29,11 @@ export default function GitLabProjects() {
       client
         .getGitlabProjects()
         .then((res) => {
-          setProjects(res.data);
+          let data = res.data;
+          for (let projectId in data) {
+            data[projectId].toggling = false;
+          }
```
audrium, line 35: Very minor: instead of mutating the data object we could do it like this:
```
const data = Object.entries(res.data).map(([id, val]) => ({
  [id]: { ...val, toggling: false },
}));
```
but I guess it's a bit less readable, so we can leave it as is

reana-ui/src/pages/profile/components/GitLabProjects.js
```diff
@@ -2,7 +2,7 @@
 	-*- coding: utf-8 -*-
 
   This file is part of REANA.
-  Copyright (C) 2020 CERN.
+  Copyright (C) 2020, 2021, 2022 CERN.
```
audrium, line 5: Extremely minor: we shouldn't add 2021 here

audrium (approved):
Works nicely! :rocket: 
P.S. Thanks for testing instructions

reana-ui/src/pages/profile/components/GitLabProjects.js
mdonadoni:
```js
let newProjects = {};
for (const [id, details] of Object.entries(res.data)) {
  newProjects[id] = {
    ...details,
    toggling: false,
  };
}
```

What do you think?

audrium: Looks good :+1:

PR #265, opened 2022-05-02
wf-details: show the logs of the workflow engine

### LINKED ISSUE #166: ui: refine workflow logs
If the workflow fails at the workflow engine level, before starting any jobs, the UI shows no error logs. While you can get the complete logs with the client, it might be confusing for some users to see a failed workflow without any errors on the UI. 

Possible solutions

*  Show only the workflow engine logs if the workflow fails before creating jobs
* Add another tab (e.g. next to "Step") or option in drop down to show workflow  engine logs. Could be always present.

mdonadoni:
Closes #166

### DIFF
--- a/reana-ui/src/pages/workflowDetails/WorkflowDetails.js
+++ b/reana-ui/src/pages/workflowDetails/WorkflowDetails.js
@@ -103,8 +103,16 @@ function WorkflowDetails() {
 
   const panes = [
     {
-      menuItem: { key: "logs", icon: "terminal", content: "Logs" },
-      render: () => <WorkflowLogs id={workflow.id} />,
+      menuItem: { key: "job-logs", icon: "terminal", content: "Job logs" },
+      render: () => <WorkflowLogs workflow={workflow} />,
+    },
+    {
+      menuItem: {
+        key: "engine-logs",
+        icon: "cogs",
+        content: "Engine logs",
+      },
+      render: () => <WorkflowLogs engine workflow={workflow} />,
     },
     {
       menuItem: {
--- a/reana-ui/src/pages/workflowDetails/components/WorkflowLogs.js
+++ b/reana-ui/src/pages/workflowDetails/components/WorkflowLogs.js
@@ -12,24 +12,42 @@ import { useEffect, useState } from "react";
 import PropTypes from "prop-types";
 import findKey from "lodash/findKey";
 import { useDispatch, useSelector } from "react-redux";
-import { Icon, Dropdown, Label, Loader } from "semantic-ui-react";
+import { Icon, Dropdown, Label, Loader, Message } from "semantic-ui-react";
 
 import { fetchWorkflowLogs } from "~/actions";
+import { NON_FINISHED_STATUSES } from "~/config";
 import { statusMapping } from "~/util";
 import { getWorkflowLogs, loadingDetails } from "~/selectors";
 import { CodeSnippet, TooltipIfTruncated } from "~/components";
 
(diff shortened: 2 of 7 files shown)


reana-ui/src/pages/workflowDetails/components/WorkflowLogs.js
```diff
@@ -21,15 +21,26 @@ import { CodeSnippet, TooltipIfTruncated } from "~/components";
...
-  const [selectedStep, setSelectedStep] = useState();
+function EngineLogs({ logs }) {
+  return logs ? (
+    <CodeSnippet dollarPrefix={false} classes={styles.logs}>
+      {logs}
+    </CodeSnippet>
+  ) : (
+    <Message
+      icon="info circle"
+      content="There are no workflow engine logs available."
```
audrium, line 32: Minor: perhaps "There are no workflow engine logs available yet." message would better inform users that it will soon be populated. Or we add a sentence describing that the logs will be shown once the workflow finishes

audrium (approved):
Works nicely! I personally like this approach to display the engine logs in a different tab.
I think it's better than trying to fit it along the job steps :+1: 
@tiborsimko please share your observations about this

reana-ui/src/pages/workflowDetails/components/WorkflowLogs.js
```diff
@@ -21,15 +21,26 @@ import { CodeSnippet, TooltipIfTruncated } from "~/components";
...
-  const [selectedStep, setSelectedStep] = useState();
+function EngineLogs({ logs }) {
+  return logs ? (
+    <CodeSnippet dollarPrefix={false} classes={styles.logs}>
+      {logs}
+    </CodeSnippet>
+  ) : (
+    <Message
+      icon="info circle"
+      content="There are no workflow engine logs available yet."
```
tiborsimko, line 32: Whlist the workflow is running/queued/pending, it would be better to say something like:

- The workflow engine logs will be available after the workflow run finishes.

And, when there the execution finishes, usually there will be something, also for failed/stopped/etc runs. But if there is still nothing, we can display a generic message:

- There are no workflow engine logs for this execution run.

The motivation is simply to tell users when to expect logs to be visible.  ("yet" is too generic)

reana-ui/src/pages/workflowDetails/WorkflowDetails.js
```diff
@@ -103,9 +103,17 @@ function WorkflowDetails() {
...
     {
-      menuItem: { key: "logs", icon: "terminal", content: "Logs" },
+      menuItem: { key: "job-logs", icon: "terminal", content: "Job Logs" },
       render: () => <WorkflowLogs id={workflow.id} />,
     },
+    {
+      menuItem: {
+        key: "engine-logs",
+        icon: "cogs",
+        content: "Engine Logs",
```
tiborsimko, line 113: What about "Engine logs" case in the visible menu title,  to be consistent with BrE style rather than AmE style of title headings.

reana-ui/src/pages/workflowDetails/WorkflowDetails.js
```diff
@@ -103,9 +103,17 @@ function WorkflowDetails() {
 
   const panes = [
     {
-      menuItem: { key: "logs", icon: "terminal", content: "Logs" },
+      menuItem: { key: "job-logs", icon: "terminal", content: "Job Logs" },
```
tiborsimko, line 106: What about "Job logs" case in the visible menu title, to be consistent with BrE style rather than AmE style of title headings.

PR #269, opened 2022-05-11
router: do not change URL when showing 404 error page

### LINKED ISSUE #268: router: do not redirect to `/404` if page is not found
When a user tries to access an invalid URL, he is redirected to `/404` and an error page is shown. Instead of doing this, we should not modify the URL and just show the error page.

This is where the redirect is done:
https://github.com/reanahub/reana-ui/blob/1aa967a4bdf34a2b09d1df0c0e03c2d3ce8a4267/reana-ui/src/components/App.js#L93

mdonadoni:
Closes #268

### DIFF
--- a/nginx/reana-ui.conf
+++ b/nginx/reana-ui.conf
@@ -3,7 +3,14 @@ server {
     root   /usr/share/nginx/html;
     server_name  localhost;
     index  index.html index.htm;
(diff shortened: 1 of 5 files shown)


tiborsimko (approved):
Works nicely :+1:

PR #272, opened 2022-05-24
wf-list: add switch to hide deleted workflows

### LINKED ISSUE #218: workflow list: provide option to hide deleted workflows
**Current behaviour**

The workflow list page allows filtering workflows by status:

filter

However, this filtering does not allow to _hide_ some workflows, such as deleted workflows, by default.

**Expected behaviour**

It would be useful to offer an option to users not to see deleted workflows. This could be implemented (i) on the workflow list page via some UI element such as checkbox saying `[x] hide deleted workflows`, but perhaps (ii) even better would be to offer personalised global user settings option that people could set in their "Your profile" / "Your setting" pages. (Alongside with other interesting configurable options, such as the number of hits to display per page, in the longer-term future.)

mdonadoni:
Closes #218


reana-ui/src/pages/workflowList/components/WorkflowStatusFilter.js
```diff
@@ -2,39 +2,94 @@
...
+    if (!showDeleted) {
+      selection = removeDeleted(selection);
+      if (selection.length === 0) {
+        selection = NON_DELETED_STATUSES;
+      }
+    }
+    if (!hasSameElements(selection, statusFilter)) {
+      filter(selection);
+    }
+  }, [filter, statusFilter, showDeleted, valueList]);
```
audrium, line 64: What do you think if we would simplify this code like this:
```js
  import isEqual from "lodash/isEqual";

  useEffect(() => {
    // Keep the list of selected statuses in sync with the filter used to make the
    // requests to the API
    let selection = [...valueList];
    if (!showDeleted) selection = removeDeleted(selection);
    if (!showDeleted && !selection.length) selection = NON_DELETED_STATUSES;

    if (!isEqual(selection, statusFilter)) {
      filter(selection);
    }
  }, [filter, statusFilter, showDeleted, valueList]);
  ```
  In this way we can get rid of `hasSameElements()` and replace it with `isEqual()` from `lodash`

File: reana-ui/src/pages/workflowList/components/WorkflowStatusFilter.js [hunk 2]
```diff
@@ -2,39 +2,94 @@
...
   key: status,
   text: status,
   value: status,
   icon: statusMapping[status].icon,
 }));
 
-export default function WorkflowStatusFilters({ valueList, filter }) {
+const nonDeletedStatusOptions = statusOptions.filter(({ value: status }) =>
+  NON_DELETED_STATUSES.includes(status)
+);
```
audrium, line 33: Very minor suggestion: we could avoid defining `WORKFLOW_STATUSES` (even though it's perfectly fine to keep it, since it might be useful in the future!) by doing this:
```js
const nonDeletedStatusOptions = statusOptions.filter(
  (status) => status.value !== "deleted"
);

const nonDeletedStatuses = nonDeletedStatusOptions.map(
  (status) => status.key
);
```

reana-ui/src/pages/workflowList/components/WorkflowStatusFilter.js
mdonadoni: Thank you! I did not know about `isEqual`

reana-ui/src/pages/workflowList/components/WorkflowStatusFilter.js
mdonadoni: I think we can keep this, because as you say it will probably be useful in the future

reana-ui/src/pages/workflowList/components/WorkflowStatusFilter.js
```diff
@@ -2,39 +2,80 @@
...
+
+export default function WorkflowStatusFilters({ statusFilter, filter }) {
+  const [valueList, setValueList] = useState([]);
+  const [showDeleted, setShowDeleted] = useState(false);
```
audrium, line 38: Small issue appeared when you changed to hide workflows by default. It triggers two API requests now - one to fetch all workflow statuses and then after `useEffect` triggers it filters all except deleted. I think we should set `NON_DELETED_STATUSES` by default [here](https://github.com/reanahub/reana-ui/blob/master/reana-ui/src/pages/workflowList/WorkflowList.js#L51) to solve it

audrium (approved):
LGMT! works nicely!

Issue #160, opened 2020-11-20
ui: quota notifications, e.g. 90%+ quota usage reached


mvidalgarcia:
The idea here is to display some notifications in the UI part when a certain percentage of usage is reached. This way we can keep the user informed and they can free some disk by deleting old workflows.

We can also consider sending emails when reaching certain limits.

tiborsimko:
We do have disk quota usage numbers displayed in the "Your profile" page, but that is a bit hidden.

It would be good to display a small teaser right on the home page for better awareness, e.g. as people may know about it from Google services:

![disk-warning](https://user-images.githubusercontent.com/517546/169248588-cfbaf0ab-c7b8-471a-8e4d-e8765bb8c1ff.png)

We could similarly display a gentle teaser on the workflow list page that would show above certain threshold and that would guide the user to a page explaining what to do (sort workflows by most-disk-space-used, delete unneeded ones, set up shorter retention periods for unneeded files, etc).

P.S. The email sending shouldn't be necessary for MVP. If we introduce stricter disk quotas, users would be prevented from launching new workflows, and would be guided to this info page, so that they would create "lighter workflows" progressively.  

(Although email warning might be good to add later for GitLab <-> REANA bridge users who don't necessarily see the REANA home page so often.)

PR #275, opened 2022-05-31
ui: add notification icon to inform when critical levels of quota usage is reached

audrium:
closes https://github.com/reanahub/reana-ui/issues/160

- Adds notification icon to the header to inform users when critical levels of quota usage is reached
- Adds same messages to Profile page under `Your quota` section
- Renames `HDD` to `Disk` in Profile page quotas section to be compatible with CLI
- Removes `help` button from the header

To test:
- Change `termination_update_policy` to `disk,cpu` in [values.yaml](https://github.com/reanahub/reana/blob/master/helm/reana/values.yaml) and redeploy
- Run some workflows
- Set some quota limits:
```
kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-set -r disk -e john.doe@example.org -l 500000
kubectl exec -i -t deployment/reana-server -- flask reana-admin quota-set -r cpu -e john.doe@example.org -l 500000
```
- Verify that notifications appear as expected

Couple of examples:
- If user exceeds the quota limit:
```
You've used 105% of your disk quota. Please delete unnecessary workflows.
You've used 105% of your cpu quota. Please contact REANA administrators for quota limit increase.
```
- If user reached critical level (>85%) of the quota limit:
```
You've used 94% of your cpu quota.
You've used 90% of your disk quota. Please consider liberating some disk space by deleting unnecessary workflows.
```
- If quota usage levels are not critical:
```
You have no notifications
```

In the future notifications icon could be used to display things like:
- somebody shared a workflow with you
- "You have 19 running notebook sessions. Please consider closing some."

### DIFF
--- a/reana-ui/src/components/NotificationBell.js
+++ b/reana-ui/src/components/NotificationBell.js
@@ -1,0 +1,58 @@
+/*
+  -*- coding: utf-8 -*-
+
+  This file is part of REANA.
+  Copyright (C) 2022 CERN.
+
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+import { useSelector } from "react-redux";
+import { Link } from "react-router-dom";
+import { Popup, Icon, List } from "semantic-ui-react";
+
+import { getUserQuota } from "~/selectors";
+import { getQuotaNotifications } from "~/pages/profile/components/Quota";
+
+export default function NotificationBell() {
+  const quota = useSelector(getUserQuota);
+  const notifications = getQuotaNotifications(quota);
+  const hasNotifications = !!notifications.length;
+  return (
+    <Popup
+      trigger={
+        <Icon.Group size="large">
+          <Icon link name="bell outline" color="brown" />
+          {hasNotifications && (
+            <Icon corner="top right" name="circle" color="red" />
+          )}
+        </Icon.Group>
+      }
+      content={
+        <List divided relaxed>
+          {!hasNotifications && (
+            <List.Item>You have no notifications</List.Item>
+          )}
+          {notifications.map((notification, index) => (
+            <List.Item key={index} as={Link} to={notification.link}>
+              <List.Icon
+                color="grey"
+                verticalAlign="middle"
+                name={notification.icon}
+              />
+              <List.Content>
+                <List.Header style={{ marginBottom: "0.2em" }}>
+                  {notification.header}
+                </List.Header>
+                <List.Description>{notification.body}</List.Description>
+              </List.Content>
+            </List.Item>
+          ))}
+        </List>
+      }
+      position="bottom right"
+      on="click"
+    />
+  );
+}
--- a/reana-ui/src/components/TopHeader.js
+++ b/reana-ui/src/components/TopHeader.js
@@ -2,7 +2,7 @@
 	-*- coding: utf-8 -*-
 
 	This file is part of REANA.
-	Copyright (C) 2020 CERN.
+	Copyright (C) 2020, 2022 CERN.
 
   REANA is free software; you can redistribute it and/or modify it
   under the terms of the MIT License; see LICENSE file for more details.
@@ -12,19 +12,17 @@ import { Image, Icon, Popup, List } from "semantic-ui-react";
 import { useDispatch, useSelector } from "react-redux";
 import { Link } from "react-router-dom";
 
-import { getConfig, getUserEmail } from "~/selectors";
+import { getUserEmail } from "~/selectors";
 import LogoImg from "~/images/logo-reana.svg";
 import { userSignout } from "~/actions";
+import NotificationBell from "./NotificationBell";
 
 import styles from "./TopHeader.module.scss";
 
 export default function TopHeader() {
   const dispatch = useDispatch();
   const email = useSelector(getUserEmail);
-  const config = useSelector(getConfig);
-  const signOut = () => {
-    dispatch(userSignout());
-  };
+  const signOut = () => dispatch(userSignout());
 
   return (
     <header className={styles["top-header"]}>
@@ -32,11 +30,7 @@ export default function TopHeader() {
         <Image src={LogoImg} size="small" />
       </Link>
       <section>
-        {config.docsURL && (
-          <a href={config.docsURL} target="_blank" rel="noopener noreferrer">
-            <Icon name="question circle outline"></Icon> Help
-          </a>
-        )}
+        <NotificationBell />
         <Popup
           trigger={
             <Icon
--- a/reana-ui/src/pages/profile/components/Quota.js
+++ b/reana-ui/src/pages/profile/components/Quota.js
@@ -2,7 +2,7 @@
   -*- coding: utf-8 -*-
 
   This file is part of REANA.
-  Copyright (C) 2020 CERN.
+  Copyright (C) 2020, 2022 CERN.
 
   REANA is free software; you can redistribute it and/or modify it
   under the terms of the MIT License; see LICENSE file for more details.
@@ -12,13 +12,15 @@ import { useSelector } from "react-redux";
 import { Grid, Label } from "semantic-ui-react";
 import isEmpty from "lodash/isEmpty";
 
-import { PieChart } from "~/components";
-import { getUserQuota } from "~/selectors";
+import { PieChart, Notification } from "~/components";
+import { REANA_QUOTAS_DOCS_URL } from "~/config";
+import { getConfig, getUserQuota } from "~/selectors";
 import { healthMapping } from "~/util";
 
 import styles from "./Quota.module.scss";
 
 export default function Quota() {
+  const config = useSelector(getConfig);
   const quota = useSelector(getUserQuota);
 
   function renderPieChart(title, quota) {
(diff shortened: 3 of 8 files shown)


reana-ui/src/pages/profile/components/Quota.js
```diff
@@ -55,9 +57,52 @@ export default function Quota() {
...
+function getQuotaNotificationInfo(quotaType, exceededLimit, adminEmail) {
+  if (quotaType === "disk" && exceededLimit)
+    return "Please delete unnecessary workflows.";
+  if (quotaType === "disk" && !exceededLimit)
+    return "Please consider liberating some disk space by deleting unnecessary workflows.";
+  if (quotaType === "cpu" && exceededLimit)
+    return (
+      <span>
+        Please contact
+        <a href={`mailto:${adminEmail}`}> REANA administrators</a> for quota
```
mdonadoni, line 85: We should show the mailto link only if `adminEmail` is not `null`

reana-ui/src/components/NotificationIcon.js
```diff
@@ -0,0 +1,48 @@
...
+            <Icon corner="top right" name="circle" color="red" />
+          )}
+        </Icon.Group>
+      }
+      content={
+        <List>
+          {notifications.map((notification, index) => (
+            <List.Item key={index}>{notification}</List.Item>
+          ))}
+        </List>
```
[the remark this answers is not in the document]
audrium: Thanks for a nice suggestion! I've implemented it and added some coloring to the icons. Now it looks like this:
Screenshot from 2022-06-02 10-26-41
Screenshot from 2022-06-02 10-26-01
Screenshot from 2022-06-02 10-23-55

reana-ui/src/components/TopHeader.js
```diff
@@ -12,31 +12,25 @@ import { Image, Icon, Popup, List } from "semantic-ui-react";
...
       <Link to="/">
         <Image src={LogoImg} size="small" />
       </Link>
       <section>
-        {config.docsURL && (
-          <a href={config.docsURL} target="_blank" rel="noopener noreferrer">
-            <Icon name="question circle outline"></Icon> Help
-          </a>
-        )}
+        <NotificationIcon />
```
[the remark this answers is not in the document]
audrium: Yeah, it's very hard to come up with a name for this one. I think adding "old" or "new" might also be confusing, since it will not replace the `Notification` component. Also adding "quota" is not correct, since later it will be used for interactive sessions, etc. I've changed it to `NotificationBell` for now since I think this is a standard way to name this thing.. Let me know if you can come up with something more clear

reana-ui/src/pages/profile/components/Quota.js
```diff
@@ -55,9 +57,52 @@ export default function Quota() {
...
+        limit increase.
+      </span>
+    );
+}
+
+export function getQuotaNotifications(quota, adminEmail, hideLink = true) {
```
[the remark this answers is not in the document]
audrium: I think it's OK to keep this functionality there, since it's very related and directly used in `Quota` component. Placing these functions in `utils.py` might only pollute the file, since it will be used only by `Quota` and `NotificationBell` (which later use additional sources to collect other notifications). We might also create a separate quota utils file, but I don't think it's needed. Please share what do you think

reana-ui/src/pages/profile/components/Quota.js
```diff
@@ -55,9 +57,86 @@ export default function Quota() {
...
+            const message = (
+              <span>
+                {notification.body} Please see:{" "}
+                <a
+                  href={REANA_QUOTAS_DOCS_URL}
+                  target="_blank"
+                  rel="noopener noreferrer"
+                >
+                  {REANA_QUOTAS_DOCS_URL}
+                </a>
```
mdonadoni, line 77: (minor) What do you think about "_For more details, please check out the related [documentation page](https://docs.reana.io/advanced-usage/user-quotas)._" ?

reana-ui/src/pages/profile/components/Quota.js
```diff
@@ -55,9 +57,86 @@ export default function Quota() {
...
+
+function getQuotaNotificationInfo(quotaType, exceededLimit, adminEmail) {
+  if (quotaType === "disk" && exceededLimit)
+    return "Please delete unnecessary workflows or files.";
+  if (quotaType === "disk" && !exceededLimit)
+    return "Please consider liberating disk space by deleting some workflows or files.";
+  if (quotaType === "cpu" && exceededLimit)
+    return (
+      <span>
+        Please contact{" "}
```
mdonadoni, line 107:
```suggestion
        Please contact the{" "}
```

reana-ui/src/pages/profile/components/Quota.js
```diff
@@ -55,9 +57,86 @@ export default function Quota() {
...
+        {adminEmail ? (
+          <a href={`mailto:${adminEmail}`}>REANA administrators</a>
+        ) : (
+          "REANA administrators"
+        )}{" "}
+        for quota limit increase.
```
mdonadoni, line 113:
```suggestion
        for a quota limit increase.
```

reana-ui/src/pages/profile/components/Quota.js
```diff
@@ -55,9 +57,86 @@ export default function Quota() {
...
+      </span>
+    );
+}
+
+export function getQuotaNotifications(quota, adminEmail) {
+  let notifications = [];
+  for (const [key, value] of Object.entries(quota)) {
+    const { usage, limit, health } = value;
+    if (health !== "critical") continue;
+    const percentage = Math.round((usage?.raw / limit?.raw) * 100);
```
mdonadoni, line 123: Why do we need `?.` ?

reana-ui/src/pages/profile/components/Quota.js
```diff
@@ -55,9 +57,86 @@ export default function Quota() {
...
+    const exceededLimit = percentage >= 100;
+    const isDisk = key === "disk";
+    notifications.push({
+      header: `${isDisk ? "Disk" : "CPU"} quota ${
+        exceededLimit ? "exceeded" : "usage"
+      }`,
```
mdonadoni, line 129: When the quota has not been exceeded, what do you think about showing `High CPU/Disk usage` instead of the more generic `CPU/Disk quota usage`?

reana-ui/src/pages/profile/components/Quota.js
```diff
@@ -55,9 +57,86 @@ export default function Quota() {
...
+      icon: isDisk ? "disk" : "microchip",
+      type: exceededLimit ? "critical" : "warning",
```
mdonadoni, line 131: What do you think about using `error` instead of `critical`? By doing so, we use the same terminology of the `Notification` component.

reana-ui/src/pages/profile/components/Quota.js
```diff
@@ -55,9 +57,86 @@ export default function Quota() {
...
+      link: "/profile",
+      body: (
+        <span>
+          You've used {percentage}% of your {key} quota.{" "}
```
mdonadoni, line 135:
```suggestion
          You've used {percentage}% of your { isDisk ? "disk" : "CPU" } quota.{" "}
```

(very minor) WDYT? In this way, we show `CPU` instead of `cpu`.

reana-ui/src/util.js
```diff
@@ -35,7 +35,7 @@ export const statusMapping = {
  */
 export const healthMapping = {
   healthy: "green",
-  warning: "orange",
+  warning: "brown",
```
mdonadoni, line 38: Why do we need to change this from orange to brown?

reana-ui/src/pages/profile/components/Quota.js
```diff
@@ -55,9 +57,86 @@ export default function Quota() {
...
+        for quota limit increase.
+      </span>
+    );
+}
+
+export function getQuotaNotifications(quota, adminEmail) {
+  let notifications = [];
+  for (const [key, value] of Object.entries(quota)) {
+    const { usage, limit, health } = value;
+    if (health !== "critical") continue;
```
mdonadoni, line 122: Should we show a notification even when `health === "warning"`?

reana-ui/src/components/NotificationBell.js
```diff
@@ -0,0 +1,61 @@
...
+        <Icon.Group size="large">
+          <Icon link name="bell outline" color="brown" />
+          {hasNotifications && (
+            <Icon corner="top right" name="circle" color="red" />
+          )}
+        </Icon.Group>
+      }
+      content={
+        <List divided relaxed>
+          {notifications.map((notification, index) => (
```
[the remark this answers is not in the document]
audrium: Nice catch, fixed it

reana-ui/src/components/NotificationBell.js
```diff
@@ -0,0 +1,61 @@
...
+            <List.Item key={index} as="a" href={notification.link}>
+              <List.Icon
+                verticalAlign="middle"
+                name={notification.icon}
+                color={healthMapping[notification.type]}
```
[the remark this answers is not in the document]
audrium: It was a suggestion from @tiborsimko to make a small indication of how critical the notification messages are by coloring the icon. It currently looks like this:
Screenshot from 2022-06-08 10-07-58
Screenshot from 2022-06-08 10-07-30
If we would change it to display the grey color it would look like this:
Screenshot from 2022-06-08 10-10-53

I don't know, both seem to be OK with me, maybe I'm already used to the colored ones. @tiborsimko , WDYT, should we switch back to the grey icons?

reana-ui/src/components/NotificationBell.js
```diff
@@ -0,0 +1,61 @@
...
+              />
+              <List.Content>
+                <List.Header style={{ marginBottom: "0.2em" }}>
```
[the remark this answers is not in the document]
audrium: I think we don't have a rule/convention for this. I personally think that if the component needs a single small css tweak it's ok to do it directly inside the jsx. If there are more than one styling change, then it should go to a dedicated file. Now it would look a bit weird to place this one statement inside a new file. WDYT?

reana-ui/src/pages/profile/components/Quota.js
audrium: For safety, just in case something is wrong with `usage` and `limit` values. E.g. it's `undefined`

reana-ui/src/pages/profile/components/Quota.js
audrium: I've tried to use the same terminology as defined [here](https://github.com/reanahub/reana-commons/blob/master/reana_commons/config.py#L430) and [here](https://github.com/reanahub/reana-ui/blob/031d6dc21163285e7374624c87bf18dfd5287712/reana-ui/src/util.js#L39) and to reuse the type to select the icon color. Indeed if we go for the grey color it could be changed!

reana-ui/src/components/NotificationBell.js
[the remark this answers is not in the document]
mdonadoni: I think we can keep it like this, given that it is the only styling rule in the whole file. We will move it if we add more styling changes in the future

reana-ui/src/pages/profile/components/Quota.js
mdonadoni: My only concern is that, as you said in https://github.com/reanahub/reana-ui/pull/275#discussion_r892324346, this is not coherent with what we get from the server. If the quota status returned by the server is `warning` no notifications are shown, while if `critical` is received it can become either a `warning` or `critical` notification depending on whether we have exceeded the limit or not.

I think we should strive to keep a one-to-one correspondence between what we receive from the server and the notifications. Otherwise, if we are not careful we might get into situations were reana-ui and reana-client report different health statuses to the user, which is confusing. This would also help in having code easier to maintain: if we decide to change the healthy/warning/critical thresholds, we only have to do it on the server-side and not also in reana-ui.

What do you think? How should we proceed?

reana-ui/src/util.js
mdonadoni: > I took the liberty to change it to brown to make it more in line with other colors used inside profile page. Since we introduced notifications there which have brown (warning) and red (error) colors, keeping also the orange there seemed a bit too much. Now this brown color match the warning notification color (you can see the 70% label and disk usage warning in the image bellow).

:+1:

> But indeed there is inconsistency between what we get from the server and what we display. E.g. Critical level is >= 85 percent and we display it as a warning even though the pie chart label color is red

We can continue this discussion in the [other thread](https://github.com/reanahub/reana-ui/pull/275#discussion_r892446483)!

reana-ui/src/pages/profile/components/Quota.js
mdonadoni: Yes, it might be too early to generate notifications at 60%. Maybe we can continue the discussion in the [other thread](https://github.com/reanahub/reana-ui/pull/275#discussion_r892446483), given that this is quite related to that?

reana-ui/src/pages/profile/components/Quota.js
mdonadoni: What I was thinking is that if either `usage?.raw` or `limit?.raw` is undefined, then `percentage` becomes `NaN` but we use it anyway in the following part of code. I am not sure how we should handle this, given that this issue is probably present also in other parts of the code (e.g. [here](https://github.com/reanahub/reana-ui/blob/9d7b4bbcd54b12da901ed7e24a04f347ec6f354e/reana-ui/src/pages/profile/components/Quota.js#L28)). What do you think? Should we address this in another issue?

reana-ui/src/pages/profile/components/Quota.js
audrium: Totally agree that it's better to keep consistent to what we receive from the server side! One easy way to achieve it would be to simply render all critical quota notifications (>=85 percent) in red. It would be in line with what we do for quota pie chart labels. Example:
Screenshot from 2022-06-08 16-35-12
Right now we would show it like this:
Screenshot from 2022-06-08 16-43-45
As you can see it's not matching pie chart label color.. It's kind of nice to distinguish between exceeded and non exceeded notification messages, but we should probably add this new level to server side to keep it consistent everywhere. cc @tiborsimko

reana-ui/src/components/NotificationBell.js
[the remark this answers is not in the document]
audrium: Ok, switched to the grey color!

reana-ui/src/pages/profile/components/Quota.js
audrium: So I've changed the way we calculate health status in [reana-db](https://github.com/reanahub/reana-db/pull/171). Now it's fully consistent and looks like this:
Screenshot from 2022-06-10 11-28-15
Screenshot from 2022-06-10 11-28-53

mdonadoni (approved):
Fantastic! Great job 🚀

PR #284, opened 2022-06-20
wf-list: make chosen sorting criteria visible in dropdown

### LINKED ISSUE #283: ui: selected sort-by value remains hidden when navigating away
**Current behaviour**

When a user selects certain sorting order, for example oldest-first:

sortby1

and then navigates away into the same page or next page, the selection value stays "hidden" and user sees the following:

sortby2

This is problematic because there is no way to see which option was selected.

**Expected behaviour**

One would expect the selected choice to stay visible in the drop-down, as we do for selected "status" value. (Only simpler here because we can pick exactly sort-by criteria and not more.)  

This is what users are accustomed to seeing in regular search engines elsewhere.

The visible values of the drop-down could be perhaps:

Sort by
- latest first
- oldest first
- most cpu used
- most disk used

so that when selected value of the drop-down is visible, it would make sense as a standalone verb (and ideally fitting into one line; currently "most disk quota used" is two lines; we could experiment with best looks and other cosmetics).

mdonadoni:
Closes #283


reana-ui/src/pages/workflowList/components/WorkflowSorting.js
```diff
@@ -12,16 +12,15 @@ import PropTypes from "prop-types";
 import { Dropdown } from "semantic-ui-react";
 
 const sortOptions = [
-  { key: 1, text: "Newest", value: "desc" },
-  { key: 2, text: "Oldest", value: "asc" },
-  { key: 3, text: "Most disk quota used", value: "disk-desc" },
-  { key: 4, text: "Most cpu quota used", value: "cpu-desc" },
+  { key: 1, text: "Latest first", value: "desc" },
+  { key: 2, text: "Oldest first", value: "asc" },
+  { key: 3, text: "Most disk used", value: "disk-desc" },
```
audrium, line 17: Minor: we could capitalize the name and use `Disk` like we do for quota notifications, etc ([example](https://user-images.githubusercontent.com/4990025/173036195-bbd644da-5254-467a-aa7c-0b2ae4c1aaf9.png))

audrium (approved):
Works nicely! Reduced font size is still ok, text fits inside the input box nicely when resizing!

PR #293, opened 2022-08-12
wf-details: improve retention message

tiborsimko:
Improves retention message verbiage for new, not yet run workflows.

Status before:

retention-verbiage-finishes-before

Status after:

retention-verbiage-finishes-after


mdonadoni (approved):
LGTM!

PR #294, opened 2022-08-12
wf-details: improve approximate time before retention rule

### LINKED ISSUE #291: ui: improve approximate time announcements in retention messages
**Current behaviour**

When someone configures a retention period of 10 days, and inspects the workflow that freshly finished, the web UI shows:

retention-week

This seems to send a contradictory message about when the files will be deleted: in 10 days? or in a week? This may be confusing especially since the workflow run has just freshly terminated.  

Note that we do show exact time to users, if they hover over "in a week", they would see the exact time being indicated. However, the phrase "In a week" looks formatted just like an ordinary text, not something users would expect to go and hover over. So I think majority won't notice.

Hence the feeling of "apparent contradiction".

**Expected behaviour**

For matters as important as removing workflow files, we should be more precise here. We have two possible directions:

1. **Softening the language.** Since our time display library shows "approximate verbiage" for date time points, we can soften the estimation and print "in about a week" instead of "In a week".

2. **Using more exact times.** Instead of "In a week", we can print more exact timing,  such as "After 2022-08-22 23:59". Note that I'm keeping an element of uncertainty here by means of the word "after", since our retention deletion daemon is not guaranteed to run at an exact hour or minute. Also, we could retain hover, and print "in a week" when hovering.  (This would be basically inverting current normal text <-> hover text display.)

3. **Combining the two instead of hover.** We can combine estimate and exact into one phrase, and display something like "2022-08-22 23:59 (in a week)" which would solve the "undiscoverable hover" UX problem mentioned above.

In my eyes, the file deletion is quite an important event, so I would tend to displaying exact timings first and approximate verbiage later; and since the text is printed in a normal font, I'd probably vote for 3 removing hover.

mdonadoni:
Closes #291

### DIFF
--- a/reana-ui/src/pages/workflowDetails/components/WorkflowRetentionRules.js
+++ b/reana-ui/src/pages/workflowDetails/components/WorkflowRetentionRules.js
@@ -48,9 +48,11 @@ export default function WorkflowRetentionRules({ id }) {
     if (!nextRule) {
       return "Retention rules";
     }
-    const { workspaceFiles, retentionDays, timeBeforeExecution } = nextRule;
-    const when =
-      timeBeforeExecution || `${retentionDays} days after workflow finishes`;
+    const { workspaceFiles, retentionDays, applyOn, timeBeforeExecution } =
+      nextRule;
+    const when = applyOn
+      ? `after ${applyOn} (${timeBeforeExecution})`
+      : `${retentionDays} days after workflow finishes`;
     return (
       <>
         Note: files matching the pattern{" "}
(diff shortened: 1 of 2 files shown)


tiborsimko (commented):
Works nicely. One cosmetic comment: the displayed time reads "22nd Aug 2022 23:59" while I guess I would prefer to see "2022-08-22T23:59:00" for better consistency with the other modified file time stamps that are being displayed in the file listing table below. If people eyeball or scan the page rapidly, it'll be nice to see the same YYYY-MM-DDThh:mm:ss format pattern everywhere, saving some contextual visual fatigue switch. (And we could perhaps display "22nd Aug 2022 23:59" in a hover, as we do for workflow termination times such as "Finished 3 hours ago"?)

mdonadoni:
> (And we could perhaps display "22nd Aug 2022 23:59" in a hover, as we do for workflow termination times such as "Finished 3 hours ago"?)

I think it is not needed, given that it doesn't really add any more information or details. What do you think?

tiborsimko:
> I think it is not needed, given that it doesn't really add any more information or details. What do you think?

Fully agreed. BTW do we add "(in a week)" also to the title?

retention-after-in-a-week

tiborsimko (approved):

PR #295, opened 2022-09-21
wf-details: improve retention rules table

### LINKED ISSUE #292: ui: improve retention block display style
**Current behaviour**

When users expand the retention info block on the workflow workspace tab, they see:

retention-table

This has some UX issues in my eyes, because it is visually hard to distinguish "two independent table blocks" on this page, the workspace retention block vs the file listing block. This would be especially apparent when users dedine several workspace retention rules.

(Also, note the first table heading background is being displayed in grey, while the second table heading background is being displayed in white. So the header items of the retention table seem to be given visually more prominence when compared to the headers of the file listing  table, while it is more of a "additional information" content rather than the "main content" of this tab.)

**Expected behaviour**

We should try to better distinguish the two blocks visually, so that it is easier to see the page overview at a glance.

For example, we can print the workspace retention block in bluish foreground tones on bluish background, similarly to boxes we have elsewhere:

retention-table-gitlab

In this way, it would be easier to distinguish the two blocks, and the retention information would use a blue theme "additional info" style, whilst the workspace file listing information would use the regular "main content" style. 

(Also, we could harmonise the relative shading of table header rows when compared to table content rows, but that's another minor cosmetic issue.)

P.S. The goal of this issue is to design a graphical mean to better visually distinguish the two blocks. Using bluish theme for the retention block was just one example how we could achieve this relatively easily, We could also go for other alternative proposals to achieve the distinction.

mdonadoni:
Closes #292

### DIFF
--- a/reana-ui/src/pages/workflowDetails/components/WorkflowRetentionRules.js
+++ b/reana-ui/src/pages/workflowDetails/components/WorkflowRetentionRules.js
@@ -68,7 +68,7 @@ export default function WorkflowRetentionRules({ id }) {
         {generateTitle(nextRule)}
       </Accordion.Title>
       <Accordion.Content index={0} active={showRules}>
-        <Table compact>
+        <Table compact basic="very" className={styles["rules-table"]}>
           <Table.Header>
             <Table.Row>
               <Table.HeaderCell>Workspace files</Table.HeaderCell>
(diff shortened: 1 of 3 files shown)


tiborsimko:
Works nicely :+1: One comment: the graphical presentation of the file pattern is still in grey, which cuts the colour harmony a bit. What about presenting the file patterns in the regular text style? Something like:

Before:

pattern-before

After:

pattern-after

tiborsimko:
> What do you think about using a label with colors matching the ones of the table?

That would also do, obviously! Provided that such a "passive text" looks sufficiently different from a "clickable" button :wink:

mdonadoni:
Current style:

tiborsimko (approved):
Works nicely :+1:

PR #306, opened 2022-12-08
ui: allow markup in announcement

### LINKED ISSUE #304: announcement: usage of HTML markup in the announcement text
**Current behaviour**

Currently, the announcement system does not seem to allow _easily_ to use some HTML markup in the announcement message.

For example:

```yaml
components:
  reana_ui:
    announcement: "<strong>Wednesday December 7th 2022 08:00:00 CET:</a> REANA will be upgraded to latest release bringing several new features. (<a href=\"https://cern.service-now.com/service-portal?id=outage&n=OTG0074500\">OTG0074500</a>)"
```

**Expected behaviour**

We may want to make it easy for administrators to use _some_ markup in order to display selected important information such as intervention dates in bold, allow linking to external pages such as OTG notices, etc.

The markup should be considered largely "trusted" for display, as it should be the responsibility of the cluster administrator to make sure that the HTML markup is reasonably small and displays correctly.

However, we should also pay attention to retain safety and prevent possible exploits via unsafe access to config maps etc.

**Note**

This issue is created to check how we escape and display helm values for the announcement message, and see whether we can create an easy way to allow _some_ HTML markup for the admins to use, all the while retaining good safety against theoretical injecting of unsafe markup.

mdonadoni:
Closes #304

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -21,6 +21,7 @@ Version 0.9.0 (UNRELEASED)
 - Changes the workflow-details page to show the workspace's retention rules.
 - Changes the workflow-details page to show the duration of the workflow's jobs.
 - Changes the deletion of a workflow to always clean up the workspace.
+- Changes the announcements to support limited HTML markup.
 
 Version 0.8.2 (2022-02-15)
 ---------------------------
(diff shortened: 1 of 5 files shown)


audrium (approved):
Works nicely, we could also drop a message in the changelog for this feature

Screenshot from 2022-12-08 13-44-20

PR #309, opened 2023-01-31
ui: suggest deleting also an open session in a deletion modal

audrium:
closes https://github.com/reanahub/reana-client/issues/649


reana-ui/src/components/WorkflowDeleteModal.js
```diff
@@ -38,9 +38,9 @@ export default function WorkflowDeleteModal() {
           <Message.Content>
             <Message.Header>Workspace deletion!</Message.Header>
```
mdonadoni, line 39: What do you think about changing also the header to something like "Deletion of workspace and interactive sessions!"?

PR #311, opened 2023-02-09
auth: add option for generic OIDC SSO to login

goseind:
Refer to https://github.com/reanahub/reana-server/issues/513 for details.

Fixes https://github.com/reanahub/reana-server/issues/513


reana-ui/.eslintrc.json
```diff
@@ -14,7 +14,7 @@
     "prettier/react"
   ],
   "rules": {
-    "prettier/prettier": "error",
+    "prettier/prettier": "off", // turn of as it causes unnecessary errors
```
mdonadoni, line 17: We should revert this back to the original value. You can avoid CI errors by running `yarn fmt` before committing and pushing to GitHub 😄

reana-ui/src/pages/signin/Signin.js
```diff
@@ -28,14 +28,14 @@ export default function Signin() {
   const [formData, setFormData] = useState({ email: "", password: "" });
   const location = useLocation();
 
-  const handleClick = () => {
+  const handleClick = (url) => {
```
mdonadoni, line 31: I think we should change the name of `url` to something more descriptive, what do you think about `ssoProvider` or something similar?

reana-ui/src/pages/signin/Signin.js
```diff
@@ -59,9 +59,21 @@ export default function Signin() {
...
         {config.cernSSO && (
           <>
-            <Button basic fluid size="large" onClick={handleClick}>
+            <Button basic fluid size="large" onClick={() => handleClick("cern_openid")}>
               Sign in with CERN Single Sign-On
             </Button>
+            {(config.localUsers || config.keycloakSSO) && (
+              <Divider section horizontal>
+                or
+              </Divider>
```
mdonadoni, line 68: I think it would look better not putting the divider between the CERN SSO and Keycloak SSO buttons, but only before the login form for local users. What do you think? 

This issue is very minor, as I do not expect both CERN SSO and Keycloak SSO to be enabled at the same time.

reana-ui/src/pages/signin/Signin.js
```diff
@@ -59,9 +59,21 @@ export default function Signin() {
...
+            )}
+          </>
+        )}
+        {config.keycloakSSO && (
+          <>
+            <Button basic fluid size="large" onClick={() => handleClick("keycloak")}>
+              Sign in with {config.keycloakTitle} Single Sign-On
```
mdonadoni, line 75: Maybe we should allow administrators to customize the whole text?  For example, they might want to show `Sign in` or `Sign in with ESCAPE IAM`. /cc @tiborsimko

reana-ui/src/client.js
```diff
@@ -17,8 +17,8 @@ import { stringifyQueryParams } from "~/util";
 export const CONFIG_URL = `${api}/api/config`;
 export const USER_INFO_URL = `${api}/api/you`;
 export const USER_SIGNUP_URL = `${api}/api/register`;
-export const USER_OAUTH_SIGNIN_URL = (next) =>
-  `${api}/api/oauth/login/cern_openid?${stringifyQueryParams({ next })}`;
+export const USER_OAUTH_SIGNIN_URL = (next, sso) =>
```
mdonadoni, line 20: Extremely minor comment: I think it might be more "natural" to first provide the name of the SSO provider and then the `next` redirection URL?

goseind: Can you provide an example, please? Not sure I understand.

reana-ui/src/pages/signin/Signin.js
goseind: Allowing just the `title` to be changed would give more certainty to the layout of the text I'd say. But you guys decide 😃

reana-ui/src/pages/signin/Signin.js
goseind: How about just a divider without text? I think in our instance we'll allow both options.

reana-ui/.eslintrc.json
goseind: :see_no_evil: didn't mean to commit that, but it was showing me errors of things I didn't change myself, anyways fixed them now 😄

reana-ui/src/pages/signin/Signin.js
tiborsimko: For now, I guess it is enough to show the login provider name. It should be enough to boot. 

(And, if a need arises to display more text, and if we go for the `login` provider list in the Helm chart as discussed in the other PR, then we could simply add a new "login_button_text" or some such amongst properties there, should this need really arise.)

BTW see also the styling of the Zenodo login, using also simple layout with just login provider name changed:

zenodo-login

The "or" separator is between external and internal login techniques... But for REANA, we would typically use only SSO or only Keycloak or only Local, so there would be typically only one button. (And later down the line, we could deprecate CERN SSO and move to CERN Keycloak, and even later down the time line, we could even deprecate Local and use only Keycloak, so that REANA would be single-user out-of-the-box , and multi-user when somebody adds Keycloak.)

reana-ui/src/pages/signin/Signin.js
```diff
@@ -59,9 +59,28 @@ export default function Signin() {
...
+              basic
+              fluid
+              size="large"
+              onClick={() => handleClick("cern_openid")}
+            >
               Sign in with CERN Single Sign-On
             </Button>
+            {(config.localUsers || config.loginEnabled) && <Divider></Divider>}
+          </>
+        )}
```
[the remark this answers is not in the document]
goseind: I removed it and added a small margin instead as otherwise, the buttons are directly beneath each other, which does not look very appealing. Also changed the logic to display the local user OR divider if either button is configured.

PR #319, opened 2023-05-03
wf-details: make list of steps scrollable

### LINKED ISSUE #318: wf-details: make list of steps scrollable
In the "Job logs" tab of the workflow details page, the dropdown selector used to choose the desired workflow step is not scrollable, even though some workflow steps are not visible.

As an example, in this case there are many more steps which are hidden (e.g. merge_0, merge_1, ...):

giuseppe-steduto:
Closes #318


mdonadoni (approved):
LGTM!

PR #326, opened 2023-05-22
wf-details: add preview of ROOT files

### LINKED ISSUE #322: preview: support previewing ROOT files
Physics workflows' output and temporary data files are often produced in ROOT format.

These cannot be currently previewed when browsing the Workspace tab files.

Let's investigate adding support for previewing ROOT files using e.g. [jsroot](https://github.com/root-project/jsroot).

mdonadoni:
Closes #322

How to test:
- Run reana-demo-alice-pt-analysis
- Open the file `AnalysisResults.root`

### DIFF
--- a/reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
+++ b/reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
@@ -2,37 +2,39 @@
 	-*- coding: utf-8 -*-
 
 	This file is part of REANA.
-	Copyright (C) 2020, 2021, 2022 CERN.
+	Copyright (C) 2020, 2021, 2022, 2023 CERN.
 
   REANA is free software; you can redistribute it and/or modify it
   under the terms of the MIT License; see LICENSE file for more details.
 */
 
-import { useEffect, useState } from "react";
-import { useDispatch, useSelector } from "react-redux";
-import PropTypes from "prop-types";
+import { buildGUI } from "jsroot";
+import { uniqueId } from "lodash";
 import sortBy from "lodash/sortBy";
+import PropTypes from "prop-types";
+import { useEffect, useRef, useState } from "react";
+import { useDispatch, useSelector } from "react-redux";
 import {
   Button,
   Icon,
+  Loader,
+  Message,
   Modal,
   Segment,
   Table,
-  Loader,
-  Message,
 } from "semantic-ui-react";
 
+import { fetchWorkflowFiles } from "~/actions";
+import client, { WORKFLOW_FILE_URL } from "~/client";
+import { Pagination, Search } from "~/components";
+import { applyFilter } from "~/components/Search";
 import {
   getWorkflowFiles,
   getWorkflowFilesCount,
   loadingDetails,
 } from "~/selectors";
-import { fetchWorkflowFiles } from "~/actions";
-import client, { WORKFLOW_FILE_URL } from "~/client";
 import { getMimeType } from "~/util";
 import { WorkflowRetentionRules } from ".";
-import { Pagination, Search } from "~/components";
-import { applyFilter } from "~/components/Search";
 
 import styles from "./WorkflowFiles.module.scss";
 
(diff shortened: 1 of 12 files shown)


mdonadoni:
Currently there is some padding around the preview of ROOT files and of PDF files. Should I remove it?





Also, what do you think about moving all the logic regarding the preview of files (e.g. checking constraints, fetching the file, how to display each type of file, etc.) to a new React component, so that it is not mixed with the logic of the "Workspace" tab (e.g. file sorting, etc.)?

tiborsimko:
Works nicely :+1: In view of longer term, it'll be really good to separate the previewer logic into a new component.

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
```diff
@@ -79,12 +81,54 @@ const PREVIEW_MIME_PREFIX_WHITELIST = {
   },
   "application/pdf": {
     serverPreviewable: true,
-    display: (content) => {
-      return <object data={content} className={styles["pdf-object"]} />;
-    },
+    display: (content, name) => (
+      <object data={content} className={styles["pdf-object"]}>
+        <a href={content} target="_blank" rel="noopener noreferrer">
```
giuseppe-steduto, line 86: This link should work as a fallback in case the browser is not able to display the PDF file, right? If that's the case, what do you think about displaying a `Message` component (just like is done with other non-previewable types of files) instead of just the link?

giuseppe-steduto (commented):
LGTM! I just left a small comment about the fallback for the PDF preview.

giuseppe-steduto:
I very slightly prefer the version without padding, both for the ROOT file preview and for the PDF preview.
Also, I agree on extracting the logic of previewing the files to a new React component.

mdonadoni:
I have just found out that the final js bundle is much bigger than before (sizes reported by `yarn build` after gzip):
Before
```
225.62 kB  build/static/js/main.bebbe785.js
101.64 kB  build/static/css/main.730fe37a.css
```
After
```
735.92 kB (+510.31 kB)  build/static/js/main.1debd93c.js
101.71 kB (+68 B)       build/static/css/main.0c6eca96.css
[many other smaller files...]
```

The build time has also increased:
- `yarn` 2:51 to 4:17
- `yarn build` 28s to 1:02

I will investigate how to improve bundle size and/or build time

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
```diff
@@ -0,0 +1,213 @@
...
+
+/**
+ * Check if the given file name matches any given mime-type
+ * @param {Array} list Array of mime-types to check against
+ * @param {String} fileName File name to check
+ * @return {Boolean} Extension that matches the file name
+ */
+function matchesMimeType(list, fileName) {
+  const mimeType = getMimeType(fileName);
+  return mimeType && list.find((ext) => mimeType.startsWith(ext));
```
giuseppe-steduto, line 127:
```suggestion
  return (
    mimeType &&
    (list.find((ext) => mimeType === ext) ||
      list.find((ext) => mimeType.startsWith(ext)))
  );
```

What do you think about this approach? In this way, exact matches in the MIME types have "priority" over partial matches, regardless of the order in the `PREVIEW_MIME_PREFIX_WHITELIST` map (e.g. if `"text/"` comes before `"text/html"`.

reana-ui/src/util.js
```diff
@@ -230,6 +230,11 @@ export function getMimeType(fileName) {
   if (WHITELIST.find((ext) => fileName.endsWith(ext))) {
     return "text/plain";
   }
+  if (fileName.endsWith(".root")) {
+    // ROOT does not have a registered MIME-type yet
+    // See https://github.com/root-project/root/issues/6771
+    return "application/x-root";
+  }
```
giuseppe-steduto, line 237: What do you think about using the [`define`](https://github.com/broofa/mime#mimedefinetypemap-force--false) method from the `mime` module? We could also use it to define in a flexible way further non-standard mappings (for example, `.sh` files are not previewed by REANA)
```javascript
mime.define(
    {
      "application/x-root": ["root"],
      "text/x-shellscript": ["sh"],
    },
    true
  );
```

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
```diff
@@ -0,0 +1,213 @@
...
+        size={size}
+        url={getFileURL(workflow, fileName)}
+      />
+    );
+  }
+
+  return (
+    <Modal open closeIcon onClose={onClose ? onClose : () => {}}>
+      <Modal.Header>{fileName}</Modal.Header>
+      <Suspense
```
giuseppe-steduto, line 185: Since, according to the docs, "Only Suspense-enabled data sources will activate the Suspense component", such as the components loaded with `lazy` (see [the note in this page](https://react.dev/reference/react/Suspense#:~:text=Show%20less-,Note,-Only%20Suspense%2Denabled)), maybe it makes more sense to move this inside the `ROOTPreview` component. Another option would be to create some sort of wrapper around the `_request` method to make it communicate with `Suspense`, but that's something we can maybe address in another issue. What do you think?

giuseppe-steduto (commented):
Nice work! I like way better the refactored version that extracts the preview component and everything works as expected. I just left a few small comments / ideas, mainly on the detection of the MIME types

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
```diff
@@ -0,0 +1,213 @@
...
+  "application/x-root": ROOTPreview,
+};
+
+const FILE_SIZE_LIMIT = 5 * 1024 ** 2; // 5MB
+
+/**
+ * Check if the given file name matches any given mime-type
+ * @param {Array} list Array of mime-types to check against
+ * @param {String} fileName File name to check
+ * @return {Boolean} Extension that matches the file name
```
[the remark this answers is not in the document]
mdonadoni: Updated also taking into account the new changes

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
mdonadoni: I have updated this to return the longest matching "mime-type prefix"

reana-ui/src/util.js
mdonadoni: Done! I have created a new `Mime` local object, so that we do not modify the global `mime` variable provided by the library

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
mdonadoni: It's true that only `ROOTPreview` uses it for now, but future previewer might be lazy loaded for the same reason as jsroot (i.e. big dependency). For this reason, I prefer having `Suspense` outside of `ROOTPreview`, so that the fallback `Loader` is defined only once (and not into every lazy loaded component).

Regarding `_request` being "Suspense-enabled", I think that might be a cool feature, but that needs quite a lot of work so it's probably better to investigate and implement it in another issue/PR.

giuseppe-steduto (approved):
LGTM!

PR #334, opened 2023-07-11
nginx: serve gzip-compressed static files

### LINKED ISSUE #328: build: consider compressing served static files
After adding `jsroot` to preview ROOT files, JS source files have gotten bigger, see https://github.com/reanahub/reana-ui/pull/326#issuecomment-1564485956. This might get worse in the future when new dependencies will be added to the project. 

In that same PR, we have already implemented lazy loading to avoid loading `jsroot` when not necessary.
However, we should also consider compressing static files before serving them:
- check which format to use (depending on popularity, browser support, server side support, performance, ...)
- decide whether it's worth it to enable compression
- start serving compressed static files (JS sources, but also CSS), either:
  - compressed on-the-fly at runtime (see [NGINX docs about compression](https://docs.nginx.com/nginx/admin-guide/web-server/compression/))
  - generating already compressed sources at build-time and serving them (see [`gzip_static` directive of NGINX](https://docs.nginx.com/nginx/admin-guide/web-server/compression/#sending-compressed-files) as an example)

mdonadoni:
Compress JS/CSS files with gzip after `yarn build` and configure nginx
to serve the pre-compressed files.

Closes #328

### DIFF
--- a/nginx/reana-ui.conf
+++ b/nginx/reana-ui.conf
@@ -3,6 +3,10 @@ server {
     root   /usr/share/nginx/html;
     server_name  localhost;
     index  index.html index.htm;
+
+    #Enable serving of pre-compressed files
+    gzip_static on;
+
     location / {
         try_files $uri /index.html;
     }
--- a/reana-ui/package.json
+++ b/reana-ui/package.json
@@ -34,7 +34,8 @@
     "fmt": "prettier --write .",
     "ci": "run-p lint prettier",
     "eject": "craco eject",
-    "postinstall": "semantic-ui-css-patch"
+    "postinstall": "semantic-ui-css-patch",
+    "postbuild": "find build \\( -name '*.js' -o -name '*.css' \\) -exec gzip -k9S .gz {} \\;"
   },
   "browserslist": {
     "production": [


mdonadoni:
Regarding the comments made in the related issue

> check which format to use (depending on popularity, browser support, server side support, performance, ...)

I have chosen gzip as it is supported out-of-the-box by the `nginx` docker image, while also being the most popular format used on the web (see [here](https://almanac.httparchive.org/en/2021/compression#trends-in-http-compression)). Brotli would need an additional extension which is not included in the official `nginx` docker image.

> decide whether it's worth it to enable compression

When opening the homepage of REANA (with cache disabled), the transferred size goes from 1,64MB to 620kB (-62%). When opening a ROOT file, the additional JS file needed goes from 1,78MB to 510kB (-71%).
> start serving compressed static files (JS sources, but also CSS), either:
> 
>   * compressed on-the-fly at runtime (see [NGINX docs about compression](https://docs.nginx.com/nginx/admin-guide/web-server/compression/))
> 
>   * generating already compressed sources at build-time and serving them (see [`gzip_static` directive of NGINX](https://docs.nginx.com/nginx/admin-guide/web-server/compression/#sending-compressed-files) as an example)

Given that nginx only serves static files, it makes more sense to pre-compress them instead of compressing them on-the-fly at each request. For this reason, the gzipped-version of JS/CSS files are generated at build time.

giuseppe-steduto (approved):
Nice job and good performance optimization! Also, gzip seems widely supported by every browser, so I personally approve this choice, and having both versions of the files together with `gzip_static` set to `on` (rather than `always`) makes the UI still compatible even with the (rare) clients that do not support / accept gzip.

PR #335, opened 2023-07-11
config: proxy requests to backend when using dev server

### LINKED ISSUE #320: dev: login not working in debug mode
Logging in when using the development server (`yarn start`) is not working in Chrome (version 112.0.5615.137).

The login flow works well, but all the subsequent requests to the REST API return an error (401), as the session cookie is blocked by Chrome:

mdonadoni:
Configure the development server to proxy `/api` requests to
`$REANA_SERVER_URL`, instead of making direct requests to
`$REANA_SERVER_URL` from the client. This means that all the requests,
including the ones to the REST API, are now made to the same origin.
This improves the situation with regard to mechanisms like CORS or
SameSite cookies.

Closes #320


giuseppe-steduto (commented):
This PR works as intended, but introduces an error when trying to open an interactive session link, as clicking on the Jupyter icon now points to the wrong URL in developer mode. Since the `INTERACTIVE_SESSION_URL` function in `client.js` returns an URL that does not point to `/api/...`, but rather directly to `${sessionUrl}`, and since with this PR all the requests are made to the local React server, the HTTP request to the interactive session link is not made anymore to `$REANA_SERVER_URL`, as was instead done before.

The quickest fix that comes to my mind is to keep the `.env.development` file and change `client.js` as follows:
```diff
- export const INTERACTIVE_SESSION_URL = (sessionUri, reanaToken) =>  `${api}${sessionUri}?token=${reanaToken}`;
+ export const INTERACTIVE_SESSION_URL = (sessionUri, reanaToken) =>  `${reanaServerApi}${sessionUri}?token=${reanaToken}`;
```
where `reanaServerApi` is a new constant and `reanaServerApi = process.env.REACT_APP_SERVER_URL || window.location.origin;`
But perhaps there's a cleaner solution.

mdonadoni:
Fixed by matching the URLs of Jupyter notebooks. The pattern used to match them is not perfect, as it matches "more" than it should, but I think that's good enough for the development server.

giuseppe-steduto (approved):
I agree, good enough for the development server. Seems to work perfectly now!

PR #338, opened 2023-07-18
util: change wf duration calculation depending on status

### LINKED ISSUE #337: Time after which the workflow is stopped incorrectly updated
The timer that usually shows the workflow duration, and that should say after how much time a workflow has been stopped doesn't pause when the workflow is stopped. This results in the "stopped after X minutes Y seconds" phrase with X and Y constantly increasing as time passes. This happens both in the timer of the workflow and in those of the jobs:

image
(In this picture the workflow was stopped a few seconds after it was started, but the screenshot was taken 35 minutes later: the time kept increasing)

How to reproduce:
1. Stop a running workflow (e.g. with the `reana-client stop -w workflow --force` command): the timer that was saying `running for X minutes Y seconds` now says `stopped after X minutes Y seconds`.
2. After some minutes, the same timer will say `stopped after W minutes Z seconds`, constantly increasing as time passes.

This is likely due to the fact that we always calculate the duration of a workflow in the same way, regardless of the workflow (or job) status:

```js
workflow.duration = formatDuration(
      getDuration(startedMoment, finishedMoment)
    );
```

giuseppe-steduto:
Change the way in which the displayed workflow duration is calculated
depending on the status of the workflow: if the workflow is finished or
deleted, take as "finished" moment the value `run_finished_at`, but if
the workflow is stopped, consider `run_stopped_at` as the ending moment
relative to which the duration is calculated.

Closes #337

### DIFF
--- a/reana-ui/src/util.js
+++ b/reana-ui/src/util.js
@@ -160,6 +160,13 @@ function parseWorkflowDates(workflow) {
   const createdMoment = moment.utc(workflow.created);
   const startedMoment = moment.utc(workflow.progress.run_started_at);
   const finishedMoment = moment.utc(workflow.progress.run_finished_at);
+  const stoppedMoment = moment.utc(workflow.progress.run_stopped_at);
+  // Mapping between workflow status and the end moment to use for calculating the duration
+  const endMomentStatusMapping = {
+    finished: finishedMoment,
+    stopped: stoppedMoment,
+    deleted: finishedMoment.isValid() ? finishedMoment : stoppedMoment,
+  };
   workflow.createdDate = createdMoment.format("Do MMM YYYY HH:mm");
   workflow.startedDate = startedMoment.format("Do MMM YYYY HH:mm");
   workflow.finishedDate = finishedMoment.format("Do MMM YYYY HH:mm");
@@ -177,7 +184,7 @@ function parseWorkflowDates(workflow) {
     }
 
     workflow.duration = formatDuration(
-      getDuration(startedMoment, finishedMoment)
+      getDuration(startedMoment, endMomentStatusMapping[workflow.status])
     );
   }
   return workflow;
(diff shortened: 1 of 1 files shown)


reana-ui/src/util.js
```diff
@@ -160,6 +160,13 @@ function parseWorkflowDates(workflow) {
   const createdMoment = moment.utc(workflow.created);
   const startedMoment = moment.utc(workflow.progress.run_started_at);
   const finishedMoment = moment.utc(workflow.progress.run_finished_at);
+  const stoppedMoment = moment.utc(workflow.progress.run_stopped_at);
+  // Mapping between workflow status and the end moment to use for calculating the duration
+  const endMomentStatusMapping = {
+    finished: finishedMoment,
+    stopped: stoppedMoment,
+    deleted: finishedMoment,
```
mdonadoni, line 168: This does not work if the workflow is first stopped and then deleted, as `run_finished_at` is not set and the duration keeps increasing. What do you think about calculating the duration like this? (not tested)
```js
workflow.duration = formatDuration(
  getDuration(startedMoment, stoppedMoment || finishedMoment)
);
```

reana-ui/src/util.js
```diff
@@ -190,7 +197,15 @@ export function parseLogs(logs) {
   const parsedLogs = JSON.parse(logs);
 
   for (let job of Object.values(parsedLogs.job_logs)) {
-    job.duration = formatDuration(getDuration(job.started_at, job.finished_at));
+    if (job.status === "stopped") {
+      // Hide the duration of the job, because in the job object
+      // there is no info about when the job (or the workflow) was stopped.
+      job.duration = "";
```
mdonadoni, line 203:
```suggestion
      job.duration = null;
```

What about `null` instead of `""`?  We already set the duration to null if it is not available or it is not valid:
https://github.com/reanahub/reana-ui/blob/78153048f3479c32d56b46ddb0343b1af9fa4b64/reana-ui/src/util.js#L118-L127

reana-ui/src/util.js
giuseppe-steduto: This would not work because `stoppedMoment` is not evaluated to `false` even if `workflow.progress.run_stopped_at` is `null`, because it still gets converted to a (invalid) `Moment` object. I also liked the mapping of the statuses to the "end moment", especially for readability and extensibility reasons, in case we ever wanted to display something like "queued for X minutes", but thinking about it idk how useful that could be. Anyway, I ended up slightly changing the way in which the end moment for the deleted status is calculated:
```js
deleted: finishedMoment.isValid() ? finishedMoment : stoppedMoment
```
What do you think?

mdonadoni (approved):
Looks good!

PR #340, opened 2023-07-27
preview: get maximum file size from config

mdonadoni:
Closes reanahub/reana#728

### DIFF
--- a/reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
+++ b/reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
@@ -11,9 +11,11 @@
 import sortBy from "lodash/sortBy";
 import PropTypes from "prop-types";
 import { Suspense, lazy, useEffect, useState } from "react";
+import { useSelector } from "react-redux";
 import { Button, Icon, Loader, Message, Modal } from "semantic-ui-react";
 
 import client, { WORKFLOW_FILE_URL } from "~/client";
+import { getConfig } from "~/selectors";
 import { getMimeType } from "~/util";
 
 import styles from "./FilePreview.module.scss";
@@ -115,8 +117,6 @@ const PREVIEW_MIME_PREFIX_WHITELIST = {
   "application/x-root": ROOTPreview,
 };
 
-const FILE_SIZE_LIMIT = 5 * 1024 ** 2; // 5MB
-
 /**
  * Check if the given file name matches any given mime-type prefix
  * @param {Array} list Array of mime-type prefixes to check against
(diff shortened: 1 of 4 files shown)


reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
```diff
@@ -151,9 +151,9 @@ function checkConstraints(fileName, size) {
   if (!match) {
     const fileExt = fileName.split(".").pop();
     content = `${fileExt} files cannot be previewed. Please use download.`;
-  } else if (size > FILE_SIZE_LIMIT) {
+  } else if (size > sizeLimit) {
     content = `File size is too big to be previewed (limit ${
-      FILE_SIZE_LIMIT / 1024 ** 2
+      sizeLimit / 1024 ** 2
```
giuseppe-steduto, line 156: We may want to round this number to show only one or two decimal digits at max - if `config.filePreviesSizeLimit` is not a multiple of a MiB, the result of the division may show a lot of decimal places.

giuseppe-steduto (commented):
Looks good! Just left a small comment on the formatting of the message in case the file is too big to be previewed.

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
mdonadoni: I have updated the PR to show the limit in human-readable format, basically the same as [`_human_readable_bytes`](https://github.com/reanahub/reana-db/blob/02c0e4a93c46bfaa302dd7d84647a92d1673e8c2/reana_db/models.py#L1018) from reana-db

reana-ui/src/util.js
```diff
@@ -237,6 +237,34 @@ export function formatSearch(term) {
...
+ * @param {Number} size File size in bytes.
+ * @returns The formatted human-readable file size.
+ */
+export function formatFileSize(size, digits = 2) {
+  if (size === 0) {
+    return "0 Bytes";
+  }
+
+  const units = [
+    "Bytes",
```
giuseppe-steduto, line 251: very minor: what about removing the capital letter (`bytes`) or maybe using directly the symbol `B`?

giuseppe-steduto (approved):
LGTM!

reana-ui/src/util.js
mdonadoni: I also like `bytes` or `B` more, but I kept `Bytes` to be consistent with reana-db:
```python
    @staticmethod
    def _human_readable_bytes(bytes_):
        """Convert bytes usage to human readable string."""
        if bytes_ == 0:
            return "0 Bytes"
        units = ["Bytes", "KiB", "MiB", "GiB", "TiB", "PiB", "EiB", "ZiB", "YiB"]
        ...
```

mdonadoni:
Rebased and merged!

PR #341, opened 2023-08-09
launch: display success/warning message after submission

giuseppe-steduto:
Closes https://github.com/reanahub/reana-client/issues/660.

To test launching a workflow with an unexpected property declared inside the REANA specification file, this link can be used: `http://localhost:3000/launch?url=https://github.com/giuseppe-steduto/reana-demo-root6-roofit/tree/reana-yaml-allow-unexpected-props&specification=reana-cwl.yaml`

### DIFF
--- a/reana-ui/src/pages/launchOnReana/LaunchOnReana.js
+++ b/reana-ui/src/pages/launchOnReana/LaunchOnReana.js
@@ -44,10 +44,39 @@ export default function LaunchOnReana() {
     setLoading(true);
     client
       .launchWorkflow(Object.fromEntries(query))
-      .then(({ data: { workflow_id: workflowId } }) => {
-        dispatch(clearNotification);
-        history.push(`/details/${workflowId}`);
-      })
+      .then(
+        ({
+          data: { workflow_id: workflowId, message, validation_warnings },
+        }) => {
+          if (validation_warnings) {
+            let warningMessages = [];
+            // Iterate over all keys in validation_warning
+            for (const key in validation_warnings) {
+              if (key === "additional_properties") {
+                const properties = validation_warnings[key].join(", ");
+                warningMessages.push(
+                  `Unexpected properties found in the REANA specification: ${properties}.`
+                );
+              } else {
+                // For other keys, we simply display the key and its value.
+                warningMessages.push(`${key}: ${validation_warnings[key]}`);
+              }
+            }
+
+            warningMessages.forEach((warning) => {
+              message += ` ${warning}`;
+            });
+            dispatch(
+              triggerNotification("Workflow submitted with warnings", message, {
+                warning: true,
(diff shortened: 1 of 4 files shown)


reana-ui/src/pages/launchOnReana/LaunchOnReana.js
```diff
@@ -44,10 +44,37 @@ export default function LaunchOnReana() {
...
+                // For other keys, we simply display the key and its value.
+                warningMessages.push(`${key}: ${validation_warnings[key]}`);
+              }
+            }
+
+            message = `The workflow was submitted, but some warnings were issued during validation!`;
+            dispatch(triggerNotification("Workflow submitted", message));
+            warningMessages.map((warning) => {
+              dispatch(triggerNotification("Warning", warning, true));
+            });
```
mdonadoni, line 71: Regarding the notifications, I would only dispatch one notification instead of multiple ones, that includes both something like _The workflow was submitted, but some warnings were issued during validation._, plus _Unexpected properties found in the REANA specification: ..._. The second part can perhaps be on a separate line or we could even have an empty line between the two paragraphs, you can experiment a bit to see what looks good.

I would also use a different color/icon in this case, see for example the notification we have for workflow deletion:


What do you think? Do you think it's better that we use instead the "green" notification?

reana-ui/src/pages/launchOnReana/LaunchOnReana.js
```diff
@@ -44,10 +44,37 @@ export default function LaunchOnReana() {
...
+                warningMessages.push(
+                  `Unexpected properties found in the REANA specification: ${properties}`
+                );
+              } else {
+                // For other keys, we simply display the key and its value.
+                warningMessages.push(`${key}: ${validation_warnings[key]}`);
+              }
+            }
+
+            message = `The workflow was submitted, but some warnings were issued during validation!`;
```
mdonadoni, line 67: Do you think this should maybe be returned directly by the server in the `message` field?

reana-ui/src/pages/launchOnReana/LaunchOnReana.js
```diff
@@ -44,10 +44,39 @@ export default function LaunchOnReana() {
...
-      })
+      .then(
+        ({
+          data: { workflow_id: workflowId, message, validation_warnings },
+        }) => {
+          if (validation_warnings) {
+            let warningMessages = [];
+            // Iterate over all keys in validation_warning
+            for (const key in validation_warnings) {
+              if (key === "additionalProperties") {
```
mdonadoni, line 55:
```suggestion
              if (key === "additional_properties") {
```

mdonadoni (approved):

mdonadoni:
Looks good! Only small change needed

PR #350, opened 2023-09-19
launch: fetch list of demo repos from config map

### LINKED ISSUE #323: launch: make launcher homepage customizable
The demo examples in the "Welcome" page of Launch-on-REANA are currently fixed:

This is where demo workflows are defined:
https://github.com/reanahub/reana-ui/blob/ddb62edc00712e62ea8a279bb54cdd507d226511/reana-ui/src/pages/launchOnReana/Welcome.js#L20-L36

Let's investigate how to customize this page:
- Decide how we want to make the demo examples customizable, possible options:
  - only list of examples (i.e. list of names, images and URLs)
  - custom HTML page
- Decide how to make it customizable (Helm value? ConfigMap? etc.)

giuseppe-steduto:
Closes #323.

### DIFF
--- a/reana-ui/src/pages/launchOnReana/Welcome.js
+++ b/reana-ui/src/pages/launchOnReana/Welcome.js
@@ -6,37 +6,46 @@
   under the terms of the MIT License; see LICENSE file for more details.
 */
 
-import { Button, Container, Grid, Image } from "semantic-ui-react";
+import { Button, Card, Container, Icon } from "semantic-ui-react";
 import { Link } from "react-router-dom";
 import { useSelector } from "react-redux";
 
 import { getConfig } from "~/selectors";
+
 import exampleRootImg from "~/images/example-root.png";
 import exampleAtlasRecastImg from "~/images/example-atlas-recast.png";
 import exampleCmsHiggsTauTauImg from "~/images/example-cms-htt.png";
 
 import styles from "./Welcome.module.scss";
 
-const DEMOS = [
+const DEFAULT_DEMO_REPOSITORIES = [
   {
     name: "RooFit",
     url: "https://github.com/reanahub/reana-demo-root6-roofit",
-    image: exampleRootImg,
+    image_url: exampleRootImg,
+    specification: "reana.yaml",
   },
   {
     name: "ATLAS RECAST",
     url: "https://github.com/reanahub/reana-demo-atlas-recast/",
-    image: exampleAtlasRecastImg,
+    image_url: exampleAtlasRecastImg,
+    specification: "reana.yaml",
   },
   {
     name: "CMS HiggsTauTau",
     url: "https://github.com/cms-opendata-analyses/HiggsTauTauNanoAODOutreachAnalysis",
-    image: exampleCmsHiggsTauTauImg,
+    image_url: exampleCmsHiggsTauTauImg,
+    specification: "reana.yaml",
   },
 ];
 
 export default function Welcome() {
-  const { docsURL } = useSelector(getConfig);
+  const { docsURL, launcherExamples: repositoriesFromConfig } =
+    useSelector(getConfig);
+
+  // If the REANA admin has not specified a list of demo repositories, use the
+  // default ones.
+  const launcherExamples = repositoriesFromConfig ?? DEFAULT_DEMO_REPOSITORIES;
 
   return (
     <Container text className={styles.container}>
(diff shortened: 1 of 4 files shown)


reana-ui/src/pages/launchOnReana/Welcome.js
```diff
@@ -11,32 +11,39 @@ import { Link } from "react-router-dom";
...
 import { getConfig } from "~/selectors";
+
 import exampleRootImg from "~/images/example-root.png";
 import exampleAtlasRecastImg from "~/images/example-atlas-recast.png";
 import exampleCmsHiggsTauTauImg from "~/images/example-cms-htt.png";
 
 import styles from "./Welcome.module.scss";
 
-const DEMOS = [
+const DEFAULT_DEMO_REPOSITORIES = [
```
giuseppe-steduto, line 21: Should we leave these here? In https://github.com/reanahub/reana/pull/746, I added the same repositories in the `values.yaml` file, mainly because it serves an example to admins that want to change them, whereas here the main purpose is to have a default option in case the `ui_launcher_repositories` value is removed. Do you think it is okay or does it feel like doing the same thing twice?

reana-ui/src/pages/launchOnReana/Welcome.module.scss
```diff
@@ -13,6 +13,8 @@
 }
 
 .demo-card {
+  margin-bottom: 2em;
```
mdonadoni, line 16: Now that custom images are possible, we probably need some more CSS tweaks. See as an example the following screenshot:



In the second row you can see that the images touch the button below them, plus the third image is also stretched.

As a second example, this is what happens if images are not resized (note that even our default examples change, as the ATLAS RECAST image is not square):



This is instead what happens when images are cropped (note that the ATLAS RECAST plot is cut):



Possible inspirations/help:
- https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit
- https://react.semantic-ui.com/views/card/
- https://codepen.io/MartijnCuppens/pen/NWNdOBR

mdonadoni:
PS: could you also update CHANGES.rst

mdonadoni:
PS: could you also update CHANGES.rst?

reana-ui/src/pages/launchOnReana/Welcome.module.scss
giuseppe-steduto: I would like the image not to be cropped (especially useful when there are graphs shown, as in our case), and at the same time, I would like the images/cards to be all the same height (so that the page looks tidier). I changed a little bit the layout to take advantage of the `Card` component, and took inspiration from the third link you sent to have the background filled by the blurred image.

reana-ui/src/pages/launchOnReana/Welcome.module.scss
```diff
@@ -12,15 +12,47 @@
...
+    position: relative;
+    overflow: hidden;
+  }
+
+  // Add a pseudo element to keep the aspect ratio
+  .img-container::after {
+    content: "";
+    display: block;
+    padding-bottom: calc(100% / 1); // Define ratio here
+  }
```
mdonadoni, line 37: I think we can avoid `.img-container::after` by using `aspect-ratio: 1` on `img-container`: https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio

reana-ui/src/pages/launchOnReana/Welcome.module.scss
```diff
@@ -12,15 +12,47 @@
...
+  img {
+    width: 100%;
+    height: 100%;
+    position: absolute;
+    object-fit: contain;
+  }
+
+  // Blur the background image and cover the whole container
+  // Use the negative blur value as offset to prevent the blur
+  // from mixing with the document background color
```
mdonadoni, line 48: What is the negative blur value?

reana-ui/src/pages/launchOnReana/Welcome.js
mdonadoni, line 1: Given that now there can be lots of examples on this page, let's move the last phrase (_If your analysis is hosted on an external site such..._) to the top, removing the repetition of _Gitlab, Github or Zenodo_.

Something like this:
_This page allows you to launch analysis workflows hosted on external sources such as GitHub, GitLab, Zenodo._

_If your analysis is hosted on an external site, please provide a url parameter that would point to where your analysis is hosted. See the [launcher docs](https://docs.reana.io/running-workflows/launching-workflows/) for more details._

_You can start by launching one of the following demo examples:_

reana-ui/src/pages/launchOnReana/Welcome.module.scss
```diff
@@ -12,15 +12,47 @@
...
+  padding-bottom: 1em;
+}
+
 .demo-card {
-  .img-wrapper > a {
-    height: 200px;
-    display: flex;
-    justify-content: center;
-  }
+  margin-bottom: 2em;
```
mdonadoni, line 20: I think `margin-bottom` is not needed anymore

reana-ui/src/reducers.js
```diff
@@ -147,6 +148,7 @@ const config = (state = configInitialState, action) => {
         userConfirmation: action.user_confirmation,
         quotaEnabled: action.quota_enabled,
         filePreviewSizeLimit: action.file_preview_size_limit,
+        launcherExamples: action.launcher_repositories,
```
mdonadoni, line 151:
```suggestion
        launcherExamples: action.launcher_examples,
```

reana-ui/src/pages/launchOnReana/Welcome.js
mdonadoni: Let's keep these also here

reana-ui/src/pages/launchOnReana/Welcome.module.scss
giuseppe-steduto: I had used it before as a way to prevent the blur from overflowing, but `overflow: hidden` did the trick better, so I've just deleted it.

mdonadoni (approved):

mdonadoni:
Looking good! I will rebase and merge

Issue #349, opened 2023-09-19
wf-list: let users delete all runs when deleting workflows 


giuseppe-steduto:
In reana-client, when deleting a workflow with `reana-client delete`, there is the option to specify the `--include-all-runs` flag, to delete all the runs of the workflow.

When deleting a workflow from the web UI, however, there is no such option, and only the selected run is deleted. We should consider adding a checkbox or something similar to allow the users to delete all the runs of the workflow they are trying to delete, both to stay consistent with the CLI client and because it would be an useful feature: I use it quite a lot, especially with test workflows.

This choice could be proposed in the modal for confirming the deletion of the workflow:
![image](https://github.com/reanahub/reana-ui/assets/34663336/f811e972-990f-45fa-b7bf-45e257a71b9a)
(And then update the "are you sure" warning message if the box is checked).

PR #357, opened 2023-10-12
ui: add option to delete all workflow runs

mdonadoni:
- package: upgrade jsroot to 7.5.0
- client: make parameters optional in getWorkflows
- ui: add option to delete all workflow runs

Closes https://github.com/reanahub/reana-ui/issues/349

### DIFF
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -411,11 +411,11 @@ export function fetchWorkflowRetentionRules(id) {
   };
 }
 
-export function deleteWorkflow(id, workspace = true) {
+export function deleteWorkflow(id, { workspace = true, allRuns = false } = {}) {
   return async (dispatch) => {
     dispatch({ type: WORKFLOW_DELETE_INIT });
     return await client
-      .deleteWorkflow(id, { workspace })
+      .deleteWorkflow(id, { workspace, allRuns })
       .then((resp) => {
         dispatch({ type: WORKFLOW_DELETED, ...resp.data });
         dispatch({ type: WORKFLOW_LIST_REFRESH });
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -261,14 +261,20 @@ export function fetchWorkflows({
   status,
   sort,
   showLoader = true,
-  workflowId,
+  workflowIdOrName,
 }) {
   return async (dispatch) => {
     if (showLoader) {
       dispatch({ type: WORKFLOWS_FETCH });
     }
     return await client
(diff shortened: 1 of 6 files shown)


reana-ui/yarn.lock
giuseppe-steduto, line 1: Why not updating `package.json` as well to pin the 7.5.0 version?

reana-ui/src/actions.js
```diff
@@ -407,11 +411,11 @@ export function fetchWorkflowRetentionRules(id) {
...
 
-export function deleteWorkflow(id, workspace = true) {
+export function deleteWorkflow(id, { workspace = true, allRuns = false } = {}) {
   return async (dispatch) => {
     dispatch({ type: WORKFLOW_DELETE_INIT });
     return await client
-      .deleteWorkflow(id, { workspace })
+      .deleteWorkflow(id, { workspace, allRuns })
       .then((resp) => {
         dispatch({ type: WORKFLOW_DELETED, ...resp.data });
```
giuseppe-steduto, line 420: Very minor: maybe the text of the success notification could be different when deleting _all the runs_ of a workflow (like in `r-client`)? Could be something as simple as changing "workflow" to "workflows" in "workflow successfully deleted", or could be a totally different sentence, like `r-client` does. 

Note that if we want to address this, rather than changing `r-ui`, it might be better to change the `message` property returned by `r-w-c` when deleting more than one run.

giuseppe-steduto (commented):
Works well, I like it! Just left a couple of minor comments.

And I'm gonna leave an idea here: what do you think about displaying the checkbox only when the number of total runs is greater than 1?

giuseppe-steduto:
Also, this could be a nice occasion to update the outdated UI screenshot used in the docs in the [delete workflows](https://docs.reana.io/advanced-usage/user-quotas/#delete-workflows) section.

mdonadoni:
> And I'm gonna leave an idea here: what do you think about displaying the checkbox only when the number of total runs is greater than 1?

Given that this depends on a network call I would avoid this, as otherwise the checkbox would appear after some delay, causing the layout to change and the buttons at the bottom of the modal to move around.

reana-ui/src/actions.js
mdonadoni: Created issue: https://github.com/reanahub/reana-workflow-controller/issues/531

mdonadoni:
> Also, this could be a nice occasion to update the outdated UI screenshot used in the docs in the [delete workflows](https://docs.reana.io/advanced-usage/user-quotas/#delete-workflows) section.

Done! https://github.com/reanahub/docs.reana.io/pull/185

giuseppe-steduto (approved):

PR #358, opened 2023-10-12
global: upgrade to node 18

### LINKED ISSUE #347: package: upgrade to Node.js v18/v20
We are currently relying on Node.js v16 to build reana-ui, which is very close to being EOL (2023-09-11): https://nodejs.dev/en/about/releases/

We should upgrade to Node.js v18/v20 by updating the  Dockerfile, CI and README/docs accordingly

mdonadoni:
Use node 18 instead of node 16, as the former is the default version on
Fedora/Ubuntu and the latter has reached EOL.

Closes #347

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v3
         with:
-          node-version: "16"
+          node-version: "18"
 
       - name: Install project dependencies
         run: yarn global add prettier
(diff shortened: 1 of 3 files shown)


giuseppe-steduto (approved):
LGTM! For a similar change, in version 0.8.1 we added [a line to the changelog](https://github.com/reanahub/reana-ui/blob/master/CHANGES.rst?plain=1#L58). Should we do it for this upgrade as well?

mdonadoni:
> Should we do it for this upgrade as well?

Done! (and merged)

PR #361, opened 2023-10-25
add "launch on REANA" badge and launcher URL creator

### LINKED ISSUE #324: launch: badge/URL generator
Currently, REANA does not provide a way to generate Launch-on-REANA badges or URLs, which makes it difficult to use the Launch-on-REANA functionality, as a user either needs to build the URL manually or with a script.

We should provide an easy way for users to generate Launch-on-REANA URLs/badges, see for example [binder](https://mybinder.org/).

giuseppe-steduto:
- **launch: display parameters in table**
  Change the launch page so that the optional parameters are displayed
in a table, where each row correspond to a parameter, and the value of
each of them is displayed in its JSON-serialized version.
This allows to correctly show even complex parameters like objects and
lists.

- **launch: add launcher URL and badge creator page**

 - **client: fix no-anonymous-default-export linting warning**

Closes #324.

### DIFF
--- a/reana-ui/src/pages/badgeCreator/LauncherBadgeCreator.js
+++ b/reana-ui/src/pages/badgeCreator/LauncherBadgeCreator.js
@@ -1,0 +1,319 @@
+/*
+  This file is part of REANA.
+  Copyright (C) 2023 CERN.
+
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+import { useState } from "react";
+import {
+  Button,
+  Container,
+  Divider,
+  Form,
+  Icon,
+  Popup,
+  Segment,
+} from "semantic-ui-react";
+
+import styles from "./LauncherBadgeCreator.module.scss";
+import { CodeSnippet, Title } from "~/components";
+import { api, LAUNCH_ON_REANA_BADGE_URL } from "~/config";
+import BasePage from "~/pages/BasePage";
+import { stringifyQueryParams } from "~/util";
+import { Link } from "react-router-dom";
+
+const LauncherBadgeCreator = () => {
+  const [launcherData, setLauncherData] = useState({
+    url: "",
+    analysisName: "",
+    specFileName: "",
+    parameters: [],
+  });
+
+  const [validationErrors, setValidationErrors] = useState({
+    url: null,
+    analysisName: null,
+    specFileName: null,
+    parameters: [],
+  });
+
+  const [launcherURL, setLauncherURL] = useState();
+
+  const handleInputChange = (e, { name, value }) => {
+    // Clear the validation errors relative to the field that has been changed
+    setValidationErrors((prevErrors) => ({ ...prevErrors, [name]: null }));
+    setLauncherData({ ...launcherData, [name]: value });
+  };
+
+  const handleAddParam = () => {
+    setLauncherData((prevData) => ({
+      ...prevData,
+      parameters: [...prevData.parameters, { key: "", value: "" }],
+    }));
+    setValidationErrors((prevErrors) => ({
+      ...prevErrors,
+      parameters: [...prevErrors.parameters, { key: null, value: null }],
+    }));
+  };
+
+  const handleDeleteParam = (index, paramKey) => {
+    // Clear the validation error relative to the parameter that has been deleted
+    setValidationErrors((prevErrors) => ({
+      ...prevErrors,
+      parameters: prevErrors.parameters.toSpliced(index, 1),
+    }));
+    setLauncherData((prevData) => {
+      const newParameters = [...prevData.parameters];
+      newParameters.splice(index, 1);
+      return { ...prevData, parameters: newParameters };
+    });
+  };
+
+  const handleParamChange = (index, key, value, paramKey) => {
+    // Clear the validation errors relative to the parameter that has been changed
+    setValidationErrors((prevErrors) => ({
+      ...prevErrors,
+      parameters: prevErrors.parameters.toSpliced(index, 1, {
+        key: null,
+        value: null,
+      }),
+    }));
+    setLauncherData((prevData) => {
+      const newParameters = [...prevData.parameters];
+      newParameters[index] = { ...newParameters[index], [key]: value };
+      return { ...prevData, parameters: newParameters };
+    });
+  };
+
+  const handleSubmit = () => {
+    let paramsString = null;
+    let invalid = false;
+
+    // Clear the parameter validation errors, to start with a clean slate
+    // and avoid showing e.g. duplicate key errors if the first key has been changed
+    setValidationErrors((prevErrors) => ({
+      ...prevErrors,
+      parameters: prevErrors.parameters.map(() => ({
+        key: null,
+        value: null,
+      })),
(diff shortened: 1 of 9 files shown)


reana-ui/src/pages/launchOnReana/LaunchOnReana.js
```diff
@@ -172,67 +165,73 @@ export default function LaunchOnReana() {
...
+                              >
+                                <CodeSnippet>
+                                  {JSON.stringify(value, null, 2)}
+                                </CodeSnippet>
+                              </Table.Cell>
+                            </Table.Row>
+                          ),
+                        )}
+                      </Table.Body>
+                    </Table>
```
mdonadoni, line 222: This table feels a bit too big, what do you think? Should we remove some of the padding to the cells, and maybe even reduce the font size?

reana-ui/src/pages/launchOnReana/LaunchOnReana.module.scss
```diff
@@ -58,4 +69,23 @@
...
+
+  .loader {
+    padding: 20px;
+    background-color: white;
+    border-radius: 5px;
+
+    &:before,
+    &:after {
+      top: 10px !important;
+    }
```
mdonadoni, line 89: Let's keep the default loader distance? Or is there a particular reason to use a custom distance only for the launcher?

reana-ui/src/pages/badgeCreator/BadgeLaunchCreator.js
```diff
@@ -0,0 +1,268 @@
...
+        <p>
+          Fill in the form below to generate the URL to the REANA launcher that
+          will run your analysis, as well as the markdown code for a "Launch on
+          REANA" badge that you can include in your GitHub repositories!
+        </p>
+        <Form onSubmit={handleSubmit}>
+          <Segment className={styles.segment}>
+            <Grid>
+              <Grid.Row>
+                <Grid.Column>
```
mdonadoni, line 143: I think there is no need to use grid in forms, as we can just use a few `Form.Group`? https://react.semantic-ui.com/collections/form/#group-variations-evenly-divided-group

For example:
```jsx
<Form.group widht="equals">
  ...
```

(note that you can also specify the width of a field in "grid columns": https://react.semantic-ui.com/collections/form/#field-variations-width-field )

reana-ui/src/pages/badgeCreator/BadgeLaunchCreator.js
```diff
@@ -0,0 +1,268 @@
...
+                  </Button>
+                </Grid.Column>
+              </Grid.Row>
+              {launcherURL && (
+                <Grid.Row>
+                  <Grid.Column>
+                    <Divider></Divider>
+                    <p>
+                      Here you can find the URL you can use to launch the
+                      workflow on your REANA instance:
```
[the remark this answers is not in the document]
mdonadoni: (or at least, I would change to _on this REANA instance_, as the user does not "own" the REANA instance, so it's not his/her instance)

mdonadoni (commented):
I like this a lot! 🤩

reana-ui/src/pages/launchOnReana/LaunchOnReana.js
```diff
@@ -172,67 +165,73 @@ export default function LaunchOnReana() {
...
-                  href={query.get("url")}
-                  target="_blank"
-                  rel="noopener noreferrer"
-                >
-                  {query.get("url")}
-                </a>
                 <div>
-                  <Icon name="file code outline" />
+                  <div className={styles["launch-section-title"]}>
+                    <Icon name="file code outline" />
```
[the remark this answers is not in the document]
giuseppe-steduto: I removed the icons, both before _Specification_ and before _Parameters_.

reana-ui/src/pages/badgeCreator/LauncherBadgeCreator.js
```diff
@@ -0,0 +1,297 @@
...
+import { useDispatch } from "react-redux";
+import { stringifyQueryParams } from "~/util";
+import { Link } from "react-router-dom";
+
+const LauncherBadgeCreator = () => {
+  const dispatch = useDispatch();
+  const [launcherData, setLauncherData] = useState({
+    url: "",
+    analysisName: "",
+    specFileName: "reana.yaml",
```
mdonadoni, line 34:
```suggestion
    specFileName: "",
```

Given that at the beginning the form is empty?

reana-ui/src/pages/badgeCreator/LauncherBadgeCreator.js
```diff
@@ -0,0 +1,297 @@
...
+      launcherData.parameters.forEach((param, index) => {
+        try {
+          paramsObject[param.key] = JSON.parse(param.value);
+        } catch (e) {
+          errorMessage = "The form contains invalid data!";
+          setValidationErrors((prevErrors) => ({
+            ...prevErrors,
+            parameters: {
+              ...prevErrors.parameters,
+              [param.key]: "Not a valid JSON string!",
```
mdonadoni, line 144: We are missing a couple of checks for parameters:
1. keys must not be empty
2. keys must be unique

Also note that, if two keys are the same (or if both are empty), then the errors will be shown for both input fields, even if one of them has no validation issues. Maybe we could access/store errors by index, instead of by key, to avoid this?

reana-ui/src/pages/badgeCreator/LauncherBadgeCreator.js
```diff
@@ -0,0 +1,297 @@
...
+                  width={10}
+                  placeholder="Value"
+                  name="value"
+                  value={param.value}
+                  onChange={(e) =>
+                    handleParamChange(index, "value", e.target.value, param.key)
+                  }
+                  error={
+                    validationErrors.parameters[
+                      launcherData.parameters[index].key
```
mdonadoni, line 252:
```suggestion
                      param.key
```

reana-ui/src/pages/badgeCreator/LauncherBadgeCreator.module.scss
```diff
@@ -0,0 +1,23 @@
...
+
+.container {
+  margin-top: 5em;
+}
+
+.segment {
+  margin-bottom: 1em !important;
+}
+
+.abc {
```
mdonadoni, line 17: Maybe a better name? :smile: Something like `parameters-header`

reana-ui/src/pages/badgeCreator/LauncherBadgeCreator.js
giuseppe-steduto: Done! Now the error could also be displayed on the key form, in case it's duplicate or not unique. I think all the cases should be handled now, there is just one exception:
1. Insert 2 equal keys
2. Try to create the badge
3. Get the "duplicate" error on the _second_ key
4. Delete / change the _first_ key
Now the second key still displays the "duplicate key" error, because its validation state has not been cleaned. Of course, after changing that key or trying to create the badge, the error will disappear.

I don't think that's a big deal, and adding the code to gracefully handle this edge case is not really worth it. What do you think?

mdonadoni (approved):
LGTM! The PR just needs to be rebased

PR #365, opened 2023-11-20
client: avoid using verbose flag to fetch the workflow list

giuseppe-steduto:
Closes reanahub/reana-workflow-controller#486.


mdonadoni:
No need to change API after performance improvements of https://github.com/reanahub/reana-db/pull/213

PR #372, opened 2023-12-11
release: 0.9.3

mdonadoni:
- docker: update to nginx 1.25
- docker: add metadata labels
- release: 0.9.3

TODO
- [ ] Change dates in CHANGES.rst and in Dockerfile


reana-ui/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "reana-ui",
-  "version": "0.9.2",
+  "version": "0.9.3",
```
mdonadoni, line 3: Should this instead be 0.9.2.1?

CHANGES.rst
```diff
@@ -1,6 +1,12 @@
 Changes
 =======
 
+Version 0.9.3 (2023-12-12)
+--------------------------
+
+- Adds metadata labels to Dockerfile.
+- Changes version of nginx Docker image from 1.19 to 1.25.
```
tiborsimko, line 8: Nginx with capital N?

Dockerfile
```diff
@@ -27,8 +27,19 @@ RUN apt-get update -y && \
...
-FROM docker.io/library/nginx:1.19
+FROM docker.io/library/nginx:1.25
 COPY --from=react-build /code/reana-ui/build /usr/share/nginx/html
 COPY nginx/reana-ui.conf /etc/nginx/conf.d/default.conf
 EXPOSE 80
 CMD ["nginx", "-g", "daemon off;"]
+
+LABEL org.opencontainers.image.authors="team@reanahub.io"
+LABEL org.opencontainers.image.created="2023-12-12"
+LABEL org.opencontainers.image.description="REANA UI"
```
tiborsimko, line 38: What about:

- "REANA reproducible analysis platform - web user interface component"

We could use the same description naming technique elsewhere, e.g. ...job controller component, ...Snakemake workflow engine component, etc.

reana-ui/package.json
tiborsimko: We have been numbering components independently in the past, and some bring new features, as it were, so I think we can go with independent versioning of components.

tiborsimko (approved):

CHANGES.rst
mdonadoni: I will change it to NGINX (all uppercase), see for example the main page: https://www.nginx.com/

Dockerfile
```diff
@@ -27,8 +27,19 @@ RUN apt-get update -y && \
...
+LABEL org.opencontainers.image.documentation="https://reana-ui.readthedocs.io/"
```
tiborsimko, line 39: Since each paragraph in `Dockerfile` is accompanied by a leading comment, you can precede the label setting section by:

```
# Set image labels
```

PR #387, opened 2024-01-18
fix(utils): update failed workflows duration using finished time

### LINKED ISSUE #386: ui: workflow duration keeps changing for failed workflows
When looking at the details of a failed workflow, either in the workflow list page or in the workflow details page, the workflow duration keeps updating as time goes by.

For example, in a workflow like this:
image

After refreshing the page (either manually or automatically by the REANA UI), the new duration will be different, e.g. `failed after 6h 12m 59s`. This should not happen, as that value indicates the duration of the workflow, and that's not supposed to change after the workflow terminates its execution (regardless of the status).

The issue seems to present itself only with failed workflows.

giuseppe-steduto:
- fix(utils): update failed workflows duration using finished time
  Fix the way in which the duration of failed workflows is calculated, to
  use the time at which the run was finished rather than the current time.

  Closes #386

- ci(shellcheck): exclude node_modules from the analyzed paths

### DIFF
--- a/run-tests.sh
+++ b/run-tests.sh
@@ -29,7 +29,7 @@ check_commitlint () {
 }
 
 check_shellcheck () {
-    find . -name "*.sh" -exec shellcheck {} \+
+    find . -name "*.sh" ! -path "./reana-ui/node_modules/*" -exec shellcheck {} \+
 }
 
 check_sphinx () {
--- a/reana-ui/src/util.js
+++ b/reana-ui/src/util.js
@@ -156,13 +156,16 @@ export function getDuration(start, end) {
 /**
  * Parses workflows date info in a friendly way.
  */
-function parseWorkflowDates(workflow) {
+export function parseWorkflowDates(workflow) {
   const createdMoment = moment.utc(workflow.created);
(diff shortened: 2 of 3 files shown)


mdonadoni (commented):
Looks good!

What about the `fix(utils)` scope? @tiborsimko

mdonadoni (approved):
LGTM, but it needs a rebase!

run-tests.sh
```diff
@@ -29,7 +29,7 @@ check_commitlint () {
 }
 
 check_shellcheck () {
-    find . -name "*.sh" -exec shellcheck {} \+
+    find . -name "*.sh" ! -path "./reana-ui/node_modules/*" -exec shellcheck {} \;
```
mdonadoni, line 32:
```suggestion
    find . -name "*.sh" ! -path "./reana-ui/node_modules/*" -exec shellcheck {} \+
```

mdonadoni: See https://github.com/reanahub/reana-ui/commit/7b5f29ebc604a2d27d76f8a51b437c3e561fec32

PR #394, opened 2024-02-12
unify badges style and add circular progress bar to workflow list and details pages

### LINKED ISSUE #151: ui: display workflow size in details page

### LINKED ISSUE #395: ui: make badges and progress indicator more consistent across workflow list and details pages
This discussion originated in https://github.com/reanahub/reana-ui/pull/321. Questions were arised about the placement of the workflow badges in the workflow UI, nominally:

- The GitHub / GitLab / Zenodo icon and link
- The Jupyter notebook icon and link
- The workflow size icon and value
- Possible other badges that we could add in the future (e.g. shared workflow)

At the moment the styles of these badges are inconsistent, both between each other and across different pages (workflow list, workflow details). We should redesign the components to make them more consistent (better UI), more scalable (easy to add other badges), more informative, and more usable (better UX).

Moreover, in the workflow list page the workflow box, which is an interactive element (a clickable button that links to the workflow details page) contains other interactive element (the Jupyter notebook icon and the action menu button) that link to other pages or perform other actions. The [HTML `a` element specification](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) forbids having nested `a` element / interactive content, so we should consider and fix this as well.

Attention should be put in extracting common reusable React components to be used both in the workflow list and in the workflow details page to avoid code repetition and facilitate maintainability.

Finally, in the context of this redesign and with the goal of making the workflow list and the workflow details page more consistent with each other, we could replace the linear progress bar in the workflow details page with a similar progress indicator, that would be used in both pages, and that would be more complete, by displaying the status of different steps (how many are finished, running, still to be executed, ...), e.g. on the lines of what the GitHub actions progress bar does.

giuseppe-steduto:
- feat(workflow-list): unify badges style and remove nested links (#394)
- feat(progress): add circular progress bar in workflow list page (#394)
- refactor: centralize workflow info/badges in list, details pages (#394)

Closes #151
Closes #395


reana-ui/src/components/WorkflowBadges.js
```diff
@@ -0,0 +1,66 @@
...
+    <>
+      <div className={styles.badgesContainer}>
+        {workflow.duration && (
+          <Label
+            size="tiny"
+            content={workflow.duration}
+            icon="clock"
+            as="a"
+            href={"/details/" + id}
+            target="_blank"
```
tiborsimko, line 32: A functional observation. Clicking on a badge on the workflow list page (and on workflow detailed page) leads to opening new tabs. This can quickly fill up, e.g. try to click several times on the duration badge on the worklow detailed page, it'll generate a new tab for every repeated click.

It would be better to always open in the same tab, i.e. remove `_blank` targets here and elsewhere in the code. If a user would like to open a new tab, they could always middle-click.

reana-ui/src/components/WorkflowBadges.js
```diff
@@ -0,0 +1,66 @@
...
+            rel="noopener noreferrer"
+            color={statusMapping[workflow.status].color}
+            onClick={(e) => e.stopPropagation()}
+          />
+        )}
+        {hasDiskUsage && (
+          <Label
+            size="tiny"
+            content={size.human_readable}
+            icon="hdd"
```
tiborsimko, line 42: Another functional observation. Clicking on the workspace size brings people to the default detailed record page. It would be nicer if the user landed on the workspace tab on the detailed page, because people clicking on the workspace size would probably like to consult workspace content directly.

reana-ui/src/components/WorkflowProgressCircleBar.js
```diff
@@ -0,0 +1,103 @@
+/*
+  -*- coding: utf-8 -*-
+
+  This file is part of REANA.
+  Copyright (C) 2024 CERN.
+
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
```
tiborsimko, line 10: A functional observation. When I submit a BSM workflow, immediately after submission the progress is shown as:

red

Here the "red" seems a bit harsh, so perhaps the colour is not appropriate.

Later when the engine starts and we know how many jobs are to be run, then the progress looks better:

red2

Perhaps we should show "grey" also in the first case?

reana-ui/src/pages/workflowDetails/WorkflowDetails.js
```diff
@@ -2,7 +2,7 @@
   -*- coding: utf-8 -*-
 
   This file is part of REANA.
-  Copyright (C) 2020, 2022, 2023 CERN.
+  Copyright (C) 2020, 2022, 2023, 2024 CERN.
```
tiborsimko, line 5: A functional observation. Whilst the workflow is running, on the detailed record page, a step may be "completed", but the UI still says "created" next to the step selection drop-down:

created

reana-ui/src/components/WorkflowProgressCircleBar.js
```diff
@@ -0,0 +1,103 @@
...
+import PropTypes from "prop-types";
+import styles from "./WorkflowProgressCircleBar.module.scss";
```
tiborsimko, line 12: Another cosmetic observation. Whilst the workflow is running, we are displaying (correctly) green progress bar part for those jobs that have finished. However, on the worklfow list page, this gives too much "green" for a workflow that is "blue" because it is still running.

Perhaps we may want to show the progress circle in blue if the status of the workflow is running? In this way it'll be easy to distinguish directly which workflows are finished and which are still running, and the colour consistency may look better in that way.

blue

reana-ui/src/components/WorkflowProgressCircleBar.js
```diff
@@ -0,0 +1,103 @@
...
+
+export default function WorkflowProgressCircleBar({ workflow }) {
+  const { completed, failed, running, total, status } = workflow;
+
+  const size = 80;
+  const strokeWidth = 10;
+  const radius = size / 2 - strokeWidth;
```
mdonadoni, line 19:
```suggestion
  const radius = size / 2 - strokeWidth / 2;
```
strokeWidth probably also needs to be divided by 2, as the radius is calculated fro the center to the "middle" of the stroke (in other words the inner circle has radius r - strokeWidth/2, and the outer radius is r + strokeWidth/2). Otherwise the circle will be smaller than the size of the svg:
Schermata 2024-02-28 alle 12 41 13

reana-ui/src/components/WorkflowProgressCircleBar.module.scss
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  .progress-bar-failed {
+    stroke: $sui-red;
+  }
+
+  .progress-bar-background,
+  .progress-bar-workflow-status-queued,
+  .progress-bar-workflow-status-pending,
+  .progress-bar-workflow-status-created {
```
mdonadoni, line 30: I'm not exactly sure what is the purpose of the `.progress-bar-workflow-status-...` classes?

reana-ui/src/components/WorkflowProgressCircleBar.js
```diff
@@ -0,0 +1,103 @@
...
+        />
+        <circle
+          cx={`${size / 2}`}
+          cy={`${size / 2}`}
+          r={`${radius}`}
+          className={`${styles["progress-bar-failed"]} ${
+            styles["progress-bar-workflow-status-" + status]
+          }`}
+          strokeDasharray={`${lengthFailedArc} ${circumference}`}
+          strokeDashoffset={`-${lengthFinishedArc}`}
```
mdonadoni, line 92: Why is the offset for failed and running the same? Wouldn't their arcs overlap?

reana-ui/src/pages/workflowList/WorkflowList.js
```diff
@@ -132,7 +132,7 @@ function Workflows() {
   );
 
   return (
-    <div className={styles.container}>
+    <div className={`${styles.container} ${styles["workflow-list-container"]}`}>
```
mdonadoni, line 135: I think the style `workflow-list-container` is never defined?

mdonadoni (commented):
Left some comments, but I like the redesign very much! Neat idea using an svg to show the progress circle :rocket:

reana-ui/src/components/WorkflowBadges.js
```diff
@@ -0,0 +1,66 @@
...
+  return (
+    <>
+      <div className={styles.badgesContainer}>
+        {workflow.duration && (
+          <Label
+            size="tiny"
+            content={workflow.duration}
+            icon="clock"
+            as="a"
+            href={"/details/" + id}
```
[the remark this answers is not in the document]
giuseppe-steduto: I thought about that, ended up making it clickable so that all the badges are clickable. But given this is different from the other ones, disable the link might be the best choice indeed!

reana-ui/src/components/WorkflowBadges.js
```diff
@@ -0,0 +1,66 @@
...
+            onClick={(e) => e.stopPropagation()}
+          />
+        )}
+        {hasDiskUsage && (
+          <Label
+            size="tiny"
+            content={size.human_readable}
+            icon="hdd"
+            as="a"
+            href={"/details/" + id}
```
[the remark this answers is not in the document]
giuseppe-steduto: See my comment above, I think it helps to make all these badges clickable so that users know that all these element (that look the same) behave in the same way. It would be nice to link it to the workspace tab maybe in the future!

reana-ui/src/components/WorkflowBadges.module.scss
```diff
@@ -0,0 +1,27 @@
...
+  Copyright (C) 2023 CERN.
+
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+.badgesContainer {
+  display: flex;
+
+  a.label {
```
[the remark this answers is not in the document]
giuseppe-steduto: Yes, it's supposed to look like (1). I will change the selector and remove the CSS that makes the background white!

reana-ui/src/components/WorkflowProgressCircleBar.js
giuseppe-steduto: Yes! They both start after the "finished" arc - if a workflow has some steps still running there shouldn't be failed ones (and viceversa), since as soon as a step fails, the whole workflows fails, so there's always either the running or the failed arc. Is that right?

reana-ui/src/pages/workflowList/WorkflowList.js
giuseppe-steduto: Oops, my bad, I think it's a relic from an attempt to style the page differently. Removed it!

mdonadoni:
Rebased the PR and made some changes addressing comments posted here and coming from internal discussions in a separate commit. How should we handle commits, should we just squash and merge?

reana-ui/src/components/WorkflowBadges.js
mdonadoni: The links on the disk/duration badges were removed. If we want to keep them and to make them point to specific tabs of the workflow detail page, maybe we should open an issue and do so in another PR?

reana-ui/src/components/WorkflowBadges.js
mdonadoni: See previous comment

reana-ui/src/pages/workflowDetails/WorkflowDetails.js
mdonadoni: Probably unrelated to this PR, let's check and fix this as part of another issue/PR?

reana-ui/src/components/WorkflowProgressCircleBar.js
mdonadoni: Progress bar is now blue, with solid stroke for finished workflows and dashed stroke for running ones.

reana-ui/src/components/WorkflowBadges.js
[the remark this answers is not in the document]
mdonadoni: Links removed, see previous discussion https://github.com/reanahub/reana-ui/pull/394#discussion_r1505767789

reana-ui/src/components/WorkflowProgressCircleBar.module.scss
mdonadoni: Makes sense, thanks!

reana-ui/src/components/WorkflowProgressCircleBar.js
mdonadoni: It should not matter, but all the arcs are now consecutive, without any overlaps

tiborsimko (approved):

PR #397, opened 2024-02-21
fix(workflow-details): make step labels less prominent (#397)

### LINKED ISSUE #352: wf-details: make label for step status more visually balanced
In the job logs section of the workflow-details page, the label showing the status of the single selected step is very visually prominent, and draws the attention even more than the status of the workflow:
image

We should explore some alternatives to make the page more visually balanced, while still keeping the relevant information there. A couple of preliminary ideas:
- Reformat the label so that the status icon (the circle and the dot next to the name of the steps that are displayed when opening the dropdown) is shown even in the selected step, together with the time. The result could be something similar to this:
image

- Change the style of the step status label to make it less "heavy" on the eyes

giuseppe-steduto:
Closes #352 

_Note: dependant on #394_


mdonadoni:
PS: I have rebased the PR on top of master

tiborsimko:
> What do you think @tiborsimko?

Agreed about (a) and (c).

WRT (b) the idea behind old discussions to explore moving the timing information to the selection box was that the "finished" would _also_ moved there, so to speak, and not stay outside. Basically, the selection box would show the step name prominently and after it, in parentheses, whether it is running or finished, and its duration.  Hence the middle screenshot does not fully cover these ideas, since part of the information is left outside the selection box. At this stage of things, I agree that the third screenshot is preferable to the second.

PR #403, opened 2024-03-19
feat(gitlab): add filtering and pagination (#403)

mdonadoni:
- feat(gitlab): add filtering for GitLab project lists (#403)
- feat(gitlab): add pagination for GitLab project lists (#403)
    
    BREAKING CHANGE: The REST API endpoint `gitlab_projects` now includes
    pagination details.

Closes reanahub/reana-server#518

### DIFF
--- a/reana-ui/src/client.js
+++ b/reana-ui/src/client.js
@@ -25,7 +25,8 @@ export const USER_REQUEST_TOKEN_URL = `${api}/api/token`;
 export const USER_CONFIRM_EMAIL_URL = `${api}/api/confirm-email`;
 export const CLUSTER_STATUS_URL = `${api}/api/status`;
 export const GITLAB_AUTH_URL = `${api}/api/gitlab/connect`;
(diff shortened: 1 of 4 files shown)


tiborsimko (approved):
Works nicely :+1: 

Some release-news-oriented comments:

- the headlines could announce new functionality in a feature-oriented verbiage, for example:

```
feat(gitlab): add pagination for GitLab project lists (#403)
feat(gitlab): add filtering for GitLab project lists (#403)
```

- The breaking change news text could be altered to say that the endpoint response is not being modified here in r-ui, but that the component is only adapting to the new version from r-server, something like:

```
BREAKING CHANGE: The frontend code adapts to the new REST API server  `gitlab_projects` endpoint which now includes the pagination details.
```

OTOH, it may be nice to use the same breaking change text in both r-server and r-ui, for easier referencing? In which case we could use some neutral language such as: 

```
BREAKING CHANGE: The REST API endpoint `gitlab_projects` now includes pagination details.
```

in both r-server and r-ui.

PR #405, opened 2024-04-16
feat(sessions): add modal for choosing session environment (#405)

mdonadoni:
Closes reanahub/reana-workflow-controller#569

### DIFF
--- a/reana-ui/src/components/InteractiveSessionModal.js
+++ b/reana-ui/src/components/InteractiveSessionModal.js
@@ -1,0 +1,228 @@
+/*
+  -*- coding: utf-8 -*-
+
+  This file is part of REANA.
+  Copyright (C) 2024 CERN.
+
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+import { useCallback, useEffect, useMemo, useState } from "react";
+import { useDispatch, useSelector } from "react-redux";
+import {
+  Button,
+  Form,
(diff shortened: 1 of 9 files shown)


reana-ui/src/components/InteractiveSessionModal.js
```diff
@@ -0,0 +1,194 @@
...
+              onChange={(_, { value }) => setRecommendedImage(value)}
+              search={true}
+            />
+          )}
+          {environmentType === EnvironmentType.Custom && (
+            <FormInput
+              label="Custom environment"
+              value={customImage}
+              onChange={(_, { value }) => setCustomImage(value)}
+              placeholder={"Custom container image"}
```
tiborsimko, line 180: When custom environment images are enabled, the user may try to type fancy values such as:

```
alert("XSS")
```

This leads to a success message:

```
Success!

The interactive session has been created. However, it could take several minutes to start the Jupyter Notebook. Click on the Jupyter logo to access it. 
```

And on the backend, the session pod is created but obviously not started due to the `InvalidImageName` error.

Could we plug some value sanitisation checks so that we would prevent even trying to open invalid image strings?

(Ideally on the backend side, so that we would also cover `reana-client open`, but we could also eliminate some characters already in the front-end perhaps, such as in the various XSS test strings, as tried above.)

mdonadoni: I think we should do this as part of another issue/PR, as these checks could be also done for example for the docker images specified in `reana.yaml`

PR #416, opened 2024-11-18
fix(ui): show jupyter badge only after the pod is running (#416)

### LINKED ISSUE #408: Show Jupyter icon when only when notebook is available
Currently, a message warns users that the Notebook may not be available for a few minutes. However, the Jupyter icon is displayed while the Notebook is not available, and clicking on it generates an error. This is confusing. The icon should be displayed only when the Notebook is available.

Alputer:
Closes #408


reana-ui/src/components/WorkflowActionsPopup.js
```diff
@@ -38,7 +38,7 @@ export default function WorkflowActionsPopup({ workflow, className }) {
   const isDeleted = status === "deleted";
   const isDeletedUsingWorkspace = isDeleted && size.raw > 0;
   const isRunning = status === "running";
-  const isSessionOpen = sessionStatus === "created";
+  const isSessionOpen = sessionStatus === "running";
```
tiborsimko, line 41: Works nicely :+1: 

BTW, now that we have a new design, we could amend the text _"Click on the Jupyter logo to access it"_ to say something like "Click on the Jupyter notebook badge to access it."

tiborsimko: Cosmetics: for the commit log, please use capital letter for Jupyter:

```
fix(ui): show Jupyter badge only after notebook is running (#416)
```

tiborsimko (approved):

PR #429, opened 2025-07-18
fix(workflow-list): users with no workflows can now see workflows shared with them

### LINKED ISSUE #413: ui: user with no workflows can't see shared workflows
Currently, if a user has never run any workflow, but some other user has shared workflows with him, he will not be able to see the list of shared workflows, as the homepage will instead show the welcome message:
https://github.com/reanahub/reana-ui/blob/660584e53d4c63d100bad73c39e674af12022824/reana-ui/src/pages/workflowList/WorkflowList.js#L125-L127

CameronMcClymont:
The initial view still shows no workflows as filters are set to 'owned by = you' by default:

image

But the user can change the filters to see workflows shared with them:

image

Closes #413


CameronMcClymont:
> Hi @CameronMcClymont, there are some merge conflicts here, should I take over the UI PRs?

@michaelbuchar yes please!

PS, commit message improved now.

michaelbuchar (approved):

PR #430, opened 2025-07-18
fix(profile): hide GitLab box when not configured by the admins

### LINKED ISSUE #428: your profile: hide GitLab box when not configured by the admins
### Current behaviour

When an admin installs REANA from Helm charts, the "Your profile" page shows currently the option to link to GitLab, even if the admins did not configure any GitLab integration:

Image

Clicking on the button leads to:

Image

### Expected behavior

The box should not be offered if the REANA deployment is not connected to any GitLab instance.

### Notes

This task may consist of enriching `reana-client info` related endpoints and configmaps in order to expose the information whether an instance is connected to a GitLab server (and if so, to which one) to both `reana-ui` and `reana-client`. The web intergace could then conditionally display the box, when applicable. And the user would be also informed in the CLI client about this.

See also #47

CameronMcClymont:
Hides this section of the profile when REANA_GITLAB_HOST is not set:

image

Currently the Helm value has a default so REANA_GITLAB_HOST is always set?

Closes #428


CameronMcClymont:
It looks like the Helm value in `reana/helm/reana/templates/secrets.yaml` has a default value, in which case `REANA_GITLAB_HOST` will always be set? Will this be the case on non-CERN instances?

```
REANA_GITLAB_HOST: {{ .Values.secrets.gitlab.REANA_GITLAB_HOST | default "gitlab.cern.ch" | b64enc }}
```

CameronMcClymont:
I have had a look through the codebases and `REANA_GITLAB_HOST` is only used in `reana`, `reana-server`, and `reana-workflow-controller`. In all cases, if `REANA_GITLAB_HOST` is an empty string there should be no issues.

michaelbuchar:
Hi @CameronMcClymont, this PR needs rebasing when you get a moment

michaelbuchar:
> > In all cases, if `REANA_GITLAB_HOST` is an empty string there should be no issues.
> 
> BTW setting this to empty string does not work for me anymore. We may need a small PR for the "reana" repository to handle empty strings better. (Or switch the default from "gitlab.cern.ch" to "" which would be a safer default for distribution world-wide anyway.)

Created here: https://github.com/reanahub/reana/pull/922

PR #431, opened 2025-07-22
feat(filtering): add owned_by "others" and shared_with "nobody" as workflow filters

CameronMcClymont:
### To test

- Run REANA server and UI locally
- Create a couple different users using `reana-admin`
- Run some demo workflows on each user
- Have each user share a workflow or two, leaving some unshared
- Try out the new filters from various users' perspectives

Closes https://github.com/reanahub/reana-server/issues/661


michaelbuchar:
Hi @CameronMcClymont, unfortunately there is a merge conflict here now :/

PR #432, opened 2025-07-23
docs(readme): update 'Local development' section

CameronMcClymont:
Write more comprehensive docs on how to run REANA-UI locally, especially for macOS developers.


README.md
```diff
@@ -25,16 +25,52 @@ and historical workflows.
 The detailed information on how to install and use REANA can be found in
 [docs.reana.io](https://docs.reana.io).
 
-## Development
+## Local development
```
tiborsimko, line 28: Some notes:

- I'd call the section simply "Development" , and explain the purpose in the first paragraph.
- It's better not to have subsections, since they are not numbered, a flat chapter structure will be more easily readable.
- `nvm` is one technique, another is `mise` (e.g. I'm using the latter), so we can strongly say "for example" in places.

Here is a diff proposal on how the docs could look like after the above changes:

https://simko.web.cern.ch/tmp/reana-ui-readme.md

Just for illustration; the language may need polishing etc.

CameronMcClymont: Thanks for the feedback - I have pushed changes based on your diff and a few tweaks.

tiborsimko (approved):
Thanks, I've updated the commit log headline (since we don't have "Local development" section strictly speaking anymore after renaming) as well as removed Yarn section there for simplicity.

Issue #435, opened 2025-09-19
list: display only workflows that have interactive sessions opened


tiborsimko:
### Current behaviour

On the web interface, the researcher sees a workflow list. Some of the workflows may have opened a Jupyter notebook interactive session. These have a "Notebook" badge displayed:

<img width="1056" height="996" alt="Image" src="https://github.com/user-attachments/assets/c14ef88d-4ec8-4baf-9055-55e672533641" />

The problem is when a researcher has many hundreds of workflow runs and so many pages to list: the web interface does not allow to easily filter which workflow runs have opened notebooks and which do not, so some of the notebook sessions from older runs may get "forgotten".

(It is possible to see this from the command line by using `reana-client list --sessions`, but not from the web interface.

### Expected behaviour

It would be useful to add a new filter to the web interface page allowing to list only workflows that have notebooks attached.

This could look similarly to "Show deleted runs" in the above screenshot.

### Notes

- Note also that the researcher could also launch Dask workflows, so a similar functionality may be later needed for Dask. The interface design could think ahead about this possibility as well.

- Alternatively, we have a filtering based by name already in the web interface, so we could enrich query syntax so that wehn peolpe would type "notebook" or "jupyter" the list would filter those workflow runs that have interactive sessions opened. This would not require to add an explicitl toggle such as "Show deleted runs", and could be more evolution-friendly towards Dask and other services in the future.

- Alternatively, instead of the serch interface filter, one can also think about "Your notebooks" page displayed alongside "Your workflows" page, as in a sort of new tab, which would provide this filtering.

PR #436, opened 2025-09-23
feat(workflow-list): add interactive session filter

michaelbuchar:
Add Only with sessions toggle bar to display only workflows that have interactive sessions opened.

Closes https://github.com/reanahub/reana-ui/issues/435

Screenshot 2025-09-24 at 10 24 57 AM
Screenshot 2025-09-24 at 10 25 03 AM

### DIFF
--- a/reana-ui/src/pages/workflowList/WorkflowList.js
+++ b/reana-ui/src/pages/workflowList/WorkflowList.js
@@ -114,6 +114,26 @@ function Workflows() {
     () => searchParams.get("sort") || "desc",
     [searchParams],
   );
+  const interactiveOnlyFilter = useMemo(
+    () => searchParams.get("open-sessions") === "true",
+    [searchParams],
+  );
+
+  const setInteractiveOnlyFilterInUrl = (on) => {
+    setSearchParams(
+      (prev) => {
+        const qp = new URLSearchParams(prev);
+        if (on) {
+          qp.set("open-sessions", "true");
+        } else {
+          qp.delete("open-sessions");
+        }
+        qp.delete("page"); // reset pagination when filter changes
+        return qp;
+      },
+      { replace: false },
+    );
+  };
 
   // URL writers
   const setStatusFilterInUrl = (nextStatus) => {
@@ -282,6 +302,7 @@ function Workflows() {
       sharedBy,
       sharedWith: sharedWithParam,
       sort: sortDir,
+      ...(interactiveOnlyFilter ? { type: "interactive" } : {}),
     };
   }, [
     pagination,
@@ -293,6 +314,7 @@ function Workflows() {
     sharedWithFilter,
     sharedWithMode,
     sortDir,
+    interactiveOnlyFilter,
   ]);
 
   const lastParamsRef = useRef();
@@ -395,6 +417,8 @@ function Workflows() {
           showDeleted={showDeletedMode}
           setShowDeleted={setShowDeletedInUrl}
           statusExplicit={statusExplicit}
+          interactiveOnlyFilter={interactiveOnlyFilter}
+          setInteractiveOnlyFilter={setInteractiveOnlyFilterInUrl}
           ownedByFilter={ownedByFilter}
           setOwnedByFilter={setOwnedByFilter}
           sharedWithFilter={sharedWithFilter}
--- a/reana-ui/src/pages/workflowList/WorkflowList.module.scss
+++ b/reana-ui/src/pages/workflowList/WorkflowList.module.scss
@@ -43,7 +43,7 @@
 }
 
 :global(.ui.text.container).workflow-list-container {
-  max-width: 825px !important;
+  max-width: 1200px !important;
 }
 
 .paginationRow {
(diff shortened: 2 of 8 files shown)


reana-ui/src/pages/workflowList/WorkflowList.js
```diff
@@ -108,6 +110,7 @@ function Workflows() {
             sharedWith: sharedWithFilter,
             sort: sortDir,
             showLoader,
+            ...(sessionFilter ? { type: "interactive" } : {}),
```
CameronMcClymont, line 113: Are there any other workflow types in REANA besides 'interactive'? If not, I wonder why this isn't a boolean? Unless more types are planned for the future. Anyone know?

reana-ui/src/pages/workflowList/WorkflowList.js
```diff
@@ -50,6 +50,7 @@ function Workflows() {
   const [refreshedAt, setRefreshedAt] = useState(currentUTCTime());
   const [pagination, setPagination] = useState({ page: 1, size: PAGE_SIZE });
   const [statusFilter, setStatusFilter] = useState(NON_DELETED_STATUSES);
+  const [sessionFilter, setSessionFilter] = useState(false);
```
CameronMcClymont, line 53: Maybe this is better called `typeFilter` to stay consistent? Or `interactiveOnly`? At least for me 'session' makes me think it refers to browser sessions or tokens or something

reana-ui/src/pages/workflowList/WorkflowList.module.scss
```diff
@@ -43,5 +43,5 @@
 }
 
 :global(.ui.text.container).workflow-list-container {
-  max-width: 825px !important;
+  max-width: 1200px !important;
```
CameronMcClymont, line 46: I think it would be nice to keep this at `825px` - maybe the switch filters could be stacked vertically to save some width?

reana-ui/src/pages/workflowList/components/WorkflowSessionFilters.js
```diff
@@ -0,0 +1,58 @@
...
+
+export default function WorkflowSessionFilters({
+  enabled: enabledProp,
+  filter,
+}) {
+  const [enabled, setEnabled] = useState(false);
+
+  useEffect(() => {
+    // Keep local UI in sync with the parent value
+    if (!isEqual(enabled, enabledProp)) {
```
CameronMcClymont, line 24: For my own learning, why use `isEqual` here instead of `!==` if they are both booleans?

reana-ui/src/pages/workflowList/components/WorkflowSessionFilters.js
```diff
@@ -0,0 +1,58 @@
...
+import { useEffect, useState } from "react";
+import { Checkbox, Grid } from "semantic-ui-react";
+
+export default function WorkflowSessionFilters({
+  enabled: enabledProp,
+  filter,
+}) {
+  const [enabled, setEnabled] = useState(false);
+
+  useEffect(() => {
```
CameronMcClymont, line 22: Can't you remove this whole `useEffect` if you replace the `enabled: enabledProp` with a new `initiallyEnabled` prop and then have `const [enabled, setEnabled] = useState(initiallyEnabled);`?

reana-ui/src/pages/workflowList/components/WorkflowSessionFilters.js
```diff
@@ -0,0 +1,58 @@
...
+        mobile={16}
+        tablet={4}
+        computer={3}
+        className="center aligned"
+      >
+        <Checkbox
+          toggle
+          label="Only with sessions"
+          checked={enabled}
+          onChange={(_, { checked }) => setEnabled(!!checked)}
```
CameronMcClymont, line 48: Clean 😍

CameronMcClymont (commented):
Seems to work well from my brief testing!

reana-ui/src/pages/workflowList/WorkflowList.js
michaelbuchar: I'm not sure, I know there is batch and interactive. There might be others types for launching Dask workflows, Jupyter Notebooks etc.

reana-ui/src/pages/workflowList/WorkflowList.js
michaelbuchar: Good point 👍 , I used sessionFilter because the API response uses fields like session_status/session_type, but I agree it could be confusing since “session” might also imply browser/session tokens. I’m happy to rename it for clarity, interactiveOnly feels the most descriptive of what the toggle actually does. Does that work for you, or would you prefer typeFilter to keep it consistent with other filters?

reana-ui/src/pages/workflowList/WorkflowList.module.scss
michaelbuchar: I see what you mean, stacking vertically would allow us to keep 825px. The reason I bumped it to 1200px is that at 825px things felt pretty tight, and Tibor suggested we keep the filters in one line, which needs more horizontal space. Given that, I’d lean toward the wider layout so the filters don’t feel cramped. But if the team prefers keeping 825px and stacking vertically, happy to align on whichever direction we decide!

reana-ui/src/pages/workflowList/WorkflowList.js
michaelbuchar: Changed!

PR #442, opened 2025-10-22
feat(workflow-list): add more pagination size options 

michaelbuchar:
- Replaced native select with Semantic UI Dropdown
- Updated default page size from 5 to 10
- Pagination and results per page dropdown aligned on same row
- Added new SCSS styles
- Removed unnecessary overflow in WorkflowSharingFilter style

### DIFF
--- a/reana-ui/src/pages/workflowList/WorkflowList.js
+++ b/reana-ui/src/pages/workflowList/WorkflowList.js
@@ -12,7 +12,7 @@ import moment from "moment";
 import { useEffect, useMemo, useRef, useState } from "react";
 import { useSearchParams } from "react-router-dom";
 import { useDispatch, useSelector } from "react-redux";
-import { Container, Dimmer, Icon, Loader } from "semantic-ui-react";
+import { Container, Dimmer, Dropdown, Icon, Loader } from "semantic-ui-react";
 import isEqual from "lodash/isEqual";
 
 import { fetchWorkflows } from "~/actions";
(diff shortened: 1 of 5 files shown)


reana-ui/src/pages/workflowList/WorkflowList.js
```diff
@@ -64,7 +64,14 @@ function Workflows() {
     const n = parseInt(searchParams.get("page") || "", 10);
     return Number.isFinite(n) && n > 0 ? n : 1;
   }, [searchParams]);
-  const pagination = useMemo(() => ({ page, size: PAGE_SIZE }), [page]);
+  const [pageSize, setPageSize] = useState(() => {
+    const n = parseInt(searchParams.get("page-size") || "", 10);
```
CameronMcClymont: Same again here, the `base` parameter is 10 by default

reana-ui/src/pages/workflowList/WorkflowList.js
```diff
@@ -61,10 +66,17 @@ function Workflows() {
...
 
   const page = useMemo(() => {
-    const n = parseInt(searchParams.get("page") || "", 10);
+    const n = parseInt(searchParams.get("page") || "");
     return Number.isFinite(n) && n > 0 ? n : 1;
   }, [searchParams]);
-  const pagination = useMemo(() => ({ page, size: PAGE_SIZE }), [page]);
+  const [pageSize, setPageSize] = useState(() => {
+    const n = parseInt(searchParams.get("page-size") || "");
+    return Number.isFinite(n) && n > 0 ? n : DEFAULT_PAGE_SIZE;
```
CameronMcClymont: It could be worth extracting this `Number.isFinite(n) && n > 0` check into its own `isValidPageSize(n)` function since it's used in multiple places and it would make it easier to change the definition of a 'valid' page size in the future.

PR #469, opened 2026-01-30
feat(workflow-delete): guard deletion of restarted runs

### LINKED ISSUE #448: Deleting a restarted workflow run removes the original workflow workspace
### **Description**

When restarting a workflow and deleting the restarted run via the UI, the workspace of the original workflow run is removed, leaving the original run in an inconsistent (out-of-sync) state.

This results in a finished workflow whose workspace no longer exists on disk, which is later detected by `reana-admin check-workflows` as an error.

---

### **Steps to reproduce**

1. Run a workflow:

   ```bash
   reana-client run -w helloworld-demo
   ```

   This creates workflow run `helloworld-demo.3`.

2. Restart the workflow:

   ```bash
   reana-client restart -w helloworld-demo
   ```

   This creates workflow run `helloworld-demo.3.1`.

3. Delete the restarted workflow run (`helloworld-demo.3.1`) using the REANA UI. Do not check the `Delete all the runs of the workflow` checkbox.

4. Run the admin consistency check:

   ```bash
   kubectl exec -i -t deployment/reana-server -- flask reana-admin check-workflows --date-start '2025-12-01' --admin-access-token $REANA_ADMIN_ACCESS_TOKEN   
   ```

---

### **Observed behavior**

The admin check reports the original workflow run (`helloworld-demo.3`) as out-of-sync:

```
ERROR: The workspace '/var/reana/users/.../workflows/<workflow-id>' does not exist.
```

The original workflow run:

* Is marked as `finished` in the database
* No longer has a corresponding workspace on disk

---

### **Expected behavior**

One of the following should happen instead:

1. The workflow workspace of the original run should **not** be deleted when deleting a restarted run
2. The user should be clearly warned that deleting the restarted run will also remove the workspace of the original run and result in data loss
3. All related runs (e.g. `3` and `3.1`) should have their statuses updated consistently (e.g. both marked as deleted).

In any case, deleting a restarted run should not leave the original run in a broken, out-of-sync state.

michaelbuchar:
Closes #448

### DIFF
--- a/reana-ui/src/components/WorkflowDeleteModal.js
+++ b/reana-ui/src/components/WorkflowDeleteModal.js
@@ -8,99 +8,412 @@
   under the terms of the MIT License; see LICENSE file for more details.
 */
 
-import { useEffect, useState } from "react";
+import { useCallback, useEffect, useMemo, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import { Button, Checkbox, Icon, Message, Modal } from "semantic-ui-react";
 
 import { closeDeleteWorkflowModal, deleteWorkflow } from "~/actions";
 import client from "~/client";
-import { NON_DELETED_STATUSES } from "~/config";
+import styles from "./WorkflowDeleteModal.module.css";
+import {
+  NON_DELETED_STATUSES,
+  NON_FINISHED_STATUSES,
+  WORKFLOW_RUNS_PREFETCH_PAGE_SIZE,
+} from "~/config";
 import {
   getWorkflowDeleteModalItem,
   getWorkflowDeleteModalOpen,
 } from "~/selectors";
 
-export default function WorkflowDeleteModal() {
-  const dispatch = useDispatch();
-  const open = useSelector(getWorkflowDeleteModalOpen);
-  const workflow = useSelector(getWorkflowDeleteModalItem);
-  const [allRuns, setAllRuns] = useState(false);
-  const [allRunsTotal, setAllRunsTotal] = useState();
+// Max number of runs to show inline in checkbox labels before truncating
+const MAX_RUN_LABELS_SHOWN = 6;
+const DELETE_RETRY_ATTEMPTS = 2;
 
-  const { id, name, run, size } = workflow ?? {};
+const getErrorMessage = (err) =>
+  err?.response?.data?.message ||
+  err?.response?.data?.error ||
+  err?.message ||
+  String(err);
+
+function DeleteWarningMessage({ sizeHumanReadable, hasRelatedRuns }) {
+  return (
+    <Message icon warning>
+      <Icon name="warning circle" />
+      <Message.Content>
+        This action will delete also the workflow&apos;s workspace
+        {sizeHumanReadable ? ` (${sizeHumanReadable})` : ""} and any open
+        interactive session attached to it. Please make sure to download all the
+        files you want to keep before proceeding.
+        {hasRelatedRuns && (
+          <>
+            <br />
+            <br />
+            This workflow run has restarts. All restarts share the same
+            workspace. Deleting this run would remove the shared workspace and
+            leave the other runs in an inconsistent state. If you proceed, the
+            related runs will also be marked as deleted.
+          </>
+        )}
+      </Message.Content>
+    </Message>
+  );
+}
+
+function useWorkflowRuns({ open, name, run }) {
+  const [state, setState] = useState({
+    total: null,
+    items: [],
+    relatedIds: [],
+    loading: false,
+  });
 
   useEffect(() => {
-    // reset local state on workflow change
-    setAllRuns(false);
-    setAllRunsTotal(null);
-  }, [id]);
-
-  useEffect(() => {
-    if (!name) {
+    if (!open || !name) {
+      setState({ total: null, items: [], relatedIds: [], loading: false });
       return;
     }
+
+    let cancelled = false;
+    const selectedFullName = run ? `${name}.${run}` : name;
+    const selectedGroup = getWorkspaceGroup(selectedFullName);
+
+    setState((s) => ({ ...s, loading: true }));
 
     client
       .getWorkflows({
         workflowIdOrName: name,
         status: NON_DELETED_STATUSES,
-        // request minimum number of workflows as we are only interested in `total`
-        pagination: { size: 1, page: 1 },
+        pagination: { size: WORKFLOW_RUNS_PREFETCH_PAGE_SIZE, page: 1 },
       })
-      .then((resp) => setAllRunsTotal(resp.data.total))
-      .catch((err) =>
-        console.log(
-          `Error while fetching number of workflows with name ${name}`,
-          err,
-        ),
+      .then((resp) => {
+        if (cancelled) return;
+        const items = resp.data.items || [];
+        const relatedIds = selectedGroup
+          ? items
+              .filter((w) => getWorkspaceGroup(w?.name) === selectedGroup)
+              .map((w) => w.id)
+              .filter(Boolean)
+          : [];
+        setState({ total: resp.data.total, items, relatedIds, loading: false });
+      })
+      .catch((err) => {
+        if (cancelled) return;
+        console.error(`Error while fetching runs for workflow ${name}`, err);
+        setState({ total: null, items: [], relatedIds: [], loading: false });
+      });
+
+    return () => {
+      cancelled = true;
+    };
+  }, [open, name, run]);
+
+  return state;
+}
+
+// Returns the workspace group for a full workflow name
+// helloworld-demo.1   -> helloworld-demo.1
+// helloworld-demo.1.1 -> helloworld-demo.1
+// helloworld-demo.1.2 -> helloworld-demo.1
+const getWorkspaceGroup = (fullName) => {
+  if (!fullName) return null;
+  const parts = String(fullName).split(".");
+  let i = parts.length - 1;
+  while (i >= 0 && /^\d+$/.test(parts[i])) i--;
+  const numeric = parts.slice(i + 1); // ["7","1"]
+  const base = parts.slice(0, i + 1).join("."); // "helloworld-demo"
+  if (!base || numeric.length === 0) return null;
+  return `${base}.${numeric[0]}`; // first numeric suffix = original run
+};
+
+// Formats a full workflow run name into a short human-readable run label
+// Examples:
+// helloworld-demo.7   -> #7
+// helloworld-demo.7.1 -> #7.1
+const formatRunLabel = (fullName) => {
+  if (!fullName) return "";
+  const parts = String(fullName).split(".");
+  let i = parts.length - 1;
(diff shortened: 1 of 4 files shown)


reana-ui/src/components/WorkflowDeleteModal.js
```diff
@@ -20,87 +20,308 @@ import {
...
+    const selectedGroup = getWorkspaceGroup(selectedFullName);
+
+    setState((s) => ({ ...s, loading: true }));
+
+    client
+      .getWorkflows({
+        workflowIdOrName: name,
+        status: NON_DELETED_STATUSES,
+        // request minimum number of workflows as we are only interested in total
+        pagination: { size: 1000, page: 1 },
```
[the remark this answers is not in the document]
tiborsimko: BTW isn't size 1000 impacting performance? The comment says we are fetching the minimum number of workflows, but that does not seem to be the case...

reana-ui/src/components/WorkflowDeleteModal.js
```diff
@@ -20,87 +20,308 @@ import {
...
-  }, [name]);
+  const onToggleAllRuns = useCallback((_, data) => {
+    const checked = Boolean(data.checked);
+    setAllRuns(checked);
+    // If user opts to delete all runs, related runs are necessarily deleted too
+    if (checked) setConfirmRelatedDeletion(true);
+  }, []);
 
-  if (!workflow) return null;
+  const onDelete = useCallback(async () => {
```
tiborsimko: AI: Missing error handling in `onDelete`.

```
const onDelete = useCallback(async () => {   
  // ...         
  await dispatch(deleteWorkflow(id, { allRuns: true }));  
  // No try/catch - if delete fails, user gets no feedback                                                                                                   
```

The async function doesn't catch errors. If deletion fails, the modal may close without informing the user.

reana-ui/src/components/WorkflowDeleteModal.js
```diff
@@ -8,99 +8,382 @@
...
 
-  const onCloseModal = () => {
-    dispatch(closeDeleteWorkflowModal());
-  };
+      // If this run has restarts, delete workspace once and mark related runs deleted too
+      if (hasRelatedRuns) {
+        if (!confirmRelatedDeletion) return;
+
+        await dispatch(
+          deleteWorkflow(id, { allRuns: false, deleteWorkspace: true }),
```
tiborsimko, line 262: ### AI.1 Parameter name mismatch: `deleteWorkspace` vs `workspace`

The new code passes `deleteWorkspace` but the action destructures `workspace`:

```js
// WorkflowDeleteModal.js passes:
deleteWorkflow(id, { allRuns: false, deleteWorkspace: true })
deleteWorkflow(id, { allRuns: false, deleteWorkspace: false })

// actions.js expects:
deleteWorkflow(id, { workspace = true, allRuns = false } = {})
```

Since `deleteWorkspace` is never destructured, `workspace` silently defaults to `true` in every call. The "skip workspace deletion for related runs" path is broken — every related run deletion will attempt to delete the already-deleted workspace.

**Fix:** Replace `deleteWorkspace` with `workspace` in all three call sites in `WorkflowDeleteModal.js`.

reana-ui/src/config.js
```diff
@@ -67,3 +67,9 @@ export const NON_DELETED_STATUSES = WORKFLOW_STATUSES.filter(
  * List of HTML tags that can be present in an announcement.
  */
 export const ALLOWED_HTML_TAGS_ANNOUNCEMENT = ["a", "b", "em", "i", "strong"];
+
+/**
+ * Pagination size used when we prefetch workflow runs in a single request
+ * (delete modal needs the list to detect related restarts sharing a workspace).
+ */
+export const WORKFLOW_RUNS_PREFETCH_PAGE_SIZE = 1000;
```
tiborsimko, line 75: ### AI.2 Truncation of runs not handled

`WORKFLOW_RUNS_PREFETCH_PAGE_SIZE = 1000` caps the prefetch, but if `total > 1000`, the `relatedRunIds` list will be incomplete. Some related runs could be missed and left in an inconsistent state after deletion.

**Fix:** Either warn the user when `total > items.length`, or paginate until all runs are fetched.

reana-ui/src/components/WorkflowDeleteModal.js
```diff
@@ -8,99 +8,382 @@
...
+        );
+
+        // delete related runs without deleting workspace again
+        const otherIds = relatedRunIds.filter((otherId) => otherId !== id);
+        const results = await Promise.allSettled(
```
tiborsimko, line 267: ### AI.3 Partial deletion with no recovery

If some related run deletions fail in the `Promise.allSettled` path, the primary run and its workspace are already deleted. The remaining orphaned runs reference a deleted workspace with no way to retry or recover.

**Fix:** Consider deleting the workspace last (after all related runs are marked deleted), or add a retry mechanism for failed deletions.

reana-ui/src/components/WorkflowDeleteModal.module.css
```diff
@@ -0,0 +1,3 @@
+.allRunsCheckboxSpacing {
```
tiborsimko, line 1: ### AI.4 Missing license header in new CSS file

The new `WorkflowDeleteModal.module.css` file lacks the REANA MIT license header that other source files in the project include.

**Fix:** Add the standard REANA copyright/license comment block at the top of the file.

reana-ui/src/components/WorkflowDeleteModal.js
```diff
@@ -8,99 +8,382 @@
...
-  }, [name]);
+  const onToggleAllRuns = useCallback((_, data) => {
+    const checked = Boolean(data.checked);
+    setAllRuns(checked);
+    // If user opts to delete all runs, related runs are necessarily deleted too
+    if (checked) setConfirmRelatedDeletion(true);
+  }, []);
 
-  if (!workflow) return null;
+  const onDelete = useCallback(async () => {
```
tiborsimko, line 245: ### AI.5 Duplicate error feedback

The original `deleteWorkflow` action's `.catch()` dispatches `errorActionCreator`, which shows a toast notification. The new `onDelete` function also catches errors locally and sets `deleteError` state, displaying an in-modal error message. The user may see both a toast and an inline error for the same failure.

**Fix:** Either suppress the action's built-in error notification when called from this modal, or remove the local `deleteError` state and rely solely on the existing toast.

reana-ui/src/components/WorkflowDeleteModal.js
```diff
@@ -8,99 +8,382 @@
...
-      return;
+  const hasRelatedRuns = relatedRunIds.length > 1;
+  const showAllRunsCheckbox = Boolean(allRunsTotal && allRunsTotal > 1);
+
+  const onCloseModal = useCallback(() => {
+    dispatch(closeDeleteWorkflowModal());
+  }, [dispatch]);
+
+  const selectedFullName = run ? `${name}.${run}` : name;
+  const selectedGroup = useMemo(
```
tiborsimko, line 180: ### AI.6 Excessive `useMemo` usage

Several `useMemo` calls wrap trivial operations like `labels.join(", ")` or `formatRunList()` where the memoization overhead likely exceeds the computation cost. Affected: `relatedRunListText`, `relatedRunListTitle`, `allRunListText`, `allRunListTitle`, `selectedGroup`.

**Fix:** Remove `useMemo` from these trivial derivations and compute them inline.

reana-ui/src/components/WorkflowDeleteModal.js
```diff
@@ -8,99 +8,382 @@
...
+          ? items
+              .filter((w) => getWorkspaceGroup(w?.name) === selectedGroup)
+              .map((w) => w.id)
+              .filter(Boolean)
+          : [];
+        setState({ total: resp.data.total, items, relatedIds, loading: false });
+      })
+      .catch((err) => {
+        if (cancelled) return;
+        console.log(`Error while fetching runs for workflow ${name}`, err);
```
tiborsimko, line 99: ### AI.7 `console.log` should be `console.error`

In the `useWorkflowRuns` hook's `.catch()`, `console.log` is used for an error condition. The `onDelete` function correctly uses `console.error` for the same purpose.

**Fix:** Change `console.log` to `console.error` in the `useWorkflowRuns` catch block for consistency.

reana-ui/src/components/WorkflowDeleteModal.js
michaelbuchar: Fixed, nice catch!

reana-ui/src/config.js
michaelbuchar: Created a warning, thanks!

reana-ui/src/components/WorkflowDeleteModal.js
michaelbuchar: Added a retry mechanism!

reana-ui/src/components/WorkflowDeleteModal.module.css
michaelbuchar: Done!

reana-ui/src/components/WorkflowDeleteModal.js
michaelbuchar: Also nice catch, this is correct, thanks!

reana-ui/src/components/WorkflowDeleteModal.js
michaelbuchar: Agreed, changed!

reana-ui/src/components/WorkflowDeleteModal.js
michaelbuchar: Fixed!

reana-ui/src/components/WorkflowDeleteModal.js
tiborsimko: One minor concern worth noting: Every run in the loop is deleted with workspace: true, meaning each deletion asks the server to remove the workspace. The
  original design intended workspace: false for related runs (though it was broken by AI.1). If the server handles redundant workspace-deletion requests
  idempotently this is fine, but it's worth verifying — otherwise the first related run deletes the workspace, and subsequent runs may error trying to delete
  an already-gone workspace.

michaelbuchar: workspace=True is intentional here... workspace=False does not reliably mark the related run as deleted. The delete path requires workspace deletion and requests with workspace=False end up being rejected/ignored rather than deleting the workflow.

tiborsimko:
Commit headline suggestions:

1. feat(workflow-delete): guard deletion of restarted runs (#469)
2. feat(workflow-delete): warn before deleting restarted runs (#469)

Commit body suggestion:

    Prevent deleting a single workflow run when it belongs to a
    restart series sharing the same workspace, since that would
    leave sibling runs in an inconsistent state. The delete modal
    now warns about shared workspaces and offers to delete all
    runs in the restart series together.

    Closes #448

See also https://github.com/reanahub/reana-client/pull/759#discussion_r3116238631 - we can share the same terminology for both Web UI and CLI UI

tiborsimko (approved):

PR #491, opened 2026-03-31
fix(auth): redirect to sign-in page on session expiry

tiborsimko:
When a user's server-side session expires (e.g. after the configured session TTL), navigating within the UI still shows stale page content with only an error notification at the top, because the Redux auth state retains the user's email from the previous successful load. The user has to hard-reload the browser to actually see the sign-in page.

Fix this by registering a global 401 response handler on the HTTP client that clears the auth state whenever any API call returns Unauthorized. This causes the RequireAuth route guard to immediately redirect to the sign-in page, regardless of which API call first detects the expired session.

The handler skips "no active tokens" 401 responses, which are returned for SSO users who have a valid session but have not yet been granted a REANA API token. Without this exception, such users would be caught in a sign-in redirect loop.

### DIFF
--- a/reana-ui/src/actions.js
+++ b/reana-ui/src/actions.js
@@ -14,6 +14,7 @@ import client, {
   CONFIG_URL,
   INTERACTIVE_SESSIONS_CLOSE_URL,
   INTERACTIVE_SESSIONS_OPEN_URL,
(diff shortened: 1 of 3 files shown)


michaelbuchar (approved):
Looks good to me!

PR #492, opened 2026-04-01
fix(auth): limit self-logout to session expiry 401 responses

tiborsimko:
The global 401 handler previously logged out users on any 401 response (except "no active tokens"), which mistakenly included also 401 responses from GitLab, which may have incorrectly redirect users to the sign-in page in case of GitLab token issues.

This commit fixes the problem by analysing 401 responses and by triggering self-logout only when the 401 response messages match known session-expiry messages "User not signed in" or "User not logged in".

### DIFF
--- a/reana-ui/src/client.js
+++ b/reana-ui/src/client.js
@@ -21,6 +21,16 @@ export function isNoActiveTokensError(error) {
   );
 }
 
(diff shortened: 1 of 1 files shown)


michaelbuchar (approved):
Works on my end! 🚀

PR #494, opened 2026-04-15
feat(profile): show CPU quota period window

michaelbuchar:
Closes https://github.com/reanahub/reana-server/issues/774

Screenshot 2026-04-16 at 9 31 41 AM

### DIFF
--- a/reana-ui/src/pages/profile/components/Quota.js
+++ b/reana-ui/src/pages/profile/components/Quota.js
@@ -11,6 +11,7 @@
 import { useSelector } from "react-redux";
 import { Grid, Label } from "semantic-ui-react";
 import isEmpty from "lodash/isEmpty";
+import moment from "moment";
 
 import { PieChart, Notification } from "~/components";
 import { REANA_QUOTAS_DOCS_URL } from "~/config";
@@ -18,6 +19,22 @@ import { getConfig, getUserQuota } from "~/selectors";
 import { healthMapping } from "~/util";
 
 import styles from "./Quota.module.scss";
+
+export function getQuotaPeriodWindow(quota = {}) {
+  const {
+    quota_period_months: periodMonths,
+    quota_period_start_at: periodStartAt,
+  } = quota;
+  if (!periodMonths || !periodStartAt) return null;
+
+  const periodStart = moment.utc(periodStartAt);
+  if (!periodStart.isValid()) return null;
+
+  return {
+    startLabel: periodStart.format("ll"),
+    endLabel: periodStart.clone().add(periodMonths, "months").format("ll"),
+  };
+}
 
 export default function Quota() {
   const config = useSelector(getConfig);
@@ -27,6 +44,7 @@ export default function Quota() {
     const { usage, limit, health } = quota;
     const percentage = Math.round((usage?.raw / limit?.raw) * 100);
     const hasLimit = limit?.raw > 0;
+    const quotaPeriodWindow = getQuotaPeriodWindow(quota);
 
     return (
       !isEmpty(quota) && (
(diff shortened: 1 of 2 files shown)


tiborsimko (commented):
Inline review of the quota period window display. Two items, grouped by severity (HIGH / MEDIUM).

reana-ui/src/pages/profile/components/Quota.js
```diff
@@ -19,6 +20,22 @@ import { healthMapping } from "~/util";
...
 import styles from "./Quota.module.scss";
 
+export function getQuotaPeriodWindow(quota = {}) {
+  const {
+    quota_period_months: periodMonths,
+    quota_period_start_at: periodStartAt,
+  } = quota;
+  if (!periodMonths || !periodStartAt) return null;
+
+  const periodStart = moment(periodStartAt);
```
tiborsimko: 🤖 **Use `moment.utc(...)` for backend timestamps (HIGH)**

`quota_period_start_at` is serialised by reana-server as a naive (no-tz suffix) ISO string from a UTC-stored `datetime` (e.g. `"2026-04-01T13:06:32.992595"`). Plain `moment(periodStartAt)` parses that string in the *browser's* local timezone, so a user in PDT viewing a value the backend recorded as `2026-04-01T00:30:00 UTC` will see it as `Mar 31, 2026` in the UI — and the `endLabel` will be off by the same amount, since it's derived by `add(periodMonths, "months")` from the same misinterpreted moment.

The rest of the codebase consistently uses `moment.utc(...)` for backend-provided timestamps — see `util.js:150,152,165-168`, `parseWorkflowDates`, `getDuration`. The new code should follow the same convention.

Suggested fix:

```js
const periodStart = moment.utc(periodStartAt);
if (!periodStart.isValid()) return null;

return {
  startLabel: periodStart.format("ll"),
  endLabel: periodStart.clone().add(periodMonths, "months").format("ll"),
};
```

reana-ui/src/pages/profile/components/Quota.js
```diff
@@ -19,6 +20,22 @@ import { healthMapping } from "~/util";
...
+  if (!periodStart.isValid()) return null;
+
+  return {
+    startLabel: periodStart.format("ll"),
+    endLabel: periodStart.clone().add(periodMonths, "months").format("ll"),
+  };
+}
```
tiborsimko, line 37: 🤖 **`getQuotaPeriodWindow` is exported but has no unit test (MEDIUM)**

The function is a pure data transform — exactly the kind of thing this codebase tests in `src/util.test.js` with `test.each`. Without coverage we have no regression guard for the null-input branches, the invalid-date branch, or the month-boundary clamping (Jan 31 + 1 month → Feb 28). The reana-db PR has the same arithmetic on the backend; pinning the frontend's behaviour locks the contract.

Suggested fix — colocate a `Quota.test.js` (or extend `util.test.js`) with cases like:

```js
import { getQuotaPeriodWindow } from "./Quota";

test.each([
  [{}, null],
  [{ quota_period_months: 3 }, null],
  [{ quota_period_start_at: "2026-04-01T00:00:00" }, null],
  [{ quota_period_months: 3, quota_period_start_at: "not-a-date" }, null],
  [
    { quota_period_months: 3, quota_period_start_at: "2026-01-31T00:00:00" },
    { startLabel: "Jan 31, 2026", endLabel: "Apr 30, 2026" },
  ],
])("getQuotaPeriodWindow(%p) === %p", (quota, expected) => {
  expect(getQuotaPeriodWindow(quota)).toEqual(expected);
});
```

michaelbuchar: Integration tests done locally, maybe we can add this in a separate PR?

reana-ui/src/pages/profile/components/Quota.js
michaelbuchar: Nice catch, fixed!

tiborsimko (approved):

PR #496, opened 2026-04-21
fix(dask): hide empty service log component selector

### LINKED ISSUE #493: service-logs: Dask workflow shows no service logs and empty component selector
### Current behaviour
When running a Dask workflow on a REANA deployment with Dask enabled, the workflow finishes successfully, but the Service logs tab is empty. The page also currently renders a component dropdown even though there are no components to choose from.

Image

### Expected behaviour

For a Dask workflow, one of these should happen:

1. the Service logs tab should show the available Dask service logs or
2. if there are genuinely no service log components to show, the Component selector should be hidden/removed and the empty state should explain that no service components were created for this workflow.

### How to reproduce:

I used a local dev REANA deployment with Dask enabled and the Dask operator installed.

I ran the Dask demo workflow from:

```console
/src/reana-demo-dask-coffea
````

using:

```console
source reana-venv/bin/activate
reana-client run -w dask-test
```

The workflow completed successfully.

michaelbuchar:
Hide the Service Logs component selector when no Dask service log components are available, redirect stale component URLs to the base service logs view and show clearer empty state messages.

Closes #493

Screenshot 2026-04-21 at 1 27 34 PM
Screenshot 2026-04-21 at 1 27 40 PM
Screenshot 2026-04-21 at 1 27 48 PM

### DIFF
--- a/reana-ui/src/pages/workflowDetails/components/WorkflowLogs.js
+++ b/reana-ui/src/pages/workflowDetails/components/WorkflowLogs.js
@@ -50,16 +50,21 @@ EngineLogs.propTypes = {
 // UI design should be changed to allow for multiple services to be displayed at once if another service is introduced.
 function ServiceLogs({ isDask, components, workflowStatus }) {
   const isExecuting = NON_FINISHED_STATUSES.includes(workflowStatus);
+  const hasComponents = Array.isArray(components) && components.length > 0;
   // start empty to avoid reading components[0] before data arrives
   const [selectedComponentIndex, setSelectedComponentIndex] = useState("");
   const { id: workflowId, job: componentFromPath } = useParams();
   const navigate = useNavigate();
+  const serviceLogsPath = `/workflows/${workflowId}/service-logs`;
 
   // Keep dropdown selection and URL in sync (must run before any early return to satisfy hooks rule)
   useEffect(() => {
     if (!isDask || isExecuting) return;
-    if (!Array.isArray(components) || components.length === 0) {
+    if (!hasComponents) {
       if (selectedComponentIndex !== "") setSelectedComponentIndex("");
+      if (componentFromPath) {
+        navigate(serviceLogsPath, { replace: true });
+      }
       return;
     }
     if (componentFromPath) {
@@ -75,7 +80,7 @@ function ServiceLogs({ isDask, components, workflowStatus }) {
     }
     // Standardize URL to first component when missing/invalid
     navigate(
-      `/workflows/${workflowId}/service-logs/${encodeURIComponent(components[0].component)}`,
+      `${serviceLogsPath}/${encodeURIComponent(components[0].component)}`,
       { replace: true },
     );
   }, [
@@ -83,7 +88,7 @@ function ServiceLogs({ isDask, components, workflowStatus }) {
     isExecuting,
     components,
     componentFromPath,
-    workflowId,
+    serviceLogsPath,
     navigate,
     selectedComponentIndex,
   ]);
@@ -110,15 +115,25 @@ function ServiceLogs({ isDask, components, workflowStatus }) {
     );
   }
 
-  const dropdownOptions = Object.entries(components).map(([id, component]) => ({
-    key: id,
+  if (!hasComponents) {
+    return (
+      <Message
+        icon="info circle"
+        content={"No service logs were captured for this workflow."}
+        info
+      />
+    );
+  }
+
+  const dropdownOptions = components.map((component, index) => ({
+    key: String(index),
     text: component.component,
     icon: {
       name: "dot circle outline",
       size: "small",
       color: "green",
     },
-    value: id,
+    value: String(index),
   }));
 
   return (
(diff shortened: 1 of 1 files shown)


reana-ui/src/pages/workflowDetails/components/WorkflowLogs.js
```diff
@@ -110,15 +116,27 @@ function ServiceLogs({ isDask, components, workflowStatus }) {
...
+        content={
+          "No service components were created for this workflow, so there are no service logs to display."
+        }
+        info
+      />
+    );
+  }
+
+  const dropdownOptions = components.map((component, index) => ({
+    key: component.component || index,
```
tiborsimko: This can produce duplicate dropdown option keys when several service-log entries have the same `component` value. That is a realistic case for recreated Dask scheduler pods, where multiple entries can be `component: "scheduler"`. Could we keep the key unique, e.g. `key: String(index)`, or use `pod_uid`/`pod` if available?

reana-ui/src/pages/workflowDetails/components/WorkflowLogs.js
```diff
@@ -110,15 +116,27 @@ function ServiceLogs({ isDask, components, workflowStatus }) {
...
   }
 
-  const dropdownOptions = Object.entries(components).map(([id, component]) => ({
-    key: id,
+  if (!hasComponents) {
+    return (
+      <Message
+        icon="info circle"
+        content={
+          "No service components were created for this workflow, so there are no service logs to display."
```
tiborsimko: Minor wording point: this condition means no service-log components were returned, not necessarily that no service components were created. A Dask service can exist while no logs were persisted yet. Could this say something like “No service logs were captured for this workflow.”?

reana-ui/src/pages/workflowDetails/components/WorkflowLogs.js
```diff
@@ -75,15 +80,16 @@ function ServiceLogs({ isDask, components, workflowStatus }) {
...
     // Standardize URL to first component when missing/invalid
     navigate(
-      `/workflows/${workflowId}/service-logs/${encodeURIComponent(components[0].component)}`,
+      `${serviceLogsPath}/${encodeURIComponent(components[0].component)}`,
       { replace: true },
     );
   }, [
     isDask,
     isExecuting,
+    hasComponents,
```
tiborsimko: 🤖 **Drop redundant `hasComponents` from the effect dep array**

`hasComponents` is fully derived from `components` (`Array.isArray(components) && components.length > 0`). When `components` changes identity, the effect already re-runs and the derived boolean is re-evaluated. Listing both bloats the dependency list without changing behaviour, and obscures which inputs the effect actually depends on.

Suggestion: drop `hasComponents` from the dep array and rely on `components` to cover it. eslint's `react-hooks/exhaustive-deps` rule is satisfied as long as the source dep is tracked.

```js
}, [
  isDask,
  isExecuting,
  components,            // covers hasComponents
  componentFromPath,
  serviceLogsPath,
  navigate,
  selectedComponentIndex,
]);
```

reana-ui/src/pages/workflowDetails/components/WorkflowLogs.js
```diff
@@ -110,15 +116,27 @@ function ServiceLogs({ isDask, components, workflowStatus }) {
...
+          "No service components were created for this workflow, so there are no service logs to display."
+        }
+        info
+      />
+    );
+  }
+
+  const dropdownOptions = components.map((component, index) => ({
+    key: component.component || index,
     text: component.component,
```
tiborsimko, line 130: 🤖 **Dropdown text renders empty for components without a name**

`text: component.component` will render an empty dropdown row if `component.component` is `null` or `undefined`. Looking at `extract_component_name_from_pod_name` on the operator side, `None` is a possible return value for pod names that match neither "worker" nor "scheduler". That's pathological in practice (the operator only adds finalizers to those two label selectors), but if it ever happens the UI would silently show a blank entry rather than something the user can identify.

Suggestion: fall back to a safe placeholder so the row is at least readable.

```js
text: component.component || "(unnamed)",
```

reana-ui/src/pages/workflowDetails/components/WorkflowLogs.js
michaelbuchar: Fixed!

reana-ui/src/pages/workflowDetails/components/WorkflowLogs.js
michaelbuchar: Fixed

reana-ui/src/pages/workflowDetails/components/WorkflowLogs.js
michaelbuchar: Fixed

reana-ui/src/pages/workflowDetails/components/WorkflowLogs.js
michaelbuchar: I don't think this can happen in normal current Dask flows, because the operator only processes scheduler or worker pods and those names should always have a component name. So this looks more like defensive hardening than a fix for a realistic current bug. Should I still address it?

tiborsimko: Given the upstream operator guarantees a name for the only two pod kinds it manages, leaving the code as-is is reasonable.

tiborsimko (approved):

PR #499, opened 2026-05-04
feat(workflow-details): reorder log tabs

tiborsimko:
Reorder workflow log tabs from Engine -> Job -> Service to Engine -> Service -> Job to match the scope hierarchy: engine logs cover the whole workflow, service logs cover the (cross- step) Dask cluster, and job logs are per-step. The new order also matches the reana-client log render order, keeping CLI and Web UI consistent.

Harden the default-tab lookup so it derives the Engine and Job indices via tabKeys.indexOf() rather than hardcoding 0 and 1, which would otherwise silently break for Dask workflows after the Service logs pane changes their numerical position.


michaelbuchar (approved):
Works nicely on my end!

PR #506, opened 2026-06-04
feat(workflow-details): add shareable links for workspace files

Sohn123:
To allow users to share specific files of a workspace we introduce the possibility of creating and copying a sharelinkfor specific files in a workspace. Opening this link directly navigates to the preview of the file.

The links follow the following pattern:
```
/workflows/<wokflow_id>/workspace?name=<file_path>
http://localhost:3000/workflows/3566dfe0-e02d-4030-b4d5-3ed43a6f3306/workspace?name=inputs.json
```

I added the copy link action as a quick action to the right of the file overview:
image

Also opening the file preview navigates to the link so copying the link from the url bar is possible.

### DIFF
--- a/reana-ui/src/components/CopyButton.js
+++ b/reana-ui/src/components/CopyButton.js
@@ -1,0 +1,64 @@
+/*
+  -*- coding: utf-8 -*-
+
+  This file is part of REANA.
+  Copyright (C) 2026 CERN.
+
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+import PropTypes from "prop-types";
+import { useEffect, useRef, useState } from "react";
+import { Button, Icon, Popup } from "semantic-ui-react";
+import copy from "copy-to-clipboard";
+
+const COPY_CHECK_TIMEOUT = 1000;
+
+export default function CopyButton({
+  text,
+  as: Trigger = Button,
+  label,
+  icon = "copy outline",
+  timeout = COPY_CHECK_TIMEOUT,
+  ...props
+}) {
+  const [copied, setCopied] = useState(false);
+  const timeoutRef = useRef();
+
+  useEffect(() => () => clearTimeout(timeoutRef.current), []);
+
+  const handleClick = async () => {
+    const ok = await copy(text, { format: "text/plain" });
+    if (!ok) return;
+    setCopied(true);
+    clearTimeout(timeoutRef.current);
+    timeoutRef.current = setTimeout(() => setCopied(false), timeout);
+  };
+  const triggerProps =
+    Trigger === Icon ? { name: icon } : { icon, content: label };
+
+  return (
+    <Popup
+      inverted
+      trigger={
+        <Trigger
+          {...triggerProps}
+          {...props}
+          aria-label={props["aria-label"] || "Copy to clipboard"}
+          onClick={handleClick}
+        />
+      }
+      open={copied}
+      content="Copied to clipboard!"
+    />
+  );
+}
+
+CopyButton.propTypes = {
+  text: PropTypes.string.isRequired,
+  as: PropTypes.elementType,
+  label: PropTypes.string,
+  icon: PropTypes.string,
+  timeout: PropTypes.number,
+};
--- a/reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
+++ b/reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
@@ -11,12 +11,20 @@
 import sortBy from "lodash/sortBy";
 import PropTypes from "prop-types";
 import { useEffect, useState, useMemo } from "react";
-import { useSearchParams } from "react-router-dom";
+import { useSearchParams, useNavigate, useLocation } from "react-router-dom";
 import { useDispatch, useSelector } from "react-redux";
-import { Icon, Loader, Message, Segment, Table } from "semantic-ui-react";
+import {
+  Button,
+  Icon,
+  Loader,
+  Message,
+  Segment,
+  Table,
+} from "semantic-ui-react";
 
 import { fetchWorkflowFiles } from "~/actions";
-import { Pagination, Search } from "~/components";
+import { WORKFLOW_FILE_URL } from "~/client";
+import { Pagination, Search, CopyButton } from "~/components";
 import {
   getWorkflowFiles,
   getWorkflowFilesCount,
@@ -30,12 +38,30 @@ import styles from "./WorkflowFiles.module.scss";
 
 const PAGE_SIZE = 15;
 
+function getDownloadURL(workflow, fileName) {
+  return WORKFLOW_FILE_URL(workflow, fileName, { preview: false });
+}
+
+function openPreviewParams(prev, name) {
+  const next = new URLSearchParams(prev);
+  next.set("name", name);
+  return next;
+}
+
+function closePreviewParams(prev) {
+  const next = new URLSearchParams(prev);
+  next.delete("name");
+  return next;
+}
+
 export default function WorkflowFiles({ id, page = 1, onPageChange }) {
   const dispatch = useDispatch();
   const workflowRefresh = useSelector(getWorkflowRefresh);
   const loading = useSelector(loadingDetails);
   const _files = useSelector(getWorkflowFiles(id));
   const filesCount = useSelector(getWorkflowFilesCount(id));
+  const navigate = useNavigate();
+  const location = useLocation();
 
   const [files, setFiles] = useState();
   const [sorting, setSorting] = useState({ column: null, direction: null });
@@ -52,11 +78,18 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
     [page],
   );
 
-  // Keep input in sync when URL changes
+  // Sync filePreview state from URL params
   useEffect(() => {
     const urlSearch = searchParams.get("search") || "";
+    const fileName = searchParams.get("name") || "";
+
+    if (fileName) {
+      setFilePreview({ workflow: id, fileName });
+    } else {
+      setFilePreview(null);
+    }
     setSearchText((prev) => (prev !== urlSearch ? urlSearch : prev));
-  }, [searchParams]);
+  }, [searchParams, id]);
 
   useEffect(() => {
     dispatch(fetchWorkflowFiles(id, pagination, searchQuery));
@@ -75,7 +108,7 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
       setSorting({ direction: "ascending", column: clickedColumn });
       return;
     }
-    setFiles(files.reverse());
+    setFiles([...files].reverse());
     setSorting({
       ...sorting,
       direction: sorting.direction === "ascending" ? "descending" : "ascending",
@@ -98,6 +131,21 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
     />
   );
 
+  const openPreview = (name, fileSize) => {
+    setSearchParams((prev) => openPreviewParams(prev, name), {
+      replace: false,
+      state: { internal: true, size: fileSize },
+    });
+  };
+
+  const closePreview = () => {
+    if (location.state?.internal) {
+      navigate(-1);
+    } else {
+      setSearchParams((prev) => closePreviewParams(prev), { replace: true });
+    }
+  };
+
   // Submit search on Enter/click, then reset to page 1 (remove ?page)
   const submitSearch = () => {
     const q = searchText.trim();
@@ -114,6 +162,9 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
     );
     resetSort();
   };
+
+  const buildShareLink = (name) =>
+    `${window.location.origin}/workflows/${id}/workspace?name=${encodeURIComponent(name)}`;
 
   return files === null ? (
     <Message
@@ -151,6 +202,7 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
                   Modified {headerIcon("lastModified")}
                 </Table.HeaderCell>
                 <Table.HeaderCell
+                  width={2}
                   sorted={
                     sorting.column === "size.raw" ? sorting.direction : null
                   }
@@ -158,6 +210,10 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
                 >
                   Size {headerIcon("size.raw")}
                 </Table.HeaderCell>
+                <Table.HeaderCell
+                  width={3}
+                  textAlign="center"
+                ></Table.HeaderCell>
               </Table.Row>
             </Table.Header>
 
@@ -167,13 +223,7 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
                   <Table.Row
                     key={name}
                     className={styles["files-row"]}
-                    onClick={() =>
-                      setFilePreview({
-                        workflow: id,
-                        fileName: name,
-                        size: size.raw,
-                      })
-                    }
+                    onClick={() => openPreview(name, size.raw)}
                   >
                     <Table.Cell>
                       <Icon name="file" />
@@ -181,16 +231,37 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
                     </Table.Cell>
                     <Table.Cell>{lastModified}</Table.Cell>
                     <Table.Cell>{size.human_readable || size.raw}</Table.Cell>
+                    <Table.Cell
+                      className={styles["copy-link-cell"]}
+                      onClick={(e) => e.stopPropagation()}
+                    >
+                      <Button
+                        icon="eye"
+                        size="mini"
+                        aria-label={`Preview ${name}`}
+                        onClick={() => openPreview(name, size.raw)}
+                      />
+                      <Button
+                        size="mini"
+                        icon="download"
+                        as="a"
+                        href={getDownloadURL(id, name)}
+                        aria-label={`Download ${name}`}
+                      />
+                      <CopyButton
+                        text={buildShareLink(name)}
+                        size="mini"
+                        icon="linkify"
+                        aria-label={`Copy shareable link to ${name}`}
+                      />
+                    </Table.Cell>
                   </Table.Row>
                 ))}
             </Table.Body>
           </Table>
 
           {filePreview && (
-            <FilePreview
-              {...filePreview}
-              onClose={() => setFilePreview(null)}
-            />
+            <FilePreview {...filePreview} onClose={closePreview} />
           )}
 
           {filesCount > PAGE_SIZE && (
--- a/reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
+++ b/reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
@@ -11,12 +11,14 @@
 import sortBy from "lodash/sortBy";
 import PropTypes from "prop-types";
 import { Suspense, lazy, useEffect, useState } from "react";
+import { useLocation, useNavigationType } from "react-router-dom";
 import { useSelector } from "react-redux";
 import { Button, Icon, Loader, Message, Modal } from "semantic-ui-react";
 
 import client, { WORKFLOW_FILE_URL } from "~/client";
 import { getConfig } from "~/selectors";
-import { formatFileSize, getMimeType } from "~/util";
+import { formatFileSize, getMimeType, parseFiles } from "~/util";
+import { CopyButton } from "~/components";
 
 import styles from "./FilePreview.module.scss";
 
@@ -167,33 +169,115 @@ function getFileURL(workflow, fileName, preview = true) {
   return WORKFLOW_FILE_URL(workflow, fileName, { preview });
 }
 
-export default function FilePreview({ workflow, fileName, size, onClose }) {
+export default function FilePreview({ workflow, fileName, onClose }) {
   const config = useSelector(getConfig);
+  const location = useLocation();
+  const isSharedLink = !location.state?.internal;
+  const [size, setSize] = useState(null);
+  const [error, setError] = useState(null);
+  const shareUrl = `${window.location.origin}/workflows/${workflow}/workspace?name=${encodeURIComponent(fileName)}`;
+  const navigationType = useNavigationType();
+
+  useEffect(() => {
+    const stateSize = location.state?.size;
+    if (stateSize !== undefined && navigationType === "PUSH") {
+      setSize(stateSize);
+      return;
+    }
+    setSize(null);
+    setError(null);
+    client
+      .getWorkflowFiles(
+        workflow,
+        { page: 1, size: 1, file_name: fileName },
+        JSON.stringify({ name: [fileName] }),
+      )
+      .then((resp) => {
+        const items = parseFiles(resp.data.items ?? []);
+        const match = items.find((f) => f.name === fileName);
+        if (match === undefined) {
+          throw new Error("File not found");
+        }
+        setSize(match?.size?.raw ?? 0);
+      })
+      .catch((err) => {
+        setError(err);
+        setSize(0);
+      });
+  }, [workflow, fileName, location.state, navigationType]);
 
   let preview = null;
-  // Check whether the file can be previewed
-  const message = checkConstraints(fileName, size, config.filePreviewSizeLimit);
-  if (message) {
-    preview = <Modal.Content scrolling>{message}</Modal.Content>;
-  } else {
-    const mimeType = matchesMimeType(
-      Object.keys(PREVIEW_MIME_PREFIX_WHITELIST),
-      fileName,
-    );
-    const Preview = PREVIEW_MIME_PREFIX_WHITELIST[mimeType];
+
+  if (error && isSharedLink) {
     preview = (
-      <Preview
-        workflow={workflow}
-        fileName={fileName}
-        size={size}
-        url={getFileURL(workflow, fileName)}
-      />
+      <Modal.Content>
+        <Message
+          icon="exclamation triangle"
+          content="This file cannot be previewed. It may have been removed from this workspace or you may not have access to it anymore."
+          warning
+        />
+      </Modal.Content>
     );
+  } else if (error && !isSharedLink) {
+    preview = (
+      <Modal.Content>
+        <Message
+          icon="exclamation triangle"
+          content="An error occurred while loading the file preview. Please try again later or use the download button."
+          warning
+        />
+      </Modal.Content>
+    );
+  } else if (size === null) {
+    preview = (
+      <Modal.Content>
+        <Loader
+          active
+          className={styles["dark-loader"]}
+          inline="centered"
+          content="Loading file preview..."
+        />
+      </Modal.Content>
+    );
+  } else {
+    const message = checkConstraints(
+      fileName,
+      size,
+      config.filePreviewSizeLimit,
+    );
+    if (message) {
+      preview = <Modal.Content scrolling>{message}</Modal.Content>;
+    } else {
+      const mimeType = matchesMimeType(
+        Object.keys(PREVIEW_MIME_PREFIX_WHITELIST),
+        fileName,
+      );
+      const Preview = PREVIEW_MIME_PREFIX_WHITELIST[mimeType];
+      preview = (
+        <Preview
+          workflow={workflow}
+          fileName={fileName}
+          size={size}
+          url={getFileURL(workflow, fileName)}
+        />
+      );
+    }
   }
 
   return (
     <Modal open closeIcon onClose={onClose ? onClose : () => {}}>
       <Modal.Header>{fileName}</Modal.Header>
+      {isSharedLink && !error && (
+        <Modal.Content>
+          <Modal.Description>
+            <Message
+              icon="info circle"
+              content="Workspace files can change over time — you're viewing the current version of this file."
+              info
+            />
+          </Modal.Description>
+        </Modal.Content>
+      )}
       <Suspense
         fallback={
           <Modal.Content>
@@ -208,11 +292,14 @@ export default function FilePreview({ workflow, fileName, size, onClose }) {
       >
         {preview}
       </Suspense>
-      <Modal.Actions>
-        <Button primary as="a" href={getFileURL(workflow, fileName, false)}>
-          <Icon name="download" /> Download
-        </Button>
-      </Modal.Actions>
+      {!error && (
+        <Modal.Actions>
+          <CopyButton text={shareUrl} label="Copy link" icon="linkify" />
+          <Button primary as="a" href={getFileURL(workflow, fileName, false)}>
+            <Icon name="download" /> Download
+          </Button>
+        </Modal.Actions>
+      )}
     </Modal>
   );
 }
@@ -220,6 +307,5 @@ export default function FilePreview({ workflow, fileName, size, onClose }) {
 FilePreview.propTypes = {
   workflow: PropTypes.string.isRequired,
   fileName: PropTypes.string.isRequired,
-  size: PropTypes.number.isRequired,
   onClose: PropTypes.func,
 };
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -4,6 +4,7 @@ The list of contributors in alphabetical order:
 
 - [Alastair Lyall](https://orcid.org/0009-0000-4955-8935)
 - [Alp Tuna](https://orcid.org/0009-0001-1915-3993)
+- [Alexander Sohn](https://orcid.org/0009-0000-3673-4637)
 - [Audrius Mecionis](https://orcid.org/0000-0002-3759-1663)
 - [Cameron McClymont](https://orcid.org/0009-0002-0176-5251)
 - [Daan Rosendal](https://orcid.org/0000-0002-3447-9000)
--- a/reana-ui/package.json
+++ b/reana-ui/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "dependencies": {
     "axios": "^1.13.2",
+    "copy-to-clipboard": "^4.0.2",
     "dompurify": "^3.3.1",
     "jsroot": "~7.5.0",
     "lodash": "^4.17.23",
@@ -12,7 +13,6 @@
     "prop-types": "^15.8.1",
     "query-string": "^8.1.0",
     "react": "^18.3.0",
-    "react-copy-to-clipboard": "^5.0.2",
     "react-dom": "^18.3.0",
     "react-minimal-pie-chart": "^8.0.1",
     "react-redux": "^8.1.3",
--- a/reana-ui/src/components/CodeSnippet.js
+++ b/reana-ui/src/components/CodeSnippet.js
@@ -9,13 +9,11 @@
 */
 
 import { useState } from "react";
-import { Icon, Popup } from "semantic-ui-react";
-import { CopyToClipboard } from "react-copy-to-clipboard";
+import { Icon } from "semantic-ui-react";
 import PropTypes from "prop-types";
 
+import CopyButton from "./CopyButton";
 import styles from "./CodeSnippet.module.scss";
-
-const COPY_CHECK_TIMEOUT = 1500;
 
 export default function CodeSnippet({
   children,
@@ -26,17 +24,7 @@ export default function CodeSnippet({
   dollarPrefix,
   classes,
 }) {
-  const [copied, setCopied] = useState(false);
   const [revealed, setRevealed] = useState(false);
-
-  const handleCopied = () => {
-    setCopied(true);
-
-    const timeout = setTimeout(() => {
-      setCopied(false);
-      clearTimeout(timeout);
-    }, COPY_CHECK_TIMEOUT);
-  };
 
   const toggleRevealed = () => {
     setRevealed(!revealed);
@@ -65,31 +53,27 @@ export default function CodeSnippet({
       >
         {children}
       </div>
-      {reveal && (
-        <Icon
-          name={revealed ? "eye slash" : "eye"}
-          className={styles["action-icon"]}
-          onClick={toggleRevealed}
-        />
-      )}
-      {copy && (
-        <Popup
-          trigger={
-            <CopyToClipboard
+      {(reveal || copy) && (
+        <div className={styles["actions"]}>
+          {reveal && (
+            <Icon
+              name={revealed ? "eye slash" : "eye"}
+              className={styles["action-icon"]}
+              onClick={toggleRevealed}
+            />
+          )}
+          {copy && (
+            <CopyButton
+              as={Icon}
+              className={styles["action-icon"]}
               text={accessChildren(children)
                 .map((line) => {
                   return Array.isArray(line) ? line.join("") : line;
                 })
                 .join("\n")}
-              onCopy={handleCopied}
-            >
-              <Icon name="copy outline" className={styles["action-icon"]} />
-            </CopyToClipboard>
-          }
-          content="Copied!"
-          open={copied}
-          inverted
-        />
+            />
+          )}
+        </div>
       )}
     </div>
   );
--- a/reana-ui/src/components/CodeSnippet.module.scss
+++ b/reana-ui/src/components/CodeSnippet.module.scss
@@ -59,4 +59,9 @@
       opacity: 1;
     }
   }
+
+  .actions {
+    display: flex;
+    gap: 0.5em;
+  }
 }
--- a/reana-ui/src/components/__tests__/CopyButton.test.js
+++ b/reana-ui/src/components/__tests__/CopyButton.test.js
@@ -1,0 +1,51 @@
+/*
+  This file is part of REANA.
+  Copyright (C) 2026 CERN.
+
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+import { fireEvent, render, screen } from "@testing-library/react";
+import copy from "copy-to-clipboard";
+import { Icon } from "semantic-ui-react";
+
+import { CopyButton } from "..";
+
+jest.mock("copy-to-clipboard");
+
+describe("CopyButton", () => {
+  beforeEach(() => {
+    copy.mockResolvedValue(true);
+  });
+
+  test("copies text from a button", async () => {
+    render(<CopyButton text="share URL" label="Copy link" />);
+
+    fireEvent.click(screen.getByRole("button", { name: "Copy to clipboard" }));
+
+    expect(copy).toHaveBeenCalledWith(
+      "share URL",
+      expect.objectContaining({ format: "text/plain" }),
+    );
+    expect(await screen.findByText("Copied to clipboard!")).toBeInTheDocument();
+  });
+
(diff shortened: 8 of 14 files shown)


reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
```diff
@@ -181,6 +235,65 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
...
+                                  },
+                                },
+                              )
+                            }
+                          />
+                        }
+                        open={copied === name}
+                      >
+                        <Popup.Content>
+                          <p>Link copied to clipboard!</p>
```
tiborsimko: Thanks for the draft! The links work well. 

Some high-level observations from testing the PR behaviour:

- In the future, it may be useful to have a "delete" icon for files. So you can think about that from the design point of view.
- (Also, "Prune workspace" may now be a bit hidden if people try to find a way how to delete all the files and not do it one-by-one.)
- The version argument in permalinks to files feels a bit artificial, because we don't have a way to revert to previous versions of a file if a user overwrites them. Basically, the permlink should be considered as a pointer to the file name, whatever version the workspace had. I would drop it from the permalink so as not to raise some expectations that we cannot fulfill, and just print a warning for people if they come from a link that the files may have changed in the meantime.
- If somebody clicks on preview, they can visualise a file and they see a button "Download" there. It may be perhaps useful to have also the "Permanent link" button there too, for user convenience.

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
```diff
@@ -55,8 +86,29 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
...
     const urlSearch = searchParams.get("search") || "";
+    const fileName = searchParams.get("name") || "";
+    const version = searchParams.get("version") || "";
+
+    // Check if the version matches the last modified date of the file to ensure we are sharing the correct version
+    if (fileName && version && files) {
+      let size = null;
+      for (const file in files) {
+        if (files[file].name === fileName) {
+          size = files[file].size;
```
tiborsimko: **🤖 `FilePreview` is given a size object instead of a number, bypassing the size limit on every open (HIGH)**

`reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js:97`

```js
for (const file in files) {
  if (files[file].name === fileName) {
    size = files[file].size;          // ← {raw, human_readable} object
    setOutdated(files[file].lastModified !== version);
    break;
  }
}
setFilePreview({ workflow: id, fileName, size: size });
```

The original code passed `size.raw` (a number). The new code passes the whole `size` object. Two consequences:

- `FilePreview.propTypes.size` is `PropTypes.number.isRequired` — propTypes warning on every open.
- `checkConstraints` does `size > sizeLimit`, which is always `false` for objects, so **the configured preview size limit is silently bypassed for every link-driven open**, even on the happy path where the file is present on page 1.

Note this is distinct from the separate off-page failure where `size` ends up `null` because the named file isn't in the current page's slice — that one fires only on deep-link opens, this one fires on *every* open. Both need fixing.

Suggested fix —

```js
size = files[file].size.raw;
```

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
```diff
@@ -55,8 +86,29 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
   // Keep input in sync when URL changes
   useEffect(() => {
     const urlSearch = searchParams.get("search") || "";
+    const fileName = searchParams.get("name") || "";
+    const version = searchParams.get("version") || "";
+
+    // Check if the version matches the last modified date of the file to ensure we are sharing the correct version
+    if (fileName && version && files) {
```
tiborsimko: **🤖 Shared links to files outside the current page open a half-initialised preview (HIGH)**

`WorkflowFiles.js:93-109`

The effect only inspects `files`, which is the *current page's* slice (and the copied link at `:265` omits `page` and `search`, so a recipient always starts on page 1). If the shared file isn't on the recipient's current page, the loop never matches, so:

- `size` stays `null`. `null > sizeLimit` is `false`, so the preview-size limit is bypassed.
- `setOutdated` is never called from this branch, so the "version mismatch" warning reflects whatever stale value was left over from a previous open — the very banner the feature added to make shared links trustworthy is unreliable on the exact case it's there for.
- The preview still opens, against a record the component never actually located.

For a shareable-link feature this is the central failure case.

Suggested fix — resolve the file's metadata explicitly before opening the modal rather than relying on the active page's slice. Either:

- fetch the file directly by name (e.g. `fetchWorkflowFiles` with `search=<exact name>` and `size=1`) and read `size.raw` + `lastModified` from the response; or
- have the copied link include `page` (and `search`) so the recipient lands on the right page, with a fallback fetch when the file still isn't there (e.g. it moved pages between share and open).

While reworking the effect, two small follow-ons worth folding in:

1. Reset `setOutdated(false)` at the top of the effect so its value can't leak across opens (independently of the off-page case above — e.g. if the user opens a fresh link after dismissing a previous "outdated" modal, the banner shouldn't carry over).
2. Replace the `for…in` loop with `files.find((f) => f.name === fileName)` — `for…in` iterates enumerable string keys (including any inherited from `Array.prototype` pollution) and is semantically wrong for arrays.

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
```diff
@@ -168,11 +224,9 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
                     key={name}
                     className={styles["files-row"]}
                     onClick={() =>
-                      setFilePreview({
-                        workflow: id,
-                        fileName: name,
-                        size: size.raw,
-                      })
+                      navigate(
+                        `/workflows/${id}/workspace/files?page=${pagination.page}&name=${encodeURI(name)}&version=${encodeURI(lastModified)}`,
```
tiborsimko: **🤖 `encodeURI` is the wrong function for query-param values (HIGH)**

`WorkflowFiles.js:228, 247, 265, 273, 279`

```js
`…/workspace/files?page=${pagination.page}&name=${encodeURI(name)}&version=${encodeURI(lastModified)}`
```

`encodeURI` does **not** escape `?`, `&`, `=`, `#`. A filename containing any of those characters (perfectly legal on disk) produces a broken URL: e.g. `a&b.txt` becomes `…?name=a&b.txt&version=…`, so `name` parses as `"a"` and `b.txt` becomes an extra query parameter. Embedded `#` truncates everything after it into a fragment.

Suggested fix — use `encodeURIComponent` (5 occurrences, including the HTML anchor href constructed inside the `onCopy` callback).

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
```diff
@@ -181,6 +235,65 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
...
+                        as="a"
+                        href={getDownloadURL(id, name)}
+                      />
+                      <Popup
+                        inverted
+                        trigger={
+                          <Button
+                            size="mini"
+                            icon="linkify"
+                            onClick={() =>
```
tiborsimko: **🤖 Rich HTML clipboard copy is ineffective and can break plain-text copy (MEDIUM)**

`WorkflowFiles.js:263-286`

```js
copy(`${origin}/…?name=…&version=…`, {
  format: "text/plain",
  onCopy: () => {
    handleCopied(name);
    return new ClipboardItem({ "text/html": new Blob([...]), "text/plain": ... });
  },
});
```

Two related issues:

1. `copy-to-clipboard`'s `onCopy` callback is invoked with a `DataTransfer`/`ClipboardEvent` object and is meant for `e.clipboardData.setData("text/html", …)`. The return value is ignored, so the `ClipboardItem` constructed here is silently discarded — only the plain-text payload (which `copy()` already handles) ends up on the clipboard. The rich-HTML anchor never reaches it.
2. Independent of that, `new ClipboardItem(...)` itself is undefined in browsers without it (older Firefox, any non-secure context), so the callback would throw `ReferenceError`. Because the throw happens inside `onCopy`, an exception bubbles out of `copy()` and breaks the plain-text copy too.

So if rich HTML wasn't a deliberate goal of this PR, the simplest fix is to drop the whole `onCopy` block and keep just `copy(url)` — the plain link copy already works and is what users typically expect from a "Copy link" button.

If rich HTML *is* desired, replace this with `navigator.clipboard.write([new ClipboardItem({…})])` directly, guarded by `typeof ClipboardItem !== "undefined" && navigator.clipboard?.write` and wrapped in a try/catch that falls back to plain text.

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
```diff
@@ -28,14 +38,35 @@ import FilePreview from "./filePreview/FilePreview";
 
 import styles from "./WorkflowFiles.module.scss";
 
-const PAGE_SIZE = 15;
+const PAGE_SIZE = 5;
```
tiborsimko: **🤖 `PAGE_SIZE` change from 15 to 5 looks accidental (MEDIUM)**

`WorkflowFiles.js:41`

```js
-const PAGE_SIZE = 15;
+const PAGE_SIZE = 5;
```

This looks like a leftover from local testing of pagination interacting with shared links — it isn't mentioned in the PR description and triples the click count for everyone browsing workspace files. If the smaller page is intentional, worth calling out in the commit message; otherwise restore to 15.

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
tiborsimko: **🤖 Opening/closing the preview clobbers the workspace search filter (HIGH)**

`WorkflowFiles.js:228, 247` (open) vs `:307` (close)

Both the row click and the eye button do an absolute `navigate(...)` that rebuilds the URL from scratch with only `page`, `name`, `version`:

```js
navigate(`/workflows/${id}/workspace/files?page=${pagination.page}&name=${encodeURI(name)}&version=${encodeURI(lastModified)}`);
```

…and the close handler rebuilds it with only `page`:

```js
navigate(`/workflows/${id}/workspace?page=${pagination.page}`);
```

For a user who is on `?search=foo`, this means:

1. They filter to `?search=foo` and click a result.
2. `search` is dropped from the URL, so `searchQuery = (searchParams.get("search") || "").trim()` becomes `""`.
3. The `dispatch(fetchWorkflowFiles(id, pagination, searchQuery))` effect re-runs with no filter — a wasted unfiltered page fetch in the background while the preview is open.
4. Closing the modal navigates to `…/workspace?page=…`, still without `search`, so they land on the unfiltered list and lose their place.

This breaks the primary search → preview → back-to-search flow.

There's also a secondary URL-shape inconsistency: the open path includes a dead `/files` segment (the route is `/workflows/:id/:tab?/:job?` from `App.js:100`, so `files` parses as `:job` and is unused by `WorkflowDetails` for the workspace tab), while the close path drops it back to `/workflows/<id>/workspace`. The two endpoints of a single open/close round trip therefore use different paths, polluting browser history. Worse, `WorkflowDetails`'s tab-change handler appends `jobFromPath` when building the Job Logs URL — so a user who opens a file preview and then clicks the "Job logs" tab ends up at `/workflows/<id>/job-logs/files`, where `files` is treated as a `backend_job_id`.

Suggested fix — drive the preview entirely through `setSearchParams` so unrelated params (`search`, future additions) survive, and keep the path identical at open and close. Sketch:

```js
const openPreview = (name, lastModified) =>
  setSearchParams(
    (prev) => {
      const next = new URLSearchParams(prev);
      next.set("name", name);
      next.set("version", lastModified);
      return next;
    },
    { replace: false },
  );

const closePreview = () =>
  setSearchParams(
    (prev) => {
      const next = new URLSearchParams(prev);
      next.delete("name");
      next.delete("version");
      return next;
    },
    { replace: true },
  );
```

Drop the `/files` path segment entirely — `/workflows/<id>/workspace?name=…&version=…` is enough.

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
```diff
@@ -28,14 +38,35 @@ import FilePreview from "./filePreview/FilePreview";
...
+const COPY_CHECK_TIMEOUT = 1000;
+
+/**
+ * Build the download URL of a file.
+ */
+function getDownloadURL(workflow, fileName) {
```
tiborsimko, line 41: **🤖 New `getDownloadURL` skips the `preview=false` flag (MEDIUM)**

`WorkflowFiles.js:47-49`

```js
function getDownloadURL(workflow, fileName) {
  return WORKFLOW_FILE_URL(workflow, fileName);
}
```

Existing convention (`FilePreview.js:227`) builds the download URL via `getFileURL(workflow, fileName, false)`, producing `?preview=false`. The new helper omits the flag, so the URL has no `preview` param at all and relies on server defaults. (As an aside, `WORKFLOW_FILE_URL`'s declared signature `(id, filename, preview = true)` doesn't match how callers use it — they pass `{ preview }` as an object — but that pre-existing wart is out of scope here.)

Suggested fix — call the same wrapper that `FilePreview` uses, or pass `{ preview: false }` explicitly.

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
```diff
@@ -181,6 +235,65 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
                     </Table.Cell>
                     <Table.Cell>{lastModified}</Table.Cell>
                     <Table.Cell>{size.human_readable || size.raw}</Table.Cell>
+                    <Table.Cell
+                      className={styles["copy-link-cell"]}
+                      onClick={(e) => e.stopPropagation()}
+                    >
+                      <Button
```
tiborsimko, line 238: **🤖 Icon-only buttons have no accessible name (LOW)**

`WorkflowFiles.js:242-256, 260-288`

The eye/download/link buttons render as Semantic UI `<Button icon="…">` with no text and no `aria-label`. Screen readers announce nothing meaningful.

Suggested fix — add `aria-label="Preview"`, `aria-label="Download"`, `aria-label="Copy link"`.

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
```diff
@@ -194,6 +200,15 @@ export default function FilePreview({ workflow, fileName, size, onClose }) {
   return (
     <Modal open closeIcon onClose={onClose ? onClose : () => {}}>
       <Modal.Header>{fileName}</Modal.Header>
+      {outdated && (
```
tiborsimko: **🤖 Share URL's `version` parameter implies historical retrieval that REANA doesn't support (LOW)**

`WorkflowFiles.js:265, 273, 279` (copied link) and `FilePreview.js:203-211` (banner driven by it)

The copied permanent link looks like:

```
https://…/workflows/<id>/workspace/files?name=gendata/data.root&version=2026-06-03T14:15:52
```

But REANA's workspace API only serves the *current* bytes of a file — there's no endpoint that returns the file as it was at `version=2026-06-03T14:15:52`. The `version` query parameter therefore cannot select anything; its only consumer is the "this might be a newer version of the file" warning in `FilePreview`. That makes the wording of both the URL and the banner a bit misleading, because they suggest a "share this exact revision" capability that doesn't exist.

A few options, in increasing order of intrusiveness:

1. **Tweak the banner wording only.** Make the message a property of the system rather than a timestamp comparison: e.g. *"Workspace files can change over time — you're viewing the current version of this file."* No promises about historical bytes.
2. **Tweak wording + rename the param** (e.g. `as_of` or `t`) so the URL doesn't read like a revision selector.
3. **Drop the `version` param entirely** and show the passive note above whenever the modal was opened via a deep link (`searchParams.has("name")`). Removes the `outdated` state plumbing in `WorkflowFiles.js` and `FilePreview.js`.

Worth picking one as part of the same PR since the feature explicitly trades on the "shareable version" framing. If proper workspace-file history ever lands (S3 versioning, snapshots, etc.), a real `version` selector can be reintroduced and the banner can be earned.

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
Sohn123: - Should I add the delete icon already now to the actions or should I rather put this in a separate PR later on? I think also a fourth icon would still be ok, because I think there is enough horizontal space there, but we could think about hiding them behind a kebab/burger menu
- As a more general design change I propose we could display the hidden options always maybe something like this but in prettier:
image
Maybe having 2x2 grid layout for the buttons
- Ok your point is really valid. The conservative approach of always displaying the warning might tip off users, when the file actually is the same. We could also think about renaming the parameter into a name that doesn't give the impression, that this is a persistent version we saved. But I can't think of a great alternative right now.
- Yes adding a button there shouldn't be a problem and makes sense

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
Sohn123: Will fix this

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
Sohn123: Adding a guard for older browsers makes sense and I will add that. The other part that the return value is not respected is not true. I liked the idea of having the html copy feature, but if we decide we don't want it, I am happy to remove it. I personally would prefer it, but I can imagine some people rather like the raw link

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
Sohn123: Yes good catch!

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
Sohn123: Yes this sounds reasonable. I think search param clobbering is something I should also manually review again.

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
Sohn123: See comments above. I think `as_of` or `t` could be ok, but also dropping it completly is ok. I like this wording: Workspace files can change over time — you're viewing the current version of this file.

Sohn123:
I now addressed most of the comments. I now see the following open points where we need to take a decision:
- Should the share link include the date in some form, to warn the user if the file is newer or not?
- Should the share link have the page parameter included or not? Right now it does not, which means that the preview is shown, the user clicks away and the is one page one, where the file might not be.
- Do we want to use html copying or not and if so what should be displayed as the name of the link?

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
```diff
@@ -167,33 +169,84 @@ function getFileURL(workflow, fileName, preview = true) {
...
+    client
+      .getWorkflowFiles(
+        workflow,
+        { page: 1, size: 1 },
+        JSON.stringify({ name: [fileName] }),
+      )
+      .then((resp) => {
+        const items = parseFiles(resp.data.items ?? []);
+        const match = items.find((f) => f.name === fileName);
+        setSize(match?.size?.raw ?? 0);
```
michaelbuchar, line 201: Missing file and request failure cases are both being treated as size = 0 here. This means a shared link to a renamed/pruned file can still go through the normal preview flow as if it were an empty file (and give a 404 error), instead of telling the user the file no longer exists. Can we keep a separate not found / request failed state and show an explicit message for stale links?

reana-ui/src/pages/workflowDetails/components/CopyLinkButton.js
```diff
@@ -0,0 +1,55 @@
...
+import { Button, Popup } from "semantic-ui-react";
+import copy from "copy-to-clipboard";
+
+const COPY_CHECK_TIMEOUT = 1000;
+
+export default function CopyLinkButton({ url, fileName, size, label }) {
+  const [copied, setCopied] = useState(false);
+
+  const handleClick = () => {
+    copy(url, {
```
michaelbuchar: I think copy-to-clipboard is being used a bit differently than it expects here. The onCopy callback receives clipboardData to mutate, but the returned ClipboardItem is ignored, so the rich HTML link never actually reaches the clipboard. This might mean that pasting into rich-text targets like Google Docs or Gmail still inserts only the raw URL. On browsers without ClipboardItem, this can throw and interfere with the normal URL copy. Could we either set text/html via clipboardData.setData(...) here or simplify this to plain text link copy only?

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
```diff
@@ -115,6 +157,9 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
     resetSort();
   };
 
+  const buildShareLink = (name) =>
+    `${window.location.origin}/workflows/${id}/workspace?name=${encodeURIComponent(name)}`;
+
```
michaelbuchar, line 168: I think keeping the copied share link limited to workflow + file path is the right direction here. It avoids tying the shared URL to browsing state like page/search, while the internal open/close flow still preserves the user’s place. That separation seems good to me!

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
```diff
@@ -167,33 +169,84 @@ function getFileURL(workflow, fileName, preview = true) {
...
   }
 
   return (
     <Modal open closeIcon onClose={onClose ? onClose : () => {}}>
       <Modal.Header>{fileName}</Modal.Header>
+      {isSharedLink && (
+        <Modal.Description>
+          <Message
+            icon="info circle"
+            content="Workspace files can change over time — you're viewing the current version of this file."
```
michaelbuchar: I think the updated wording is much clearer now. I'm not a huge fan of dashes but this is a minor nitpick. :) Avoiding revision style language here seems right, since the link points to the current file rather than an immutable historical version.

reana-ui/src/pages/workflowDetails/components/CopyLinkButton.js
Sohn123: According to the docs it can be used this way: https://www.npmjs.com/package/copy-to-clipboard

> (clipboardData: ClipboardItem | DataTransfer) => ClipboardItem | void. Called before the write. On the async path, receives the constructed ClipboardItem and may return a new one to replace it (useful for custom MIME types or a different plain-text fallback). On the execCommand fallback path, receives the DataTransfer object; return value is ignored.

But this only works with the latest version of the library so running yarn install is required. Otherwise it doesn't work. But I think we decided anyway to go back to basic plain text copying.

michaelbuchar: BTW, even if we keep a separate button component here, it may be worth reusing the local copy pattern already used in [CodeSnippet.js](https://github.com/reanahub/reana-ui/blob/master/reana-ui/src/components/CodeSnippet.js#L75). That would keep clipboard behavior more consistent across reana-ui.

Sohn123: The problem with that is, that the `react-copy-to-clipoard` library depends on an older version of `copy-to-clipboard` that did not support the html/text mimetype that's why I introduced the new wrapper around the `copy-to-clipboard` library. But without the need it makes sense to reuse this component 👍🏼

michaelbuchar: Thanks, that makes sense. If we’re going back to basic plain text copying for this PR, that sounds good to me. We can always revisit richer clipboard behavior separately once we want to depend on that newer library path.

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
```diff
@@ -167,33 +169,84 @@ function getFileURL(workflow, fileName, preview = true) {
...
+        { page: 1, size: 1 },
+        JSON.stringify({ name: [fileName] }),
+      )
+      .then((resp) => {
+        const items = parseFiles(resp.data.items ?? []);
+        const match = items.find((f) => f.name === fileName);
+        setSize(match?.size?.raw ?? 0);
+      })
+      .catch(() => setSize(0));
+  }, [workflow, fileName]);
```
michaelbuchar: Small point: yarn lint adds a new hook-deps warning here. Not a blocker, but I think it would be better to either include that dependency explicitly or document why we intentionally want to ignore it, so the touched file stays without warnings.

michaelbuchar:
Overall this looks very nice, great job on the feature! :rocket: One thing that would make it even stronger would be a few focused tests around the URL behavior since that is the main new logic here. In particular, it would be useful to cover opening a deep link by exact file name, preserving internal page/search when closing the preview, showing a clear error when the file is missing, and making sure copylink does not throw when richer clipboard APIs are unavailable.

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
Sohn123: Ok I added logic for explicitly handling stale links. If there is no match for a filename or the fetch throws for another reason we display the following error message:
image

reana-ui/src/pages/workflowDetails/components/CopyLinkButton.js
Sohn123: I moved the copying logic now into a separate component that is also used by the `CodeSnippet` class. This captures the reused Popup > Trigger > Message pattern better than the `react-copy-to-clipboard` component. What do you think about that?

Sohn123:
Thank you for the review @michaelbuchar :) I addressed the comments. Feel free to take another look, if you think the tests are extensive enough and if you like the new `CopyButton` class.

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
```diff
@@ -167,33 +169,111 @@ function getFileURL(workflow, fileName, preview = true) {
...
+    if (stateSize !== undefined) {
+      setSize(stateSize);
+      return;
+    }
+    setSize(null);
+    client
+      .getWorkflowFiles(
+        workflow,
+        { page: 1, size: 1 },
+        JSON.stringify({ name: [fileName] }),
```
michaelbuchar, line 193: One thing I am still a bit worried about here is that this lookup does not seem fully exact yet. We fetch the file metadata through the normal name search and then try to match the result locally, but the backend search is wildcard based rather than a literal filename lookup. That means a valid shared link to a file whose name contains characters like []*? could still be treated as missing and show the stale link warning by mistake. For a “copy link to this file” feature, I think this lookup should be exact if possible.

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
```diff
@@ -167,33 +169,111 @@ function getFileURL(workflow, fileName, preview = true) {
...
+export default function FilePreview({ workflow, fileName, onClose }) {
   const config = useSelector(getConfig);
+  const location = useLocation();
+  const isSharedLink = !location.state?.internal;
+  const [size, setSize] = useState(null);
+  const [error, setError] = useState(null);
+  const shareUrl = `${window.location.origin}/workflows/${workflow}/workspace?name=${encodeURIComponent(fileName)}`;
+
+  useEffect(() => {
+    const stateSize = location.state?.size;
```
michaelbuchar, line 182: I think error may need to be reset at the start of this effect as well. Right now it is only set in the failure path, so if a stale link puts the modal into the error state and the user then navigates to another ?name= in the same tab, the old error can still hang around and hide a valid preview. It would be safer to clear the previous error before starting the next lookup.

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
```diff
@@ -158,6 +203,10 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
                 >
                   Size {headerIcon("size.raw")}
                 </Table.HeaderCell>
+                <Table.HeaderCell
```
michaelbuchar, line 213: Small UI nit: on smaller screens there is still quite a large gap between the Size column and the action buttons. I wonder if that spacing could be tightened a bit so the actions feel a little more visually connected to the file row. What do you think?

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
Sohn123: It actually wouldn't cause an issue here, because the FilePreview component gets remountend on each navigation event, so the state gets newly initialized. But I think to make the intent clear and for defensive coding, I'll add it anyway 👍🏼

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
Sohn123: I agree with you that the lookup should be exact if possible, but I don't see an API with exact file name matching, that supports metadata lookup only. The search API supports two arguments: `file_name` and `search`. The file_name argument will expand globs and match them. The `search` argument does case insensitive substring matching. To get the tightest match we can use the name filter + the file_name filter, but even there files could slip through. Let's say there is a file named:
`a.*` then with the current API implementation the search term `.*` would match it, because the glob search is positive as well as the lower case substring check. To work around this issue, we can just query more results, but the clean variant would be to add an exact match endpoint to the API. Should we tackle this issue here or is this something we postpone?

Sohn123: But it seems like glob handling in general is not ideal. For example you can create a file named `**` and when selecting it for download, it will just download all the files at once, which is not ideal.

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
Sohn123: I made the file size column smaller, so it is now closer together on smaller screens.

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
michaelbuchar: Thanks, that makes sense!

michaelbuchar:
Thanks a lot for the changes, they look great! Small suggestion on the commit/PR wording: I think a feature area scope would fit the existing commit style here a bit better than reana-ui. Something like `feat(workflow-details): add shareable links for workspace files` feels a bit more consistent with recent subjects in this repo. Also, if there is no linked issue for this PR, I would probably drop the Closes ... line from the description.

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
michaelbuchar: Looks good!

michaelbuchar (dismissed):
LGTM! 🦭

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
```diff
@@ -167,33 +169,112 @@ function getFileURL(workflow, fileName, preview = true) {
...
+    if (stateSize !== undefined) {
+      setSize(stateSize);
+      return;
+    }
+    setSize(null);
+    setError(null);
+    client
+      .getWorkflowFiles(
+        workflow,
+        { page: 1, size: 1, fileName },
```
tiborsimko: **🤖 Deep link can report an existing file as missing (HIGH)**

`FilePreview.js` looks up the deep-linked file with the wrong query parameter:

```js
client
  .getWorkflowFiles(
    workflow,
    { page: 1, size: 1, fileName },
    JSON.stringify({ name: [fileName] }),
  )
```

`getWorkflowFiles` spreads this object directly into the URL, so the request contains `page=1&size=1&fileName=<x>&search=<json>`. The server expects `file_name`, not `fileName` (see `workflows_workspace.py:480` and `reana-server/docs/openapi.json:5527`). It ignores the unknown key and lists the workspace using only the `search` filter.

The `search` filter is a case-insensitive substring match (`reana-workflow-controller/.../utils.py:571` — `_filter.casefold() in _file[k].casefold()`). With `size: 1`, the API returns a single row, so any sibling whose name *contains* the requested filename (e.g. `reports/final.txt.bak` next to `reports/final.txt`) can occupy that slot. The client's exact-match guard then throws:

```js
const match = items.find((f) => f.name === fileName);
if (match === undefined) {
  throw new Error("File not found");
}
```

The deep link then reports that the file may have been removed even though it exists and is accessible.

The new test does not catch this because its mock returns the exact row regardless of the request parameters, and its assertion expects the incorrect `fileName` key.

Suggested fix — send the documented parameter name, drop the unnecessary substring search, and update the test assertion:

```js
client.getWorkflowFiles(workflow, {
  page: 1,
  size: 1,
  file_name: fileName,
});
```

The exact-match guard can remain as a defensive check because `file_name` also supports glob patterns.

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
```diff
@@ -98,6 +129,17 @@ export default function WorkflowFiles({ id, page = 1, onPageChange }) {
...
 
+  const openPreview = (name, fileSize) => {
+    setSearchParams((prev) => openPreviewParams(prev, name), {
+      replace: false,
+      state: { internal: true, size: fileSize },
+    });
+  };
+
+  const closePreview = () => {
+    setSearchParams((prev) => closePreviewParams(prev), { replace: false });
```
tiborsimko: **🤖 Closing a preview pushes another history entry, so Back re-opens the modal (MEDIUM)**

Both `openPreview` and `closePreview` call `setSearchParams(..., { replace: false })`:

```js
const openPreview = (name, fileSize) => {
  setSearchParams((prev) => openPreviewParams(prev, name), {
    replace: false,
    state: { internal: true, size: fileSize },
  });
};

const closePreview = () => {
  setSearchParams((prev) => closePreviewParams(prev), { replace: false });
};
```

Opening should push an entry so that Back can close the modal. Closing also pushes an entry, however, producing `…, workspace, workspace?name=X, workspace`. After closing the modal, Back therefore returns to `workspace?name=X` and reopens it. Repeating this while browsing files continually grows the history stack with stale preview entries.

Suggested fix — remove the preview entry instead of extending the history. For internally opened previews, navigate back to the workspace entry. For direct shared links, replace the current URL after removing `name`:

```js
const closePreview = () => {
  if (location.state?.internal) {
    navigate(-1);
  } else {
    setSearchParams((prev) => closePreviewParams(prev), { replace: true });
  }
};
```

reana-ui/src/components/CopyButton.js
```diff
@@ -0,0 +1,68 @@
...
+  timeout = COPY_CHECK_TIMEOUT,
+  ...props
+}) {
+  const [copied, setCopied] = useState(false);
+  const timeoutRef = useRef();
+
+  useEffect(() => () => clearTimeout(timeoutRef.current), []);
+
+  const handleClick = () => {
+    copy(text, {
```
tiborsimko: **🤖 `CopyButton` announces success before the async clipboard write completes (MEDIUM)**

This PR replaces the v3 dependency pulled in by `react-copy-to-clipboard` with `copy-to-clipboard` v4.0.2, but `CopyButton` still uses a synchronous callback as its success signal:

```js
const handleClick = () => {
  copy(text, {
    format: "text/plain",
    onCopy: () => {
      setCopied(true);
      clearTimeout(timeoutRef.current);
      timeoutRef.current = setTimeout(() => setCopied(false), timeout);
    },
  });
};
```

In v4, `copy()` returns `Promise<boolean>`. With `format` and `onCopy` supplied, the Clipboard API path invokes `onCopy` before awaiting `navigator.clipboard.write(...)`. The component therefore sets `copied = true` before the write succeeds. If the write and fallback both fail, `copy()` resolves `false` after the UI has already displayed "Copied to clipboard!".

The new test hides this mismatch by mocking `copy` as a synchronous function that invokes `onCopy`.

Suggested fix — await the documented result and show success only when it resolves `true`; no `onCopy` hook is needed for plain text:

```js
const handleClick = async () => {
  const ok = await copy(text, { format: "text/plain" });
  if (!ok) return; // optionally show failure feedback
  setCopied(true);
  clearTimeout(timeoutRef.current);
  timeoutRef.current = setTimeout(() => setCopied(false), timeout);
};
```

The test should mock `copy.mockResolvedValue(true)` and wait for the success popup.

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
```diff
@@ -167,33 +169,112 @@ function getFileURL(workflow, fileName, preview = true) {
...
+  const location = useLocation();
+  const isSharedLink = !location.state?.internal;
+  const [size, setSize] = useState(null);
+  const [error, setError] = useState(null);
+  const shareUrl = `${window.location.origin}/workflows/${workflow}/workspace?name=${encodeURIComponent(fileName)}`;
+
+  useEffect(() => {
+    const stateSize = location.state?.size;
+    if (stateSize !== undefined) {
+      setSize(stateSize);
```
tiborsimko, line 184: **🤖 Cached history state can bypass the preview size limit (LOW)**

`FilePreview` uses the size carried in history state as authoritative whenever it's present:

```js
useEffect(() => {
  const stateSize = location.state?.size;
  if (stateSize !== undefined) {
    setSize(stateSize);
    return;
  }
  setSize(null);
  setError(null);
  client.getWorkflowFiles(workflow, …)
    …
}, [workflow, fileName, location.state]);
```

Browser history preserves each entry's `state` across Back/Forward navigation and reloads, so a size recorded at click time can outlive changes to an active workflow's workspace. If the file grows from 10 bytes to 100 MB, revisiting or reloading that entry still supplies `state.size = 10`. `checkConstraints` uses the stale value, the size check passes, and the modal attempts to render the now-large file inline.

The defensive "refetch when state is missing" branch only runs when `state.size` is `undefined`, which it isn't in this case.

Suggested fix — use `state.size` only as an optimistic initial value and always refresh the metadata:

```js
useEffect(() => {
  const stateSize = location.state?.size;
  setSize(stateSize ?? null);  // optimistic, no spinner if we have a hint
  setError(null);
  client.getWorkflowFiles(workflow, { file_name: fileName, page: 1, size: 1 })
    .then(({ data }) => {
      const match = parseFiles(data.items ?? []).find((f) => f.name === fileName);
      if (!match) throw new Error("File not found");
      setSize(match.size?.raw ?? 0);
    })
    .catch(setError);
}, [workflow, fileName]);
```

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
```diff
@@ -167,33 +169,112 @@ function getFileURL(workflow, fileName, preview = true) {
...
+        />
+      );
+    }
   }
 
   return (
     <Modal open closeIcon onClose={onClose ? onClose : () => {}}>
       <Modal.Header>{fileName}</Modal.Header>
+      {isSharedLink && !error && (
+        <Modal.Description>
```
tiborsimko: **🤖 Shared-link notice bypasses the modal content wrapper (LOW)**

The shared-link banner is placed as a direct child of `<Modal>`:

```jsx
<Modal open closeIcon onClose={onClose ? onClose : () => {}}>
  <Modal.Header>{fileName}</Modal.Header>
  {isSharedLink && !error && (
    <Modal.Description>
      <Message
        icon="info circle"
        content="Workspace files can change over time — you're viewing the current version of this file."
        info
      />
    </Modal.Description>
  )}
  <Suspense …>{preview}</Suspense>
  …
</Modal>
```

Semantic UI expects `Modal.Description` beneath `Modal.Content`. Its stylesheet targets `.ui.modal > .content > .description`, and the content wrapper supplies the body padding, background, font sizing, and responsive layout. Here it renders as a direct `<div class="description">` child and bypasses those rules.

The inner `<Message>` remains visible, but its placement and spacing are not governed by the modal content layout.

Suggested fix — place the notice inside `Modal.Content`:

```jsx
{isSharedLink && !error && (
  <Modal.Content>
    <Modal.Description>
      <Message
        icon="info circle"
        content="Workspace files can change over time — you're viewing the current version of this file."
        info
      />
    </Modal.Description>
  </Modal.Content>
)}
```

Sohn123: I added the Modal.content wrapper.

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
Sohn123: I would advise against that, because each preview would then hit the API again for no good reason. I think this failure case is too rare and the consequences not bad enough, to be worth this extra load.

reana-ui/src/components/CopyButton.js
Sohn123: Done.

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
Sohn123: For me that is the expected way how the navigation should behave, so I don't really see an issue there. Opening the preview is a navigation action and it gets correctly reflected in the history.

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
Sohn123: I fixed the parameter naming, but I would still like to keep the name search parameter, because the additional constraint does not hurt, but rather narrows down the search in case of glob matching.

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
tiborsimko: Opening the preview being recorded in history is fine. The issue is specifically that clicking Close pushes a second entry instead of returning to or replacing the preview entry. After workspace → workspace?name=X → Close, pressing Back reopens X. Could we keep the push on open but use `navigate(-1)` for an internally opened preview and `replace: true` for a directly opened shared link?

reana-ui/src/pages/workflowDetails/components/filePreview/FilePreview.js
tiborsimko: Agreed that another request on every normal click would be unnecessary. We can limit revalidation to history restoration: `useNavigationType()` reports `PUSH` for the initial row click and `POP` for reload/Back/Forward. Trusting `state.size` only on `PUSH`, and fetching metadata on `POP`, would preserve the fast path while preventing stale history state from bypassing the limit.

reana-ui/src/pages/workflowDetails/components/WorkflowFiles.js
Sohn123: Ok I added the change and opening a preview is now no longer part of the navigation history.

tiborsimko: The chosen navigation behavior works for me. One mechanical issue remains in the latest commit: `location` and `navigate` are currently undefined in `WorkflowFiles`, and CI's lint/Prettier checks fail. Please add `useLocation()`/`useNavigate()` and format the new block. The `useNavigationType()` metadata-refresh change looks good.

michaelbuchar (approved):
🚀

PR #507, opened 2026-06-04
fix(workflow-list): improve sharing filter URL state handling

### LINKED ISSUE #508: Bug: Workflow sharing filters are not working
When trying to filter using the owned by/shared by button, the filtering does not work and the user gets redirected to the list overview page. Also selecting anybody does not work and displays no results.

Sohn123:
The filters for shared workflows weren't working correctly. This change carves out the parsing logic for the search params in a separate hook and changes the search parameters from shared-by to owned-by. The new URL contract looks like this:

My own workflows, that I own:
```
http://localhost:3000/
http://localhost:3000/?owned-by=you
```

All workflows including the ones shared with me by someone else:
```
http://localhost:3000/?owned-by=anybody
```

All workflows shared with me by xyz
```
http://localhost:3000/?owned-by=xyz
```

All workflows I shared with someone
```
http://localhost:3000/?shared-with=anybody
```

All workflows I shared with xyz
```
http://localhost:3000/?shared-with=xyz
```

Fixes: #508

### DIFF
--- a/reana-ui/src/components/WorkflowShareModal.js
+++ b/reana-ui/src/components/WorkflowShareModal.js
@@ -24,7 +24,12 @@ import {
   Popup,
 } from "semantic-ui-react";
 
-import { closeShareWorkflowModal, fetchWorkflowShareStatus } from "~/actions";
+import {
+  closeShareWorkflowModal,
+  fetchUsersYouSharedWith,
+  fetchWorkflowShareStatus,
+  WORKFLOW_LIST_REFRESH,
+} from "~/actions";
 import client from "~/client";
 import {
   getLoadingWorkflowShareStatus,
@@ -63,6 +68,8 @@ function WorkflowShareStatus({
       })
       .then(() => {
         dispatch(fetchWorkflowShareStatus(id));
+        dispatch(fetchUsersYouSharedWith());
+        dispatch({ type: WORKFLOW_LIST_REFRESH });
         handleUnshareWorkflowSuccess(userEmailToUnshareWith);
         setLoadingUnshareWorkflow(false);
         setConfirmUnshareOpen(false);
@@ -283,8 +290,10 @@ export default function WorkflowShareModal() {
     }
 
     Promise.allSettled(requests).then(() => {
-      // update share status
+      // update share status, refresh sharing user lists and update workflow list
       dispatch(fetchWorkflowShareStatus(id));
+      dispatch(fetchUsersYouSharedWith());
+      dispatch({ type: WORKFLOW_LIST_REFRESH });
       // reset form and share button
       resetShareForm();
       setLoadingShareWorkflow(false);
--- a/reana-ui/src/pages/workflowList/WorkflowList.js
+++ b/reana-ui/src/pages/workflowList/WorkflowList.js
@@ -9,11 +9,9 @@
 */
 
 import moment from "moment";
-import { useEffect, useMemo, useRef, useState } from "react";
-import { useSearchParams } from "react-router-dom";
+import { useEffect, useRef, useState } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import { Container, Dimmer, Dropdown, Icon, Loader } from "semantic-ui-react";
-import isEqual from "lodash/isEqual";
 
 import { fetchUsersSharedWithYou, fetchWorkflows } from "~/actions";
 import {
@@ -27,20 +25,14 @@ import {
   getWorkflowRefresh,
   getUsersSharedWithYou,
 } from "~/selectors";
-import { NON_DELETED_STATUSES } from "~/config";
 import { Title, Pagination, Search } from "~/components";
 import BasePage from "../BasePage";
 import Welcome from "./components/Welcome";
 import WorkflowFilters from "./components/WorkflowFilters";
 import WorkflowList from "./components/WorkflowList";
+import { useWorkflowListQuery } from "./useWorkflowListQuery";
+import { WORKFLOW_LIST_PAGE_SIZE_OPTIONS } from "./workflowListQuery";
 import styles from "./WorkflowList.module.scss";
-
-const pageSizeOptions = [5, 10, 20, 50, 100].map((size) => ({
-  key: size,
-  text: String(size),
-  value: size,
-}));
-const DEFAULT_PAGE_SIZE = pageSizeOptions[1].value;
 
 export default function WorkflowListPage() {
   return (
@@ -53,7 +45,6 @@ export default function WorkflowListPage() {
 function Workflows() {
   const currentUTCTime = () => moment.utc().format("HH:mm:ss [UTC]");
   const [refreshedAt, setRefreshedAt] = useState(currentUTCTime());
-  const [searchParams, setSearchParams] = useSearchParams();
   const dispatch = useDispatch();
   const config = useSelector(getConfig);
   const workflows = useSelector(getWorkflows);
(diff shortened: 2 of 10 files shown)


tiborsimko:
WRT "move share filtering to a separate row" we should discuss IRL, because there are other options such as side facet... In general, it may be good to separate the necessary fix from the new UX/UI features.

reana-ui/src/pages/workflowList/components/WorkflowSharingFilter.js
```diff
@@ -28,16 +28,12 @@ const sharingFilterOptions = [
 export default function WorkflowSharingFilters({
   ownedByFilter,
   setOwnedByFilter,
-  sharedWithFilter,
-  sharedWithMode,
-  setSharedWithFilter,
-  setSharedWithModeInUrl,
+  sharedWithUser,
```
tiborsimko: **🤖 Dead `WorkflowSharingFilter` component and SCSS module left in the tree (LOW)**

`src/pages/workflowList/components/WorkflowFilters.js` no longer imports `WorkflowSharingFilters`, and a tree-wide grep confirms there is no other caller:

```
$ grep -rn "WorkflowSharingFilter" src --include='*.js'
src/pages/workflowList/components/WorkflowSharingFilter.js: …
```

The file was *refactored* in this PR (props renamed `sharedWithFilter` → `sharedWithUser`, `sharedWithMode` → `viewingSharedWithMe`, setters renamed) instead of deleted, and `WorkflowSharingFilter.module.scss` is similarly orphaned. This is genuine dead code that will rot and confuse future readers searching for filter logic.

Suggested fix — `git rm src/pages/workflowList/components/WorkflowSharingFilter.js src/pages/workflowList/components/WorkflowSharingFilter.module.scss` in this PR.

reana-ui/src/pages/workflowList/useWorkflowListQuery.js
```diff
@@ -0,0 +1,235 @@
...
+          !WORKFLOW_LIST_CATEGORIES.includes(nextCategory) ||
+          nextCategory === WORKFLOW_LIST_DEFAULT_CATEGORY
+        ) {
+          next.delete("category");
+        } else {
+          next.set("category", nextCategory);
+        }
+        // Clear all sharing user filters when switching categories
+        next.delete("shared-by");
+        next.delete("shared-with-user");
```
tiborsimko: **🤖 Clicking "All" silently no-ops when the URL still carries legacy `?shared=true` or `?shared-with=true` (LOW)**

`setCategory` here removes the `category` URL key when the user picks the default (`"all"`), and clears `shared-by` / `shared-with-user`, but it does **not** clear the legacy keys `shared` and `shared-with`. Because `parseWorkflowListQuery` still honours those legacy keys (good for bookmarks):

```js
} else if (searchParams.get("shared-with") === "true") {
  category = "i-shared";
} else if (searchParams.get("shared") === "true" || searchParams.get("shared-by")) {
  category = "shared-with-me";
}
```

…a user who lands on `/workflows?shared=true` (a legacy bookmark) and then clicks the "All" chip ends up with URL `?shared=true` again — the parser re-derives `category = "shared-with-me"`, and the click appears to do nothing. The same trap exists for `?shared-with=true` → "All". This re-creates a milder version of the very bug this PR is fixing (filter clicks that appear to bounce back).

The symptom is narrow on purpose: `setCategory` only *deletes* the `category` URL key for the default value `"all"`, so only the "All" chip is exposed. Switching to `"mine"` / `"shared-with-me"` / `"i-shared"` writes an explicit `?category=…`, and `parseWorkflowListQuery` checks `rawCategory` first — so those chips override the legacy params even if `shared=true` is still in the URL. Nothing the new code writes can produce a stuck-legacy URL either; the bug only surfaces on bookmarks from before this PR.

Suggested fix — also strip the legacy keys in `setCategory`:

```js
next.delete("shared-by");
next.delete("shared-with-user");
next.delete("shared");          // legacy
next.delete("shared-with");     // legacy
```

reana-ui/src/pages/workflowList/components/WorkflowFilters.js
```diff
@@ -2,60 +2,126 @@
...
+  setSharedWithUser,
+  showOpenSessionsOnly,
+  setShowOpenSessionsOnly,
 }) {
+  const dispatch = useDispatch();
+  const usersSharedWithYou = useSelector(getUsersSharedWithYou, _.isEqual);
+  const usersYouSharedWith = useSelector(getUsersYouSharedWith, _.isEqual);
+
+  useEffect(() => {
+    if (category === "shared-with-me") {
```
tiborsimko: **🤖 `fetchUsersSharedWithYou` dispatched twice on mount for shared-with-me category (LOW)**

`src/pages/workflowList/WorkflowList.js` dispatches `fetchUsersSharedWithYou()` unconditionally on mount (it feeds the empty-state check `usersSharedWithYou.length === 0` further down). This `useEffect` then dispatches it again the moment `category === "shared-with-me"`:

```js
// WorkflowList.js
useEffect(() => {
  dispatch(fetchUsersSharedWithYou());
}, [dispatch]);

// WorkflowFilters.js (this hunk)
useEffect(() => {
  if (category === "shared-with-me") {
    dispatch(fetchUsersSharedWithYou());
  } else if (category === "i-shared") {
    dispatch(fetchUsersYouSharedWith());
  }
}, [dispatch, category]);
```

Both go through the same action with no debounce/dedupe, so a user landing on the page with category=shared-with-me gets two HTTP calls for the same data. Not user-visible, but wasteful and easy to forget.

Note that the unconditional dispatch in `WorkflowList.js` is **load-bearing** — it feeds the Welcome-page gate:

```js
if (!hasUserWorkflows && usersSharedWithYou.length === 0) {
  return <Welcome />;
}
```

Removing the unconditional fetch would break this for users in the `all` category who own no workflows but do have shared ones — `WorkflowFilters` only fires the fetch when `category === "shared-with-me"`, so the list never loads and they get incorrectly bounced to `<Welcome />`.

Suggested fix — keep the parent fetch in `WorkflowList.js`, and either gate this filter dispatch on `usersSharedWithYou.length === 0` (skip if already loaded), or accept the duplicate as a minor cost.

reana-ui/src/pages/workflowList/components/WorkflowFilters.js
```diff
@@ -2,60 +2,126 @@
...
+    } else if (category === "i-shared") {
+      dispatch(fetchUsersYouSharedWith());
+    }
+  }, [dispatch, category]);
+
+  const sharedByUserOptions = useMemo(
+    () => [
+      { key: "anybody", text: "anybody", value: "anybody" },
+      ...usersSharedWithYou.map((user, i) => ({
+        key: i,
```
tiborsimko: **🤖 Array index used as React `key` for user dropdown options (NIT)**

```js
...usersSharedWithYou.map((user, i) => ({
  key: i,
  text: user.email,
  value: user.email,
})),
```

`user.email` is already unique within the list and would be a more stable key than the positional index. If the backend ever reorders the list (e.g. sort by recency vs. alphabetical), index keys cause React to reconcile against the wrong items and can stick stale UI state (focus, selection) onto rows it doesn't belong to. The same pattern appears in `sharedWithUserOptions` a few lines below, and in the now-orphaned `WorkflowSharingFilter.js` (which should just be deleted per the separate cleanup comment).

Suggested fix — use `key: user.email` in both option lists.

reana-ui/src/pages/workflowList/components/WorkflowSharingFilter.js
Sohn123: Yes good catch

reana-ui/src/pages/workflowList/components/WorkflowFilters.js
Sohn123: I would say that is not really a big issue and acceptable. Because it only triggers if someone saves the shared-with-me bookmark and opens it directly and then it is not really a big load and having the additional gate logic inside the second hook is making it unnecessarily complicated.

reana-ui/src/pages/workflowList/useWorkflowListQuery.js
Sohn123: I reproduced the issue and added the required delete calls.

reana-ui/src/pages/workflowList/components/WorkflowFilters.js
Sohn123: Done

Sohn123:
I addressed the nit comments. So let's discuss the design choice for how to display the filters. I would suggest we keep both changes in the same PR, but maybe split them up into two commits, but we can also carve out the UI/UX changes into a separate PR.

michaelbuchar:
Nice refactor overall. Since this PR moves a lot of logic into the new query parsing/state layer, I think it would be worth adding a few focused tests for the main URL-driven cases as well: legacy ?shared=true / ?shared-with=true bookmarks, i-shared -> owned by -> you, shared-with-me with shared-by=<user>, invalid page normalization, and status=deleted normalization. That would give this change a much firmer safety net.

michaelbuchar:
The current commit/PR title feels a bit broad. fix(reana-ui) does not say much about what part of the UI changed, and this PR is really focused on workflow-list sharing filter/query logic. Could we make the headline a bit more specific so the history is easier to scan later? Something like fix(workflow-list): fix sharing filter URL state might fit better.

reana-ui/src/pages/workflowList/components/WorkflowFilters.js
```diff
@@ -2,59 +2,120 @@
...
+      const next = new URLSearchParams(prev);
+      next.delete("page");
+      next.delete("shared");
+      next.delete("shared-with");
+      next.delete("shared-by");
+      next.delete("shared-with-user");
+      if (on) {
+        next.set("category", "i-shared");
+      } else {
+        next.delete("category"); // "all" is the default
```
michaelbuchar: I think this is where the remaining conflict happens. When switching from “Shared with” back to “Owned by”, the other path has already set category=mine, but this branch clears category again, so the URL can fall back to the default “all” view. Could we make that switch a single atomic update so the two URL writers do not fight each other?

Sohn123: Hey @michaelbuchar ,
in the latest version I changed it quite a bit, to allow for atomic writes. In general I moved all URL handling outside of the components and got rid of the category URL flag and just use now `?owned-by` and `?shared-with` to define the sharing filter status.

reana-ui/src/pages/workflowList/useWorkflowListQuery.js
```diff
@@ -0,0 +1,204 @@
...
+        resetPage(next);
+      });
+    },
+    [setSearchParams],
+  );
+
+  const setSharing = useCallback(
+    (ownedBy, sharedWith) => {
+      updateParams(setSearchParams, (next) => {
+        next.delete("shared"); // remove legacy param on every write
```
michaelbuchar, line 167: There is one remaining legacy URL regression here. We still support old `?shared-by=<email>` bookmarks in `parseWorkflowListQuery()`, but `setSharing(undefined, undefined)` only clears shared and shared-with, not shared-by. That means a user who lands on something like `?shared-by=alice@example.org` and then switches to Owned by -> anybody can get snapped back to Alice again, because the parser falls back to ownedBy from the stale legacy param. Could we also delete shared-by when writing the new sharing state in setSharing()? That would preserve backwards compatibility for old bookmarks, while ensuring the first user interaction migrates them cleanly to the new URL contract.

Sohn123: Yes will do

michaelbuchar:
Nice cleanup on the URL state handling! One small thing before merge: the PR description still documents the old `?category=...` contract, but the implementation now uses `?owned-by=` / `?shared-with=`. Right now someone reviewing or manually validating from the PR body would be testing outdated URLs. Could you update the examples in the PR description to match the current behavior?

michaelbuchar:
One more thing, the new header `fix(wf-sharing): ...` feels a bit too ad hoc for release-please notes. It would be better to reuse an area name that already exists in the history so the generated changelog stays more consistent and easier to scan later. Something like `fix(workflow-list): improve sharing filter URL state handling` would read more clearly in my opinion.

Sohn123:
Thanks for the review @michaelbuchar :)

I updated the PR description + title accordingly. Is there an overview somewhere about the scopes that exist for the commit messages or should I just look at existing commits?

michaelbuchar:
> Thanks for the review @michaelbuchar :)
> 
> I updated the PR description + title accordingly. Is there an overview somewhere about the scopes that exist for the commit messages or should I just look at existing commits?

Yes, there is a short overview in the wiki. The main page is [Tips for Git](https://github.com/reanahub/reana/wiki/Tips-for-Git#commit-message-style) and it says to use the usual <type>(scope): <description> (#pr) format and to reuse existing scopes where possible since they appear in release notes. The [Developing](https://github.com/reanahub/reana/wiki/Developing) page points to the same guidance and says that if you are unsure, checking the history of the component is the right fallback. So for reana-ui, I usually reuse an existing scope rather than inventing a new one. Any other thoughts @tiborsimko?

reana-ui/src/components/WorkflowShareModal.js
```diff
@@ -283,8 +288,9 @@ export default function WorkflowShareModal() {
     }
 
     Promise.allSettled(requests).then(() => {
-      // update share status
+      // update share status and refresh sharing user lists
       dispatch(fetchWorkflowShareStatus(id));
```
michaelbuchar, line 294: Nice improvement, this keeps the modal state much more in sync after share/unshare. One thing still seems to be missing though: we refresh fetchWorkflowShareStatus() and fetchUsersYouSharedWith(), but not the workflow list itself. If someone is currently filtered with ?shared-with=anybody or ?shared-with=<user>, the visible list can stay stale until a manual reload or the next background refresh. Could we also trigger a workflow list refresh here so the filtered results update immediately?

Sohn123: I see the concern. I only could trigger an issue with unsharing a workflow and I fixed that by sending the refetch signal to the workflow list here as well as when unsharing a workflow.

reana-ui/src/pages/workflowList/components/WorkflowSharingFilter.js
```diff
@@ -79,44 +76,35 @@ export default function WorkflowSharingFilters({
...
-      setOwnedByFilter("you");
-      setSharedWithFilter(undefined);
-      setSelectedUser("you");
-      setSharedWithModeInUrl(false);
+      setSharing(undefined, undefined);
     } else {
-      setSharedWithFilter("anybody");
-      setSelectedUser("anybody");
-      setSharedWithModeInUrl(true);
+      setSharing(undefined, "anybody");
```
michaelbuchar, line 99: I think one behavior changed here in a way that looks broader than the original #508 fix. Switching from Shared with back to Owned by currently calls setSharing(undefined, undefined), which then falls through to the default all-accessible-workflows view instead of the old you view. So the toggle no longer returns the user to their own workflows, it returns them to the combined owned/shared list. Unless that product change is explicitly intended, could we make the Owned by switch default back to owned-by=you? I would treat this as blocking because it changes the filter semantics when moving between modes.

reana-ui/src/pages/workflowList/components/WorkflowSharingFilter.js
michaelbuchar: Small regression in request churn here: WorkflowList already fetches fetchUsersSharedWithYou() on mount, and this effect does it again unconditionally. It is not user visible, but it means opening the page now issues the same request twice every time. Could we either rely on the parent fetch or gate this one so it only runs when this component actually needs to load missing data?

michaelbuchar:
Nice cleanup overall. One thing still missing for me before merge is a focused test safety net for the new URL state layer. Most of the risk now sits in useWorkflowListQuery() and workflowListQuery.js, but I did not find workflow-list tests for the main compatibility and normalization cases: legacy ?shared=true / ?shared-with=true, shared-with -> owned-by -> you, invalid page normalization, and status=deleted normalization. I think adding a few query-level regression tests around those paths would make this refactor much safer.

reana-ui/src/pages/workflowList/components/WorkflowSharingFilter.js
Sohn123: Ok I did this as an intentional change, because I think it was quite confusing, that when you only had a workflow shared with you, you didn't see the workflow but just an empty list, when you visit the page. You had to select the owned by anybody filter. I think by default we should show all the non-deleted workflows a user can see and then narrow down via filtering. But if that is not what we want, we can also go back to the old behavior, so the default page is owned-by=you

Sohn123: I decided to move this behavior change to the more general UI/behavior rework and only keep the fix here. Updated the PR accordingly.

reana-ui/src/pages/workflowList/components/WorkflowSharingFilter.js
Sohn123: It is kind of hard to gate this properly, because in redux we have no real notion of is this data up to date/when was it fetched. I gated it now to check the length of the array, but if the user hasn't shared something yet, we might see the double fetch nevertheless. Is there a better way to gate this?

reana-ui/src/pages/workflowList/components/WorkflowSharingFilter.js
```diff
@@ -68,55 +65,46 @@ export default function WorkflowSharingFilters({
...
-    dispatch(fetchUsersYouSharedWith());
-    dispatch(fetchUsersSharedWithYou());
-  }, [dispatch]);
+    if (usersYouSharedWith.length === 0) {
+      dispatch(fetchUsersYouSharedWith());
+    }
+    if (usersSharedWithYou.length === 0) {
+      dispatch(fetchUsersSharedWithYou());
+    }
+  }, [dispatch, usersSharedWithYou.length, usersYouSharedWith.length]);
```
michaelbuchar, line 78: I think there may still be a small double-fetch window here though: WorkflowList already calls fetchUsersSharedWithYou() on mount, and this effect can still see the initial empty array before Redux updates, so both requests may fire on first render. Since the current store does not track loaded/loading state, would it be better to rely on the parent fetch for this path, or introduce an explicit flag so we can gate it more reliably?

reana-ui/src/pages/workflowList/components/WorkflowSharingFilter.js
michaelbuchar: Small follow-up: Could we use user.email instead of key: index for these option lists? It should be the more stable identifier if the backend order changes later.

reana-ui/src/pages/workflowList/components/WorkflowSharingFilter.js
Sohn123: I think relying on the parent fetch is not perfect, because it means that this component is not reusable anywhere else if the parent is not fetching the list and passing it. Also adding and maintaining an additional flag, seems like overkill for me. I think the potential double fetch is acceptable in this case, as the db load of this query is not really big.

michaelbuchar: I see the tradeoff, I do not think we need extra loaded/loading state just for this PR, but I also think the current guard still leaves a small first render double fetch window. Since this is minor and not user visible, I am ok with it for this PR, but I would prefer to keep the thread open so we remember this discussion and see if we need to touch this area again later. :)

michaelbuchar (dismissed):
Looks good to me! Thanks for all the changes. 🦭

reana-ui/src/pages/workflowList/components/WorkflowSharingFilter.js
Sohn123: Ok sounds good :) I think migrating to a library that supports loading/fetching state handling like react query might be a good approach https://tanstack.com/query/latest/docs/framework/react/overview this could replace a lot of our redux code. See #510 for a proposal :)

reana-ui/src/pages/workflowList/WorkflowList.js
```diff
@@ -66,307 +57,44 @@ function Workflows() {
...
-      }
-      rafIdRef.current = 0;
-    });
-
-    return () => {
-      if (rafIdRef.current) cancelAnimationFrame(rafIdRef.current);
-      rafIdRef.current = 0;
-    };
-  }, [dispatch, requestParams, configLoaded, reanaToken]);
+    if (lastParamsRef.current === requestParams) return;
```
tiborsimko, line 94: **🤖 Fetch can race the URL-normalization rewrite during in-app navigation (LOW)**

The workflow fetch effect now dedupes purely by object identity:

```js
const lastParamsRef = useRef();
useEffect(() => {
  if (!configLoaded) return;
  if (lastParamsRef.current === requestParams) return;
  lastParamsRef.current = requestParams;
  dispatch(fetchWorkflows(requestParams));
}, [dispatch, requestParams, configLoaded]);
```

For a stable URL the `location.search → searchParams → query → requestParams` `useMemo` chain keeps `requestParams` referentially stable, so identity equality is correct there. The gap is the **first render of a URL that an effect immediately normalizes**, reached via in-app navigation. On `?status=deleted&page=2`:

1. `parseWorkflowListQuery()` yields `page: 2, status: undefined` (the page is *not* reset during parse — only the write-back effect resets it).
2. `useWorkflowListQuery()` (`useWorkflowListQuery.js:56`) schedules a `setSearchParams` that deletes `status` and resets `page`.
3. If `configLoaded` is already `true`, the fetch effect runs on that same commit with the **stale page-2** `requestParams` and dispatches before the normalized page-1 render happens.

`fetchWorkflows` has no request correlation or cancellation — `WORKFLOWS_RECEIVED` overwrites the store unconditionally (`reana-ui/src/reducers.js:277`), so if the page-2 response lands after the page-1 response, the list shows page-2 data while the URL/pagination say page 1. The old `requestAnimationFrame` effect cancelled the superseded fetch on the cleanup render; plain identity equality does not, and deep equality would not help either since the two requests carry *different* params.

Scope/severity: cold page loads are protected — `config.isLoaded` starts `false` (`reducers.js:78`) and is only set by the network config fetch, so the no-network normalization effects rewrite the URL before the fetch ever unblocks. The race only triggers on in-app navigation to a legacy `?status=deleted&page=N` (N>1) URL (the UI never emits `status=deleted` itself) and self-heals on the next poll. Hence LOW, not blocking — but it is a behavioural regression from the previous rAF coalescing and the fix is cheap.

Suggested fix — make the parsed query already match the normalized URL so there is no stale intermediate to race: reset `page` to `1` inside `parseWorkflowListQuery` whenever it drops `status=deleted`. (Invalid pages already parse as page 1, so they don't trigger the race — but driving the fetch off the normalized query rather than the not-yet-rewritten URL also avoids the redundant duplicate request that fires after any URL cleanup.) Alternatively, restore the rAF/microtask cancellation so a superseded fetch never fires.

reana-ui/src/pages/workflowList/components/WorkflowFilters.js
```diff
@@ -2,7 +2,7 @@
   -*- coding: utf-8 -*-
 
   This file is part of REANA.
-  Copyright (C) 2020, 2022, 2023 CERN.
+  Copyright (C) 2020, 2022, 2023, 2026 CERN.
```
tiborsimko: **🤖 Modified `WorkflowSharingFilter.js` keeps a stale copyright year range (LOW)**

Every other file this PR touches had its copyright year bumped (`WorkflowFilters.js`, `props.js`, `util.js` all gained `2026`), but the sharing filter — which this PR modifies substantially (prop rename, `key` change, effect rewrite) — still reads:

```
Copyright (C) 2023 CERN.
```

`git log --format=%ad --date=format:%Y -- …/WorkflowSharingFilter.js` shows the file was edited in **2023, 2024, 2025, 2026**, so per the project header convention (current year plus any years missed by earlier contributors) it is missing three years.

Suggested fix — `Copyright (C) 2023, 2024, 2025, 2026 CERN.`

While nearby: `WorkflowFilters.js` was bumped to `2020, 2022, 2023, 2026` but its history also includes **2025** (`git log` → `2020 2022 2023 2025 2026`), so it should read `Copyright (C) 2020, 2022, 2023, 2025, 2026 CERN.`

reana-ui/src/pages/workflowList/workflowListQuery.test.js
```diff
@@ -0,0 +1,208 @@
+import { useEffect } from "react";
```
tiborsimko, line 9: **🤖 New test file is missing the REANA MIT header (NIT)**

The new test file starts directly with an import:

```js
import { useEffect } from "react";
import { fireEvent, render, screen, waitFor } from "@testing-library/react";
```

The other recently added test files carry the standard REANA/MIT header (e.g. `src/components/__tests__/CopyButton.test.js`, `src/components/__tests__/LauncherLabel.test.js`), and the project convention is that all source files start with one.

Suggested fix — prepend the standard block:

```js
/*
  This file is part of REANA.
  Copyright (C) 2026 CERN.

  REANA is free software; you can redistribute it and/or modify it
  under the terms of the MIT License; see LICENSE file for more details.
*/
```

reana-ui/src/pages/workflowList/components/WorkflowSharingFilter.js
```diff
@@ -47,7 +44,7 @@ export default function WorkflowSharingFilters({
       { key: "you", text: "you", value: "you" },
       { key: "anybody", text: "anybody", value: "anybody" },
       ...usersSharedWithYou.map((user, index) => ({
-        key: index,
+        key: user.email,
```
tiborsimko, line 47: **🤖 Leftover unused `index` map argument (NIT)**

`reana-ui/src/pages/workflowList/components/WorkflowSharingFilter.js:46` and `:58`

Both option maps still bind `index` even though the stale-key fix replaced `key: index` with `key: user.email`:

```js
...usersSharedWithYou.map((user, index) => ({ key: user.email, text: user.email, value: user.email })),
...usersYouSharedWith.map((user, index) => ({ key: user.email, text: user.email, value: user.email })),
```

`index` is now dead. Note this will **not** fail CI: the `react-app` ESLint config sets `no-unused-vars` to `['warn', { args: 'none', … }]`, so unused function arguments are not reported, and `eslint .` runs without `--max-warnings 0` anyway. So this is leftover noise, not a lint failure.

Suggested fix — drop the unused parameter in both places: `.map((user) => ({ key: user.email, text: user.email, value: user.email }))`.

reana-ui/src/pages/workflowList/workflowListQuery.js
```diff
@@ -0,0 +1,164 @@
...
+    shared = false;
+    ownedBy = undefined;
+    sharedWith = undefined;
+  } else if (query.ownedBy === "anybody") {
+    // shared-with-me (anybody)
+    shared = true;
+    ownedBy = undefined;
+    sharedWith = undefined;
+  } else if (query.ownedBy) {
+    // shared-with-me ("anybody" or specific email)
```
tiborsimko: **🤖 Misleading branch comment in `serializeQueryToApiParams` (NIT)**

```js
} else if (query.ownedBy === "anybody") {
  // shared-with-me (anybody)
  shared = true;
  …
} else if (query.ownedBy) {
  // shared-with-me ("anybody" or specific email)   ← "anybody" already handled above
  shared = false;
  ownedBy = query.ownedBy;
  …
```

At the `else if (query.ownedBy)` branch the value can only be a specific owner email — `"anybody"` was consumed by the preceding branch — so the comment is misleading.

Suggested fix — `// shared-with-me (specific owner email)`.

tiborsimko:
**🤖 PR description documents the wrong default-view contract (LOW)**

The PR description lists the bare URL under "All workflows including the ones shared with me by someone else":

```
All workflows including the ones shared with me by someone else:
  http://localhost:3000/
  http://localhost:3000/?owned-by=anybody
```

…but the implemented contract defaults the empty URL to **owned by you** (mine only). `parseWorkflowListQuery()` returns `ownedBy: "you"` when no sharing param is present (`workflowListQuery.js:88`), and `serializeQueryToApiParams()` maps that to `shared: false` with no `sharedBy`/`sharedWith` (`workflowListQuery.js:116`). So `http://localhost:3000/` is equivalent to `?owned-by=you`, not to `?owned-by=anybody`.

This matters because the default view was the subject of the review discussion: the "default to all accessible workflows" behaviour change was deliberately reverted to keep the default at `owned-by=you`, but the PR body still describes the reverted behaviour. A reviewer or manual tester following the description would validate against the wrong contract.

Suggested fix — move the bare `http://localhost:3000/` line down to the "owned by me" section alongside `?owned-by=you`, and leave only `?owned-by=anybody` under "All workflows including the ones shared with me".

reana-ui/src/pages/workflowList/WorkflowList.js
Sohn123: Thanks that's a good catch! I think this kind of problem highlights, that our current Redux setup is not super easy to use and can lead to errors. I think replacing it with react query, could help us in such cases, because it allows for example for correlation of requests. I addressed this comments as suggested by resetting the page to one if the legacy deleted is present.

Sohn123:
@tiborsimko I updated the PR description and addressed all the findings in the latest commit.

tiborsimko (approved):

PR #511, opened 2026-06-08
feat(workflow-details): permanent line anchors for logs

### LINKED ISSUE #503: ui: render workflow logs with permanent line anchors
Part of reanahub/reana-ui#377.

Replace or extend plain log rendering so engine, job, and service logs have stable line anchors.

Acceptance criteria:
- engine, job, and service logs expose line anchors such as `#L42`
- opening a URL with a line fragment scrolls to and highlights the line
- invalid or out-of-range line fragments do not crash the page
- existing job/service selection by URL continues to work

Sohn123:
Add a new LogViewer.js component that uses virtual DOM with Tanstack virtual to display logs. Each line now has a line number. The LogViewer components reads fragments of the format Ln-n where n are numbers:

```
http://localhost:3000/workflows/1816a8c1-5a88-41a5-bcb4-65f9d763ec33/job-logs/reana-run-job-3634e846-c11d-42de-a015-7c4ebf4154ad#L157-2000
http://localhost:3000/workflows/1816a8c1-5a88-41a5-bcb4-65f9d763ec33/job-logs/reana-run-job-3634e846-c11d-42de-a015-7c4ebf4154ad#L157
```

Highlighting is done in yellow:
image
There also is a quick action button to copy the permanent link. 

Multiple lines can be selected using shift click.

To allow searching inside the now virtualized list, a custom ctrl+f handler is added that handles searching:
image

image

This search is only active, when the log viewer is focused. It tries to emulate the normal browser search. It allows to search for multiline strings, that are separated by spaces or newlines. One whitespace in the query can match multiple whitespace characters. E.g.
Text
```
Hello    World
Hello
World
```

The query:
```
Hello World
```
Would have two matches.

The query
```
Hello  World
```
Would have one match

The query
```
Hello     World
```
Would have zero matches because there are more than 4 spaces in between the words.

Closes: #503

### DIFF
--- a/reana-ui/src/components/LogViewer.js
+++ b/reana-ui/src/components/LogViewer.js
@@ -1,0 +1,551 @@
+/*
+  -*- coding: utf-8 -*-
+
+  This file is part of REANA.
+  Copyright (C) 2026 CERN.
+
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+import {
+  useCallback,
+  useDeferredValue,
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+} from "react";
+import PropTypes from "prop-types";
+import { useLocation, useNavigate } from "react-router-dom";
+import { useVirtualizer } from "@tanstack/react-virtual";
+import { Icon } from "semantic-ui-react";
+import CopyButton from "./CopyButton";
+import LogSearchBar from "./LogSearchBar";
+import styles from "./LogViewer.module.scss";
+
+const ROW_HEIGHT = 28;
+const OVERSCAN = 30;
+const TAB_SIZE = 4;
+const COPY_CHECK_TIMEOUT = 750;
+const MATCH_LIMIT = 10000;
+
+export function splitLogLines(logs) {
+  if (!logs) return [];
+
+  const lines = logs.replace(/\r\n?/g, "\n").split("\n");
+  if (lines.at(-1) === "") lines.pop();
+  return lines;
+}
+
+export function parseLogFragment(hash, lineCount) {
+  const match = /^#L([1-9]\d*)(?:-([1-9]\d*))?$/.exec(hash);
+  if (!match) return null;
+
+  const first = Number(match[1]);
+  const second = Number(match[2] || match[1]);
+  if (first > second) return null;
+
+  return second <= lineCount ? { start: first, end: second } : null;
+}
+
+function getRenderedLineLength(line) {
+  return line.replaceAll("\t", " ".repeat(TAB_SIZE)).length;
+}
+
+// Compile a plain search query into a regular expression. Regex metacharacters
+// are escaped (the search is plain-text, not regex). Each run of whitespace in
+// the query becomes `\s{k,}` (k = its length), so a query matches across line
+// breaks and indentation as long as the gap holds *at least* as many whitespace
+// characters as were typed — newlines count as one each, and typing more spaces
+// requires a larger gap (`\s{1,}` is just today's `\s+`). Returns `null` for an
+// empty query.
+export function buildSearchRegex(
+  query,
+  { caseSensitive = false, wholeWord = false } = {},
+) {
+  if (!query || !query.trim()) return null;
+
+  let pattern = query
+    .replace(/[.*+?^${}()|[\]\\]/g, "\\$&")
+    .replace(/\s+/g, (run) => `\\s{${run.length},}`);
+  if (wholeWord) pattern = `\\b${pattern}\\b`;
+
+  try {
+    return new RegExp(pattern, caseSensitive ? "g" : "gi");
+  } catch {
+    return null;
+  }
+}
+
+// Scan the full log text for all (non-overlapping) matches of `regex`, returning
+// their `[start, end)` character offsets. Capped at `limit` to keep memory and
+// rendering bounded for short queries that match very often.
+export function findMatches(text, regex, limit = MATCH_LIMIT) {
+  if (!regex) return { ranges: [], capped: false };
+
+  const ranges = [];
+  regex.lastIndex = 0;
+  let match;
+  while ((match = regex.exec(text)) !== null) {
+    if (match[0] === "") {
+      // Guard against zero-length matches looping forever.
+      regex.lastIndex += 1;
+      continue;
+    }
+    ranges.push([match.index, match.index + match[0].length]);
+    if (ranges.length >= limit) {
+      return { ranges, capped: true };
+    }
+  }
+  return { ranges, capped: false };
+}
+
+// Prefix-sum of character offsets where each line begins in `lines.join("\n")`.
+// Line `i` occupies `[offsets[i], offsets[i] + lines[i].length)`.
+export function computeLineOffsets(lines) {
+  const offsets = new Array(lines.length);
+  let acc = 0;
+  for (let i = 0; i < lines.length; i++) {
+    offsets[i] = acc;
+    acc += lines[i].length + 1; // +1 for the "\n" joining the lines
+  }
+  return offsets;
+}
+
+// Index of the line containing a given character offset (largest `i` with
+// `offsets[i] <= offset`), via binary search.
+export function offsetToLine(offsets, offset) {
+  let lo = 0;
+  let hi = offsets.length - 1;
+  let ans = 0;
+  while (lo <= hi) {
+    const mid = (lo + hi) >> 1;
+    if (offsets[mid] <= offset) {
+      ans = mid;
+      lo = mid + 1;
+    } else {
+      hi = mid - 1;
+    }
+  }
+  return ans;
+}
+
+// Split a single line into ordered fragments for rendering, intersecting the
+// (sorted, non-overlapping) global `matches` with this line's character range.
+// A match that crosses a line boundary highlights only the portion on this line.
+export function getLineSegments(line, lineStart, matches, activeIndex) {
+  if (!matches || matches.length === 0 || line.length === 0) {
+    return [{ text: line, highlight: "none" }];
+  }
+
+  const lineEnd = lineStart + line.length;
+
+  // First match whose end falls within (or after) this line.
+  let lo = 0;
+  let hi = matches.length;
+  while (lo < hi) {
+    const mid = (lo + hi) >> 1;
+    if (matches[mid].end > lineStart) hi = mid;
+    else lo = mid + 1;
+  }
+
+  const segments = [];
+  let cursor = 0;
+  for (let i = lo; i < matches.length && matches[i].start < lineEnd; i++) {
+    const localStart = Math.max(0, matches[i].start - lineStart);
+    const localEnd = Math.min(line.length, matches[i].end - lineStart);
+    if (localEnd <= localStart) continue;
+
+    if (localStart > cursor) {
+      segments.push({
+        text: line.slice(cursor, localStart),
+        highlight: "none",
+      });
+    }
+    segments.push({
+      text: line.slice(localStart, localEnd),
+      highlight: i === activeIndex ? "current" : "match",
+    });
+    cursor = localEnd;
+  }
+
+  if (segments.length === 0) return [{ text: line, highlight: "none" }];
+  if (cursor < line.length) {
+    segments.push({ text: line.slice(cursor), highlight: "none" });
+  }
+  return segments;
+}
+
+// New horizontal scroll offset so a match (at `markLeft`..`markLeft+markWidth`
+// in scroll coordinates) is brought into view with a `margin` of context.
+// Returns the current `scrollLeft` unchanged when the match is already visible.
+export function revealHorizontalOffset(
+  scrollLeft,
+  clientWidth,
+  markLeft,
+  markWidth,
+  margin = 80,
+) {
+  const markRight = markLeft + markWidth;
+  if (markLeft < scrollLeft + margin) {
+    return Math.max(0, markLeft - margin);
+  }
+  if (markRight > scrollLeft + clientWidth - margin) {
+    return Math.max(0, markRight - clientWidth + margin);
+  }
+  return scrollLeft;
+}
+
+export default function LogViewer({ logs, className = "" }) {
+  const location = useLocation();
+  const navigate = useNavigate();
+  const containerRef = useRef(null);
+  const viewportRef = useRef(null);
+  const selectionAnchorRef = useRef(null);
+  const skipNextSelectionScrollRef = useRef(false);
+  const searchInputRef = useRef(null);
+  const hScrolledForRef = useRef(-1);
+  const lines = useMemo(() => splitLogLines(logs), [logs]);
+  const selection = useMemo(
+    () => parseLogFragment(location.hash, lines.length),
+    [location.hash, lines.length],
+  );
+  const permalink = `${window.location.origin}${location.pathname}${location.search}${location.hash}`;
+  const lineNumberWidth = String(lines.length).length;
+  const canvasWidth = useMemo(() => {
+    const longestLine = lines.reduce(
+      (longest, line) => Math.max(longest, getRenderedLineLength(line)),
+      0,
+    );
+    return `calc(${lineNumberWidth + longestLine + 5}ch + 32px)`;
+  }, [lineNumberWidth, lines]);
+
+  const [searchOpen, setSearchOpen] = useState(false);
+  const [query, setQuery] = useState("");
+  const [caseSensitive, setCaseSensitive] = useState(false);
+  const [wholeWord, setWholeWord] = useState(false);
+  const [activeMatch, setActiveMatch] = useState(0);
+  const deferredQuery = useDeferredValue(query);
+  // Only search while the bar is open, so closing it clears the highlights and
+  // avoids computing matches in the background.
+  const effectiveQuery = searchOpen ? deferredQuery : "";
+
+  const text = useMemo(() => lines.join("\n"), [lines]);
+  const lineOffsets = useMemo(() => computeLineOffsets(lines), [lines]);
+
+  const { matches, capped } = useMemo(() => {
+    const regex = buildSearchRegex(effectiveQuery, {
+      caseSensitive,
+      wholeWord,
+    });
+    const { ranges, capped } = findMatches(text, regex);
+    return {
+      matches: ranges.map(([start, end]) => ({
+        start,
+        end,
+        startLine: offsetToLine(lineOffsets, start),
+      })),
+      capped,
+    };
+  }, [text, lineOffsets, effectiveQuery, caseSensitive, wholeWord]);
+
+  const activeIndex =
+    matches.length > 0 ? Math.min(activeMatch, matches.length - 1) : -1;
+  const activeStartLine =
+    activeIndex >= 0 ? matches[activeIndex].startLine : -1;
+
+  const virtualizer = useVirtualizer({
+    count: lines.length,
+    getScrollElement: () => viewportRef.current,
+    estimateSize: () => ROW_HEIGHT,
+    getItemKey: (index) => index + 1,
+    overscan: OVERSCAN,
+  });
+
+  useEffect(() => {
+    selectionAnchorRef.current = null;
+  }, [logs]);
+
+  const lastScrolledHash = useRef(null);
+
+  useEffect(() => {
+    if (selection === null) {
+      selectionAnchorRef.current = null;
+      lastScrolledHash.current = null;
+    }
+  }, [selection]);
+
+  useEffect(() => {
+    if (!selection) return;
+
+    selectionAnchorRef.current = selection.start;
+    if (skipNextSelectionScrollRef.current) {
+      skipNextSelectionScrollRef.current = false;
+      lastScrolledHash.current = location.hash;
+      return;
+    }
+    if (lastScrolledHash.current === location.hash) return;
+    lastScrolledHash.current = location.hash;
+    virtualizer.scrollToIndex(selection.start - 1, { align: "start" });
+  }, [selection, virtualizer, location.hash]);
+
+  // Reset to the first match whenever the query or matching options change.
+  useEffect(() => {
+    setActiveMatch(0);
+    hScrolledForRef.current = -1;
+  }, [effectiveQuery, caseSensitive, wholeWord]);
+
+  // Keep the active match scrolled into view (vertically).
+  useEffect(() => {
+    if (!searchOpen || activeIndex < 0) return;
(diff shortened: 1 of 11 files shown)


reana-ui/src/components/LogViewer.js
```diff
@@ -0,0 +1,198 @@
...
+            search: location.search,
+          },
+          { preventScrollReset: true },
+        );
+        return;
+      }
+
+      let start = lineNumber;
+      let end = lineNumber;
+      if (event.shiftKey && selectionAnchorRef.current) {
```
michaelbuchar, line 400: The selection anchor looks like it can survive a log switch and then get reused in the next log. shift-click here reads from the existing anchor ref, but that ref does not seem to be cleared when the active job/service changes and the hash is reset. That means a user can select a line in one log, switch to a different log, and have the first shift-click generate a range anchored to the previous log instead of starting fresh in the new one. Could we reset the anchor whenever the viewed log changes, or when the current fragment/hash is cleared, so range selection cannot span across two different log sources?

reana-ui/src/components/__tests__/LogViewer.test.js
```diff
@@ -0,0 +1,163 @@
...
+      screen.getByLabelText("Clear log line selection from line 3")
+        .parentElement,
+    ).toHaveClass("selected");
+    expect(container.querySelectorAll(".selected")).toHaveLength(2);
+    await waitFor(() =>
+      expect(mockScrollToIndex).toHaveBeenCalledWith(1, { align: "start" }),
+    );
+  });
+
+  it("updates the fragment while preserving the query string", async () => {
```
michaelbuchar, line 282: Nice coverage for the fragment parsing and selection behavior. One thing still missing in my mind is a route level check that the existing URL driven job/service selection still works when a #L... fragment is present. These tests currently exercise LogViewer in isolation, so a regression in the WorkflowLogs routing/normalisation path would still slip through. Could we add one integration-style test for `/job-logs/:job#L42` and one for `/service-logs/:component#L42`?

michaelbuchar:
The overall direction looks great to me. A dedicated `LogViewer` for stable `#L...` anchors, highlighting and copyable permalinks feels like the right way to implement #503! The remaining work seems to be  just cleanup and coverage in my eyes. One minor note as well: please keep the commit header in the same style we already use in this repo and avoid custom/vague ones like feature(reana-ui) since they do not read well in release notes. Something like `feat(workflow-details): add permanent line anchors for logs` could fit better. Thanks a lot! :)

reana-ui/src/components/__tests__/LogViewer.test.js
Sohn123: I added tests for the `WorkflowLogs` component that tests now that the fragments correctly trigger a selection of the lines by calling into tanstack virtual. Is that what you had in mind or something even more general?

reana-ui/src/components/LogViewer.js
Sohn123: The underlying problem was causing the issue, was that the component was not updated if multiple jobs with the exact same log existed. I added now the job-id as a key attribute to force a rerender of the LogViewer component, which should prevent the issue.

Sohn123:
Hey @michaelbuchar ,
I added the tests for better coverage. Is that what you imagined/wanted to be tested? The issue with the reusage of a stale reference is gone now :)

reana-ui/src/pages/workflowDetails/components/__tests__/WorkflowLogs.test.js
```diff
@@ -0,0 +1,158 @@
...
+    </MemoryRouter>,
+  );
+}
+
+beforeEach(() => {
+  mockScrollToIndex.mockClear();
+  mockDispatch.mockClear();
+});
+
+test("job-logs/:job#L42 — correct job is selected and line 42 is highlighted", async () => {
```
michaelbuchar, line 183: Nice addition, thank you! This is much closer to the route level coverage I had in mind. I still have one minor gap in mind though: the fixtures only contain a single job/component, so these tests would still pass even if WorkflowLogs ignored :job or :component and always rendered the first available entry. Could we make the fixtures contain at least two different entries and then assert that the URL selects the expected one? That would fully prove the routing part too.

Sohn123: Sure I will add that! I already asked Tibor for the claude review, so we can get that in as well.

michaelbuchar:
I noticed one small bug outside the visible diff hunk in WorkflowLogs.js: in the URL sync effect, this still calls selectedJobId(nextId) instead of setSelectedJobId(nextId). Since selectedJobId is the current state value, a later :job route change would end up trying to call a string as a function. That would break the URL driven job selection flow we want to preserve alongside the #L... fragment handling.

reana-ui/src/components/LogViewer.js
```diff
@@ -0,0 +1,203 @@
...
+
+  useEffect(() => {
+    if (!selection) return;
+
+    selectionAnchorRef.current = selection.start;
+    if (skipNextSelectionScrollRef.current) {
+      skipNextSelectionScrollRef.current = false;
+      return;
+    }
+    virtualizer.scrollToIndex(selection.start - 1, { align: "start" });
```
tiborsimko, line 290: **🤖 Polling-driven log refresh snaps the viewport back to the selected line (MEDIUM)**

In `reana-ui/src/components/LogViewer.js`, the selection effect re-fires whenever the `selection` memo's identity changes, and `selection` is recomputed every time `lines.length` changes:

```js
const lines = useMemo(() => splitLogLines(logs), [logs]);
const selection = useMemo(
  () => parseLogFragment(location.hash, lines.length),
  [location.hash, lines.length],
);
...
useEffect(() => {
  if (!selection) return;
  selectionAnchorRef.current = selection.start;
  if (skipNextSelectionScrollRef.current) {
    skipNextSelectionScrollRef.current = false;
    return;
  }
  virtualizer.scrollToIndex(selection.start - 1, { align: "start" });
}, [selection, virtualizer]);
```

The `skipNextSelectionScrollRef` flag is only set by the user's own line clicks, not by external updates to the `logs` prop. And `logs` *does* update externally while the workflow is non-finished — `reana-ui/src/pages/workflowDetails/WorkflowDetails.js:127–138` polls on a timer:

```js
const refetchWorkflow = useCallback(() => {
  const options = { refetch: true, showLoader: false };
  dispatch(fetchWorkflow(workflowId, options));
  dispatch(fetchWorkflowLogs(workflowId, options));
}, [dispatch, workflowId]);

useEffect(() => {
  if (!interval.current && pollingSecs) {
    interval.current = setInterval(refetchWorkflow, pollingSecs * 1000);
  }
  return cleanPolling;
}, [dispatch, refetchWorkflow, workflowId, pollingSecs]);
```

So whenever a poll returns a different line count while the workflow runs, `lines.length` changes → `selection` is a new memo object → the effect runs without the skip flag → `virtualizer.scrollToIndex` pulls the viewport back to the original line. A user who follows a permalink to `#L42`, then scrolls down to read context, gets snapped back to line 42 on whichever polls bring fresh content. The new shareable-link feature is the most disrupted on exactly the workflows it matters for: running ones being actively debugged.

Suggested fix — gate the scroll on the URL fragment actually changing, not on `selection` object identity. One way is to remember the last hash that triggered a scroll:

```js
const lastScrolledHash = useRef(null);
useEffect(() => {
  if (!selection) return;
  selectionAnchorRef.current = selection.start;
  if (skipNextSelectionScrollRef.current) {
    skipNextSelectionScrollRef.current = false;
    lastScrolledHash.current = location.hash;
    return;
  }
  if (lastScrolledHash.current === location.hash) return;
  lastScrolledHash.current = location.hash;
  virtualizer.scrollToIndex(selection.start - 1, { align: "start" });
}, [selection, virtualizer, location.hash]);
```

Polling-driven `lines.length` changes then leave `location.hash` untouched, so no involuntary scroll. The initial deeplink and explicit hash changes still scroll as before.

reana-ui/src/components/LogViewer.js
```diff
@@ -0,0 +1,203 @@
...
+  const lineNumberWidth = String(lines.length).length;
+  const canvasWidth = useMemo(() => {
+    const longestLine = lines.reduce(
+      (longest, line) => Math.max(longest, getRenderedLineLength(line)),
+      0,
+    );
+    return `calc(${lineNumberWidth + longestLine + 5}ch + 32px)`;
+  }, [lineNumberWidth, lines]);
+
+  const virtualizer = useVirtualizer({
```
tiborsimko, line 258: **🤖 Browser Find (Cmd/Ctrl+F) no longer matches text outside the visible window (MEDIUM)**

Switching from `CodeSnippet` to the virtualised `LogViewer` removes a capability users rely on for log investigation. The previous component rendered the entire log into the DOM, so `Cmd+F` searched every line. `LogViewer` uses `@tanstack/react-virtual` and only mounts visible rows plus 10 overscan:

```js
const virtualizer = useVirtualizer({
  count: lines.length,
  getScrollElement: () => viewportRef.current,
  estimateSize: () => ROW_HEIGHT,
  getItemKey: (index) => index + 1,
  overscan: OVERSCAN,
});
...
{virtualizer.getVirtualItems().map((virtualLine) => { ... })}
```

For a 10,000-line job log, browser Find sees only ~30 rows. Searching for an error string — the main reason a user opens this view — silently returns "Phrase not found" unless the user has already scrolled near the match. Select-all + copy is truncated the same way. There is no in-viewer search to compensate.

Suggested fix — there are two complementary changes:

1. **Built-in search** (robust fix for all log sizes). Add a search input that walks the `lines` array and uses `virtualizer.scrollToIndex` to bring each match into view. Once content is virtualised the DOM is necessarily incomplete, so only an in-viewer search can deliver "find any text in this log" for large logs.
2. **Virtualisation threshold** (mitigation for the common case). Skip virtualisation when the log is small enough that the prior DOM-renders-everything behaviour was already cheap:

   ```js
   const SHOULD_VIRTUALIZE = lines.length > 5_000;
   ```

   This restores native Cmd+F for typical workflow logs. Pick the threshold deliberately — even a non-virtualised `LogViewer` is heavier than the previous `CodeSnippet` (per-line `<div>`s, anchors, selection styling, instead of a single text node), so this is not a "rendering N divs is free" assumption. Large logs remain dependent on (1).

A duplicate hidden `<pre>` would seem to restore Cmd+F for free, but introduces ambiguous Find scroll targets and duplicate selection on Cmd+A — better avoided.

reana-ui/src/components/LogViewer.js
```diff
@@ -0,0 +1,203 @@
...
+              className={`${styles.line} ${isSelected ? styles.selected : ""}`}
+              key={virtualLine.key}
+              style={{
+                height: `${virtualLine.size}px`,
+                transform: `translateY(${virtualLine.start}px)`,
+              }}
+            >
+              {selection?.start === lineNumber && (
+                <CopyButton
+                  as={Icon}
```
tiborsimko: **🤖 Permalink copy icon inherits `CopyButton`'s `as={Icon}` keyboard-accessibility gap (LOW)**

In `reana-ui/src/components/LogViewer.js`, the permalink trigger uses the same icon-as-trigger pattern that `CodeSnippet` already uses:

```jsx
{selection?.start === lineNumber && (
  <CopyButton
    as={Icon}
    text={permalink}
    icon="linkify"
    className={styles["copy-action"]}
    aria-label="Copy permalink to selected log lines"
    timeout={COPY_CHECK_TIMEOUT}
    dismissOnScroll={true}
  />
)}
```

When `CopyButton` is given `as={Icon}`, the rendered DOM is a bare Semantic UI `<i>` element — no `role`, no `tabIndex`, no keyboard activation. Tab order skips it, so a keyboard-only user can reach the line-number `<a>` (which correctly toggles selection on Enter) but cannot reach the actual copy control to obtain the permalink. The `aria-label` labels something that assistive tech can never focus.

The underlying `as={Icon}` pattern already exists in `CodeSnippet`, but this PR introduces a new permalink action using that inaccessible pattern. Worth fixing now rather than spreading it further.

Suggested fix — keep the `linkify` glyph but make the trigger a real button:

```jsx
<CopyButton
  as={Button}
  text={permalink}
  icon="linkify"
  basic
  className={styles["copy-action"]}
  aria-label="Copy permalink to selected log lines"
  ...
/>
```

`Tab` then lands on it and `Enter`/`Space` activate it, restoring parity with the line-number anchor. Migrating the existing `CodeSnippet` usage to the same shape would be a fair follow-up.

reana-ui/src/components/LogViewer.js
```diff
@@ -0,0 +1,203 @@
...
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+import { useCallback, useEffect, useMemo, useRef } from "react";
+import PropTypes from "prop-types";
+import { useLocation, useNavigate } from "react-router-dom";
+import { useVirtualizer } from "@tanstack/react-virtual";
+import { Icon } from "semantic-ui-react";
+import { CopyButton } from "~/components";
```
tiborsimko: **🤖 `LogViewer` imports `CopyButton` through the `~/components` barrel that re-exports `LogViewer` (NIT)**

`reana-ui/src/components/LogViewer.js`:

```js
import { CopyButton } from "~/components";
```

`reana-ui/src/components/index.js`:

```js
export { default as CopyButton } from "./CopyButton";
export { default as LogViewer } from "./LogViewer";
```

This is a circular import. ES module live bindings (and Babel's getter-based emulation) resolve it at render time, so reordering the barrel would not actually break it — the previous version of this comment overstated the risk. The reason to fix it is just clarity and convention: nothing else in this folder pulls a sibling through the parent barrel, and avoiding the cycle removes a small surprise for future readers.

Suggested fix — import the sibling directly:

```js
import CopyButton from "./CopyButton";
```

That matches what the other components in this folder already do for intra-folder dependencies.

reana-ui/src/components/LogViewer.js
Sohn123: I see that this is an issue with virtualization, but especially for long logs, performance was otherwise really bad. I would propose to add a the built in search and link it directly to the ctrl+f shortcut, so when someones hits control f on a page with a LogViewer component you search within the log. Is that acceptable? I think only being able to search in short logs is not acceptable

Sohn123: @tiborsimko I implemented now the version where the ctrl+f is hooked to allow full text search within the logs which has similar UI/UX to the normal browser search. I think this is an acceptable workaround for supporting large logs.

reana-ui/src/components/LogViewer.js
Sohn123: This would break the UI and require extensive style overwrites which I would rather not do and rather accept the not perfect accessibility.

reana-ui/src/components/LogViewer.js
```diff
@@ -0,0 +1,526 @@
...
+    count: lines.length,
+    getScrollElement: () => viewportRef.current,
+    estimateSize: () => ROW_HEIGHT,
+    getItemKey: (index) => index + 1,
+    overscan: OVERSCAN,
+  });
+
+  useEffect(() => {
+    selectionAnchorRef.current = null;
+  }, [logs]);
```
michaelbuchar, line 268: The stale-anchor fix looks closer now, but I think one edge case is still open. The anchor gets reset when the logs prop changes, but it does not get cleared when the #L... fragment disappears through browser back/forward or another external URL update. In that case the next shift-click can still reuse the old anchor and select the wrong range. Could we also reset the stored anchor when the hash is removed, not only when the log content changes?

reana-ui/src/components/LogViewer.js
```diff
@@ -0,0 +1,526 @@
...
+                  <CopyButton
+                    as={Icon}
+                    text={permalink}
+                    icon="linkify"
+                    className={styles["copy-action"]}
+                    aria-label="Copy permalink to selected log lines"
+                    timeout={COPY_CHECK_TIMEOUT}
+                    dismissOnScroll={true}
+                  />
+                )}
```
michaelbuchar, line 493: Nice improvement on the permalink action, but I do not think the accessibility point is fully closed yet. It still renders as as={Icon}, so even with the aria-label it is not part of the normal tab order and cannot be activated from the keyboard like a real button or link. Could we switch this to a semantic interactive element so keyboard users can reach and trigger it as well?

reana-ui/src/components/LogViewer.js
```diff
@@ -0,0 +1,526 @@
...
+  useEffect(() => {
+    if (!selection) return;
+
+    selectionAnchorRef.current = selection.start;
+    if (skipNextSelectionScrollRef.current) {
+      skipNextSelectionScrollRef.current = false;
+      lastScrolledHash.current = location.hash;
+      return;
+    }
+    if (lastScrolledHash.current === location.hash) return;
```
michaelbuchar, line 288: The repeat-scroll guard still looks a bit sticky here. lastScrolledHash is updated when a fragment is selected, but it is not cleared when the hash is removed by clicking the selected line again. Then if the user navigates back to the same #L... fragment in the same tab, the selection is restored but this early return skips scrollToIndex(...), so the target line may stay off-screen. Could we reset lastScrolledHash when the hash is cleared, so reopening the same fragment still scrolls correctly?

reana-ui/src/components/LogViewer.js
Sohn123: I added a handler for resetting the location hash fixing both issues.

reana-ui/src/components/LogViewer.js
Sohn123: I replaced it with a button keeping the same styling.

reana-ui/src/components/LogViewer.js
Sohn123: I added resetting the previous hash to the hash reset handler.

reana-ui/src/components/LogViewer.js
```diff
@@ -0,0 +1,551 @@
...
+    estimateSize: () => ROW_HEIGHT,
+    getItemKey: (index) => index + 1,
+    overscan: OVERSCAN,
+  });
+
+  useEffect(() => {
+    selectionAnchorRef.current = null;
+  }, [logs]);
+
+  useEffect(() => {
```
michaelbuchar, line 272: Nice improvement! this closes the empty hash case. I think there is still one edge case left though.. the reset only runs when the fragment becomes an empty string, not when it changes to a non-empty invalid fragment such as #foo or #L999. In that situation parseLogFragment() returns null, but selectionAnchorRef and lastScrolledHash can still keep the previous valid state. The next shift click may then reuse the stale anchor, and revisiting the same valid #L... fragment can still skip scrollToIndex(). Could we reset the stored anchor/hash whenever the fragment is missing or invalid, not only when it is exactly empty?

Sohn123: Good catch! I changed it now to reset on selection instead of so it respects the results of the hash parsing and resets every time there is an empty/invalid selection

michaelbuchar (dismissed):
LGTM, moving to in-merge! 🚀

reana-ui/src/components/LogViewer.js
```diff
@@ -0,0 +1,551 @@
...
+    if (selection === null) {
+      selectionAnchorRef.current = null;
+      lastScrolledHash.current = null;
+    }
+  }, [selection]);
+
+  const lastScrolledHash = useRef(null);
```
tiborsimko: **🤖 `lastScrolledHash` ref declared after the effect that already uses it (NIT)**

`reana-ui/src/components/LogViewer.js:270`-`277` (the `selection === null` effect and the ref declaration just below it)

```js
useEffect(() => {
  if (selection === null) {
    selectionAnchorRef.current = null;
    lastScrolledHash.current = null;   // used here…
  }
}, [selection]);

const lastScrolledHash = useRef(null);  // …declared only here
```

This works at runtime — the effect closure only reads `lastScrolledHash` when it fires, which is after the render that initialises the ref — and it is not a Rules-of-Hooks problem either, since the `useRef` is unconditional and keeps its position in the hook order. But the out-of-order declaration is a needless surprise: every other ref (`containerRef`, `viewportRef`, `selectionAnchorRef`, `skipNextSelectionScrollRef`, `searchInputRef`, `hScrolledForRef`) is grouped at the top of the component, and a reader scanning that cluster would not expect another ref wedged between two effects further down.

Suggested fix — move `const lastScrolledHash = useRef(null);` up with the other `useRef` declarations (the cluster ending at `hScrolledForRef`, `LogViewer.js:208`), so all refs are declared together and before any effect references them.

tiborsimko (commented):
**🤖 Modified files missing the 2026 copyright bump (NIT)**

`reana-ui/src/components/index.js:5` and `reana-ui/src/pages/workflowDetails/components/WorkflowLogs.js:5`

Both files are modified by this PR (`index.js` gains the `LogViewer` export; `WorkflowLogs.js` swaps `CodeSnippet` → `LogViewer` and reworks the hash handling) but neither header includes the current year. Per the project convention — the current year plus any years missed by earlier contributors, verifiable with `git log --format=%ad --date=format:%Y -- <file> | sort -u`:

- `index.js`: history is `2020 2021 2022 2023 2024 2026`, header stops at `2020, 2021, 2022, 2023, 2024 CERN.` — missing **2026**.
- `WorkflowLogs.js`: history is `2019 2020 2022 2023 2024 2025 2026`, header reads `2020, 2022, 2025 CERN.` — missing **2019, 2023, 2024, 2026**.

(The six new files added in this PR all carry `2026` correctly, and `CopyButton.js` is fine at `2026`; these two are the only modified source files with stale headers.)

Suggested fix — update both headers, preserving each file's existing indentation (`index.js` uses spaces, `WorkflowLogs.js` uses a leading tab):

- `index.js` → `Copyright (C) 2020, 2021, 2022, 2023, 2024, 2026 CERN.`
- `WorkflowLogs.js` → `Copyright (C) 2019, 2020, 2022, 2023, 2024, 2025, 2026 CERN.`

---

**🤖 `Closes:` in the commit body diverges from the repo's issue-reference style
(NIT)**

_(Commit message body — not anchorable to a diff line; raise as a PR comment or
fix on the next amend.)_

The commit body closes the issue with a colon:

```
Closes: #503
```

The repository consistently uses the colon-less form. Across the last 500
commits:

- `Closes #N` / `closes #N` (no colon): **158** occurrences
- `Closes: #N` (with colon): **1** — this commit

So `Closes: #503` is the lone outlier. Keeping the established form also keeps
the GitHub auto-close linkage and release-note scanning consistent with the rest
of the history.

Suggested fix — drop the colon on the next amend/force-push:

```
Closes #503
```

Sohn123:
@tiborsimko I updated the body of the commit and removed the colon + moved the definition of the reference

tiborsimko (approved):

PR #514, opened 2026-06-09
feat(workflow-list): rework the workflow filtering UI

Sohn123:
This branch contains multiple versions of an updated UI for the workflow list. Each commit contains a working version to check out and play around. I listed them all in the order of the commits:

## Adding a second row for filtering shared workflows

image

## Two rows separated into coarse and fine filtering

image

## Move advanced filters to a sidebar

image

## Add tabs to navigate between shared/own workflows

image

##  Keep the tabs but remove the header, so tab navigation is now on the highest level

image

After scrolling:

image

## Make the tabs part of the sticky header, so they are always visible

image

After scrolling:

image

## Replace dropdowns with radio buttons and use inline dropdowns

image


CameronMcClymont:
### Brain dump of random thoughts

- I'm in favour of moving the workflow sharing tabs into the facets list, and then it probably makes sense to keep the 'Your workflows' title. In this case:
  - Have a top-level 'all workflows'
  - Indent 'your workflows' and 'shared with you' since these are subsets of 'all workflows'
  - Move 'shared by you' out to be a checkbox alongside the 'open sessions only' and 'show deleted runs' options?
- If we did want to keep the workflow sharing filters as tabs, then maybe it makes sense to swap the position of the tabs and the search bar so the tabs are closer to the actual workflows?
- Change 'any status' facet to just 'any'
- Rename everything to use 'your' language (e.g. 'your workflows' instead of 'my workflows')
- Rename 'show deleted runs' to 'include deleted runs' and make it a switch instead of a checkbox to differentiate it from 'open sessions only' and 'shared by you' which both remove workflows whereas 'include deleted runs' adds workflows?
- Rename 'open sessions only' to 'with open sessions'?
- Add a 'workflow type' facet (snakemake, yadage, cwl, serial)
- Include the number of results for each facet (e.g. Running (4), Finished (21), ...)

Just a bunch of random thoughts take/leave what you want.

Sohn123:
Ok the main points I took away from the discussion today:
- Tabs are ok if they allow to select a distinct subset of workflows
- We want to try out designs that don't have tabs to not feature sharing as prominently
- The current checkbox design for including/excluding deleted and only showing workflows with active sessions is not intuitive
- If easily possible we would display how many results would be there if clicking a certain filter

Unfortunately with our current backend, displaying the number of items is not easily possible, so I would move this to a separate PR/Issue if we want to do that.

Based on that I propose the following new designs:

## Two mutually exclusive tabs

image

image

image


## Sharing as filter option in the sidebar

image

image


## Selection instead of checkbox

image


## New Side by Side toggles

image

image


For the options without the tabs, I would add the header back again to the page, but didn't do that yet

CameronMcClymont:
My layout preference is the first set of screenshots (Two mutually exclusive tabs). I do like the side-by-side toggles, but maybe we should stay consistent with the other filters and keep them vertical?

One unrelated thing: in the sorting dropdown, we have 'latest first' and 'oldest first'. Maybe we should reword 'latest' to 'newest' since it's a more direct antonym to 'oldest' (whereas the opposite of 'latest' is 'earliest')?

Sohn123:
@CameronMcClymont thanks for the feedback :) I first had it as two vertical buttons, but his felt like a waste of space, let's see what the others say.

A last think I'd like to propose is to make the pagination bar sticky as well + allow the filter categories on the left to be collapsible:

image

But this sticky bottom bar could of course also be combined with the mutually exclusive tabs. Same goes for the collapsible sections in the side bar.

Following decisions we need to take:
- Tabs for sharing or filtering in the sidebar
- Sticky bottom bar for pagination or not sticky as previously
- Should the sections be collapsible in the sidebar
- Do we want to have a page header like "Your workflows"

Sohn123:
After the discussion I implemented the changes now, and this the current design:

# Default Page
image

# No workflows found

image

# Multiple selected filters
image

# Loading state with previous workflows
image

# Loading state with no results

image

# Shared workflows

image

CameronMcClymont:
Nice! It looks like lots of good improvements, some comments and thoughts going from strongest opinion to least strong haha:

- We could move the 'clear filters' button away from the sorting and closer to the rest of the filters so they are logically connected and to prevent layout shift when the button appears/disappears as filters are changed
- Could the 'shared workflows' tab be renamed to 'shared with you' (like e.g. in Google Drive) or something to make it clear that these aren't workflows the user has shared with others, but rather that others have shared with the user?
- We could swap the positions of the 'show x per page' and the pagination buttons which I think is more conventional? At least this was closer to the previous layout and is the more typical layout I've seen in the past, e.g. the open data portal: https://opendata.cern.ch/search?q=&f=experiment%3ALHCb&l=list&order=desc&p=1&s=10&sort=mostrecent
- For the sharing filters in the 'your workflows' tab, we could indent the 'private' and 'shared' options so they are nested under 'all workflows' which visually signifies that they are subsets of 'all workflows' (even if this is obvious from reading the text)
- In the status filters, we could probably write 'any' instead of 'any status' since 'status' is already the category so it's redundant?

That's all I can think of! Nice work

michaelbuchar:
Nice direction overall! My main concern with the latest version is the discoverability of the cluster health indicator (and the docs links). In the current layout, they end up far enough down the page that users have to scroll quite a bit before finding them, which makes both easy to miss. Since the cluster health is an important reference points, I think it would work better closer to the top of the page (ideally in a more immediately visible area).

reana-ui/src/pages/workflowList/components/WorkflowRefinementMenu.js
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <Menu
+      secondary
+      vertical
+      fluid
+      className={styles.refinementMenu}
+      aria-label={ariaLabel}
+    >
+      {options.map((option) => (
+        <Menu.Item
```
tiborsimko, line 31: **🤖 PR514-01: New tabs and filters are not keyboard operable and expose no selected state (MEDIUM)**

The new category tabs and refinement menus use Semantic UI `Menu.Item`
components with `onClick` handlers. Semantic UI renders this combination as
an `<a>` without an `href`, `tabIndex`, keyboard handler, or selection role.
Consequently, the category, status, sharing, session, and deleted-run choices
cannot be reached or activated from the keyboard. Their selected state is
represented only by the `active` CSS class, so assistive technology cannot
announce which choice is applied.

On `master`, the corresponding filters were Semantic UI `Dropdown` and
`Checkbox` components, which are focusable and expose their state. This is
therefore a regression introduced by the rework rather than a pre-existing
gap. For example, the new status menu renders as:

```html
<div aria-label="Filter by status" class="ui fluid secondary vertical menu">
  <a class="active item"><i class="circle outline icon"></i><span>Any</span></a>
  <a class="item"><span>running</span></a>
</div>
```

The `aria-label` is on a `<div>` with no role, so it is also unlikely to be
exposed by the accessibility tree. The current tests pass because Testing
Library's `getByLabelText` is more permissive here than a browser's
accessibility tree.

There are two related gaps in the new chrome. A collapsed active section shows
an empty `activeDot` span with no accessible text, and the mobile stylesheet
sets `.refreshLabel` to `display: none`, reducing the refresh button's
accessible name to a bare timestamp.

Suggested fix — render the refinement choices as real buttons with radio-group
semantics (`role="radiogroup"`, `role="radio"`, and `aria-checked`), and expose
the category switcher as links or a proper tab list with `aria-selected`.
Announce active collapsed sections in the toggle's accessible name and give the
refresh button an explicit `aria-label` that remains descriptive on mobile.
Update the tests to exercise keyboard activation and assert the ARIA-selected
state rather than Semantic UI's internal `active` class.

reana-ui/src/pages/workflowList/WorkflowList.js
```diff
@@ -136,89 +151,134 @@ function Workflows() {
...
+    ? WORKFLOW_LIST_PAGE_SIZE_OPTIONS
+    : [
+        ...WORKFLOW_LIST_PAGE_SIZE_OPTIONS,
+        {
+          key: pageSize,
+          text: `${pageSize}`,
+          value: pageSize,
+        },
+      ].sort((a, b) => a.value - b.value);
+  const firstResult = workflowsCount > 0 ? (page - 1) * pageSize + 1 : 0;
```
tiborsimko, line 166: **🤖 PR514-02: Out-of-range pages render a contradictory range and misleading pagination controls (LOW)**

The result range in `WorkflowList.js` is calculated from the requested page
without checking that the page is within `totalPages`:

```js
const firstResult = workflowsCount > 0 ? (page - 1) * pageSize + 1 : 0;
const lastResult =
  workflowsCount > 0 ? Math.min(page * pageSize, workflowsCount) : 0;
```

For example, `?page=999` with 25 workflows and a page size of 10 renders
"Showing 9981–25 of 25 workflows" together with the empty-state message. This
can also happen without editing the URL when a valid last page becomes invalid
after workflows are deleted, unshared, or removed by polling.

The pagination control receives the same invalid active page. It renders no
active numbered item, and its Previous control points to page 998, leading to
another empty page. First, the numbered pages, Next, and Last recover to valid
pages, but Previous—the natural recovery action—does not. Worse, when the total
is at or below the page size, `showPagination` is false and no pagination
control is rendered. With five workflows, the same URL therefore shows
"Showing 9981–5 of 5 workflows" beside "No workflows found" with no direct
pagination route back. Recovery then requires another control whose setter
resets the page, such as page size, sorting, or a filter, or editing the URL.

Suggested fix — once the response total is known, replace an out-of-range page
with the last valid page (or page 1) and refetch it. Until that normalisation
completes, do not render a contradictory range or pass an invalid active page
to `Pagination`. Add coverage for a hand-edited high page and for a result set
whose final page disappears.

reana-ui/src/pages/workflowList/WorkflowList.module.scss
```diff
@@ -2,86 +2,246 @@
...
+  gap: 2em;
+}
+
+.results {
   display: flex;
-  justify-content: space-between;
-  align-items: baseline;
+  flex-direction: column;
+  min-width: 0;
+  min-height: 100vh;
```
tiborsimko, line 46: **🤖 PR514-03: `min-height: 100vh` guarantees unnecessary desktop scrolling (LOW)**

In `WorkflowList.module.scss`, `.results` is at least one full viewport high on
desktop. The workflow category header and the application's top header sit
outside that column, so even a short or empty result set makes the complete
page taller than the viewport. The footer—and therefore the documentation and
cluster-health links—is always pushed below an otherwise unnecessary viewport
of scrolling.

The results body already has a viewport-relative minimum-height clamp to keep
short result sets stable, and the empty state duplicates the same clamp in a
second module. These constraints make the full `100vh` minimum on `.results`
unnecessary.

Suggested fix — remove `min-height: 100vh` from `.results` and let one shared
results-body minimum-height rule provide the desired short-page stability.
Verify the normal, empty, and loading layouts at desktop and tablet widths and
confirm that the footer becomes reachable without gratuitous scrolling.

reana-ui/src/pages/workflowList/workflowListQuery.js
```diff
@@ -158,7 +140,7 @@ export function serializeQueryToApiParams(query) {
     },
     search: query.search.trim() || undefined,
     status,
-    shared,
+    shared: false,
```
tiborsimko, line 143: **🤖 PR514-04: `?owned-by=anybody` silently changes result-set semantics (LOW)**

The released interface writes `?owned-by=anybody` when “Owned by anybody” is
selected. On `master`, the query serialiser turns this value into `shared=true`,
which returns the union of workflows owned by the current user and workflows
shared with them. The new serialiser instead sends `shared=false` with
`sharedBy="anybody"`, which returns only workflows shared with the current
user. Owned workflows therefore disappear from URLs bookmarked or shared from
the released interface.

The same narrowing applies to the documented legacy `?shared=true` URL. The
affected links are therefore not limited to an older historical parameter;
they include the URL emitted by the current interface.

Removing the combined view may be the intended consequence of adopting two
mutually exclusive categories. The concern is that the existing URLs remain
accepted while silently selecting a narrower result set.

Suggested fix — make the migration policy explicit. Either preserve the old
union semantics for existing `owned-by=anybody` and legacy `shared=true` URLs,
or document and test that the mapping is intentionally lossy. Whichever
behaviour is chosen, add a regression test that distinguishes owned, shared,
and union result sets.

reana-ui/src/pages/workflowList/components/WorkflowStatusFilter.js
```diff
@@ -9,66 +9,80 @@
...
-export default function WorkflowStatusFilters({
+export default function WorkflowStatusFilter({
   statusFilter,
   filter,
-  includeDeleted,
-  setIncludeDeleted,
   hasStatusFilter,
 }) {
   const value = hasStatusFilter ? statusFilter : undefined;
+  const [showMore, setShowMore] = useState(
```
tiborsimko, line 27: **🤖 PR514-05: “Fewer statuses” can hide the active advanced status (LOW)**

When an advanced status such as `pending` is active, `showMore` initially
reveals it. The user can nevertheless press “Fewer statuses”, which sets
`showMore` to false while the status prop remains unchanged. The active status
then disappears from the menu, “Any” remains inactive, and no visible option
shows which status is filtering the results.

```js
const [showMore, setShowMore] = useState(
  hasStatusFilter && MORE_STATUSES.includes(statusFilter),
);
useEffect(() => {
  if (hasStatusFilter && MORE_STATUSES.includes(statusFilter)) {
    setShowMore(true);
  }
}, [hasStatusFilter, statusFilter]);
```

The synchronising effect does not reopen the list because neither
`hasStatusFilter` nor `statusFilter` changed after the click.

Suggested fix — separate the user's expansion preference from the requirement
to display the active value, for example
`showMore = expanded || MORE_STATUSES.includes(value)`. An active
advanced status should remain visible until it is cleared or replaced. Add a
test that attempts to collapse the list while an advanced status is selected.

reana-ui/src/pages/workflowList/useWorkflowListQuery.js
```diff
@@ -183,6 +210,33 @@ export function useWorkflowListQuery() {
...
   );
 
+  const clearFilters = useCallback(() => {
+    setSearchText("");
+    updateParams(setSearchParams, (next) => {
+      next.delete("search");
+      next.delete("status");
+      next.delete("show-deleted");
+      next.delete("open-sessions");
+      if (next.get("owned-by")) {
```
tiborsimko, line 220: **🤖 PR514-06: View identity is conflated with `owned-by`, and clearing reads raw URL state (LOW)**

The top-level category is inferred from whether parsed `ownedBy` is truthy,
while `clearFilters()` independently infers it from the raw presence of the
`owned-by` parameter. These two interpretations diverge for conflicting URLs.
For example, with `?owned-by=alice@example.org&shared-with=nobody`, parsing
correctly gives `shared-with` priority and displays “Your workflows”. Pressing
Clear filters then sees the raw `owned-by`, rewrites it to `anybody`, removes
`shared-with`, and unexpectedly switches to “Shared with you”.

```js
if (next.get("owned-by")) {
  next.set("owned-by", "anybody");
} else {
  next.delete("owned-by");
}
next.delete("shared-with");
```

The interface does not itself produce a URL carrying both parameters, so this
arises only from a hand-edited or externally constructed link. The concern is
the divergence between the two interpretations of “which view am I in”, not
the frequency of this example.

The special cases that exclude `ownedBy=anybody` from `hasActiveFilters` and
preserve it during clearing are further signs that one value represents both
the current view and a refinement within that view.

Suggested fix — derive clearing behaviour from the normalised query model, not
from raw URL parameters, and canonicalise mutually exclusive sharing
parameters by removing the inactive one. Keeping an explicit view discriminator
in the normalised model would also make `clearFilters` and `hasActiveFilters`
independent of the person-refinement values.

reana-ui/src/pages/workflowList/useWorkflowListQuery.js
```diff
@@ -66,6 +66,33 @@ export function useWorkflowListQuery() {
     }
   }, [searchParams, setSearchParams]);
 
+  useEffect(() => {
```
tiborsimko, line 69: **🤖 PR514-07: URL normalisation fires a duplicate, value-identical workflow fetch (NIT)**

Legacy and invalid parameters are normalised in effects after the first render.
That URL replacement creates new `searchParams`, `query`, and `requestParams`
objects even when the serialised API request is unchanged. The fetch guard uses
reference equality, so a URL such as `?shared=true`, `?page=abc`, or
`?status=deleted` dispatches the same workflow request both before and after
normalisation.

This affects precisely the legacy and hand-edited URLs that the normalisation
logic is intended to handle, adding avoidable server work and an extra loading
transition.

Suggested fix — deduplicate workflow requests by their serialised values, or
complete URL normalisation before dispatching the first request. Add a hook or
page-level test asserting that a legacy URL produces one workflow fetch while
still replacing the browser URL. The regression test should match the
application's router configuration: the new hook harnesses opt into
`v7_startTransition`, while `App` mounts a plain `BrowserRouter`, and that flag
can change the URL-update timing under test.

reana-ui/src/pages/workflowList/components/WorkflowFilters.js
```diff
@@ -26,39 +95,150 @@ export default function WorkflowFilters({
...
+  const [userDropdownOpen, setUserDropdownOpen] = useState(false);
+  const dispatch = useDispatch();
+  const usersSharedWithYou = useSelector(getUsersSharedWithYou, _.isEqual);
+  const usersYouSharedWith = useSelector(getUsersYouSharedWith, _.isEqual);
+
+  useEffect(() => {
+    if (category === "mine" && sharedWith && sharedWith !== "nobody") {
+      dispatch(fetchUsersYouSharedWith());
+    }
+  }, [dispatch, category, sharedWith]);
```
tiborsimko, line 112: **🤖 PR514-08: The recipient list is refetched on every recipient selection (NIT)**

The `fetchUsersYouSharedWith()` effect depends on the selected `sharedWith`
value. After the shared refinement is enabled, changing the dropdown from
Anybody to Alice and then Bob therefore requests the same recipient list for
each selection. The action does not cache or deduplicate these requests.

The previous component guarded this fetch using the recipient-list state; the
new lazy-loading approach has dropped that guard.

Suggested fix — trigger the request when entering the owned-and-shared mode,
not when the selected recipient changes. Depending on a derived boolean such
as `category === "mine" && sharedWith !== undefined && sharedWith !== "nobody"`,
or tracking an explicit loaded state, avoids refetching the unchanged option
list. Add a test that changes recipients and asserts that only one list request
is dispatched.

reana-ui/src/pages/workflowList/components/WorkflowStatusFilter.js
```diff
@@ -9,66 +9,80 @@
...
-const statusOptions = WORKFLOW_STATUSES.filter((s) => s !== "deleted").map(
-  (status) => ({
-    key: status,
-    text: status,
-    value: status,
-    icon: statusMapping[status].icon,
-  }),
-);
+const PRIMARY_STATUSES = ["running", "finished", "failed", "stopped"];
+const MORE_STATUSES = ["created", "queued", "pending"];
```
tiborsimko, line 19: **🤖 PR514-09: Secondary status options duplicate the canonical workflow-status list (NIT)**

`WorkflowStatusFilter` hardcodes both the primary and secondary status arrays.
The parser, however, accepts statuses from the canonical `WORKFLOW_STATUSES`
configuration. If a new non-deleted status is added there, the URL layer will
accept it while the sidebar silently omits it and cannot display its selected
state. The two arrays currently cover all seven supported non-deleted statuses,
so this is drift risk rather than a present defect.

Suggested fix — keep the intentionally curated primary list, but derive the
remaining choices from the canonical non-deleted status list, for example by
filtering out `PRIMARY_STATUSES`. Add an assertion that the combined visible
and secondary choices cover every supported non-deleted status exactly once.

reana-ui/src/pages/workflowList/components/WorkflowCategoryTabs.js
```diff
@@ -0,0 +1,56 @@
...
+};
+
+export default function WorkflowCategoryTabs({
+  category,
+  setCategory,
+  refreshedAt,
+  refresh,
+}) {
+  return (
+    <header className={styles.viewHeader}>
```
tiborsimko, line 28: **🤖 PR514-10: The page has no view heading and retains the wrong document title in the shared view (NIT)**

The visible “Your workflows” page heading was removed and replaced by category
navigation. The resulting heading hierarchy begins with the sidebar's
`<h2>Filters</h2>` and `<h3>` section headings, without a page-level heading
that identifies the workflow view.

`BasePage` also continues to receive the fixed title “Your workflows”, so the
browser document title remains “Your workflows” while the “Shared with you”
category is active.

Suggested fix — provide a page-level heading for the workflow list, visually
hidden if the design should remain unchanged, and use either a neutral
“Workflows” document title or a title derived from the active category. Keep
the category controls as navigation rather than relying on them to replace the
document heading.

reana-ui/src/pages/workflowList/components/WorkflowSorting.js
```diff
@@ -11,22 +11,31 @@
...
-  { key: 3, text: "Most Disk used", value: "disk-desc" },
-  { key: 4, text: "Most CPU used", value: "cpu-desc" },
+  { key: 1, text: "Latest first", value: "desc", icon: "sort amount down" },
+  { key: 2, text: "Oldest first", value: "asc", icon: "sort amount up" },
+  { key: 3, text: "Highest disk usage", value: "disk-desc", icon: "hdd" },
+  { key: 4, text: "Highest CPU usage", value: "cpu-desc", icon: "microchip" },
 ];
 
 export default function WorkflowSorting({ value, sort }) {
+  const selected = sortOptions.find((option) => option.value === value);
```
tiborsimko, line 24: **🤖 PR514-FU01: Keep the sorting URL contract and UI options in one place (NIT)**

`WorkflowSorting` now derives its trigger text from the local `sortOptions`
array, while the URL parser still accepts any non-empty `sort` value. An
unsupported URL such as `?sort=unexpected` therefore leaves `selected`
undefined and renders a blank trigger even though the API silently falls back
to its default ordering. The missing parser validation predates this PR, so
this is a non-blocking follow-up rather than a PR-introduced finding.

Suggested fix — export one canonical set of sort values/options and use it both
to validate the URL parameter and to render `WorkflowSorting`. Invalid values
can then normalise to `desc` while the trigger consistently displays its
default label.

reana-ui/src/pages/workflowList/components/WorkflowFilters.module.scss
```diff
@@ -2,14 +2,292 @@
...
+    background: rgba($raven, 0.35);
+    border-radius: 3px;
+  }
+
+  &::-webkit-scrollbar-track {
+    background: transparent;
+  }
+}
+
+.userDropdownOpen {
```
tiborsimko, line 64: **🤖 PR514-FU02: Exercise the sidebar overflow escape on short viewports (NIT)**

Opening a person dropdown changes the bounded `.sidebar` scroll container from
`overflow-y: auto` to `overflow-y: visible` while retaining its `max-height`.
This avoids clipping the dropdown, but on a short desktop viewport it can also
allow the sidebar to paint beyond its intended box; the exact scroll-position
behaviour is worth checking in a real browser. The identical
`.userDropdownOpen` declaration inside the `max-width: 991px` block is
redundant, because the sidebar is already unbounded there.

Suggested fix — add a short-viewport browser test covering dropdown opening,
sidebar position, and scroll retention, and remove the duplicated media-query
rule. If the overflow escape proves unstable, render the dropdown menu outside
the scrolling container instead of changing the container's overflow mode.

reana-ui/src/pages/workflowList/WorkflowList.module.scss
```diff
@@ -2,86 +2,246 @@
...
 }
 
-.title {
+:global(.ui.text.container).workflow-list-container {
+  max-width: 1380px !important;
+}
+
+.browser {
+  display: grid;
+  grid-template-columns: 230px minmax(0, 1fr);
```
tiborsimko, line 37: **🤖 PR514-FU03: Reuse the layout variables in the grid definition (NIT)**

The container declares `--workflow-sidebar-width: 230px` and
`--workflow-browser-gap: 2em` for positioning the fixed loading overlay, but
`.browser` independently hardcodes the same `230px` column and `2em` gap. The
values agree today, but changing only one copy would move the overlay out of
alignment with the results column.

Suggested fix — use `var(--workflow-sidebar-width)` and
`var(--workflow-browser-gap)` in `.browser` so the grid and overlay share one
source of layout geometry.

reana-ui/src/pages/workflowList/useWorkflowListQuery.test.js
```diff
@@ -0,0 +1,199 @@
...
+const wrapperWithEntry = (entry) =>
+  function Wrapper({ children }) {
+    return (
+      <MemoryRouter future={routerFutureFlags} initialEntries={[entry]}>
+        {children}
+      </MemoryRouter>
+    );
+  };
+
+test("selects an owned-workflow sharing refinement", async () => {
```
tiborsimko, line 33: **🤖 PR514-FU04: Consolidate query tests and cover the serialiser branches (NIT)**

Hook coverage is split between `workflowListQuery.test.js` and the new
`useWorkflowListQuery.test.js`, with overlapping router harnesses. Several
component tests also assert Semantic UI implementation classes such as
`.item`/`.active` rather than the selected state exposed to users. Meanwhile,
the pure serialiser's deleted-status combinations and
`showOpenSessionsOnly -> type: "interactive"` mapping are not directly covered.

Suggested fix — keep the hook tests and router harness in one test file, move
component assertions to accessible selected state as the controls are fixed,
and add a small table-driven serialiser test for the status/deleted matrix and
interactive-session flag.

reana-ui/src/pages/workflowList/components/WorkflowFilters.js
```diff
@@ -8,16 +8,85 @@
...
+  { value: "all", label: "All workflows", icon: "list" },
+  { value: "open", label: "Open sessions only", icon: "desktop" },
+];
+
+const DELETED_OPTIONS = [
+  { value: "hidden", label: "Hide deleted runs", icon: "eye slash" },
+  { value: "included", label: "Include deleted runs", icon: "eye" },
+];
+
+const toUserOptions = (users) => [
```
tiborsimko, line 44: **🤖 PR514-FU05: Preserve selected people while dropdown options load (NIT)**

`toUserOptions` only includes the asynchronously fetched users. On first load
of a URL such as `?shared-with=alice@example.org`, the controlled value is
therefore absent from the options until the request completes. It remains
absent if that share was later revoked, leaving an active person filter with a
blank dropdown trigger.

Suggested fix — include the current controlled value in the generated options
when it is not already present, following the page-size dropdown's handling of
out-of-list values. This keeps the applied filter visible both during loading
and for stale shared URLs.

reana-ui/src/pages/workflowList/useWorkflowContinuationCue.js
```diff
@@ -0,0 +1,63 @@
...
+    scheduleUpdate();
+    window.addEventListener("scroll", scheduleUpdate, { passive: true });
+    window.addEventListener("resize", scheduleUpdate);
+
+    return () => {
+      window.cancelAnimationFrame(frameRef.current);
+      window.removeEventListener("scroll", scheduleUpdate);
+      window.removeEventListener("resize", scheduleUpdate);
+    };
+  }, [
```
tiborsimko, line 53: **🤖 PR514-FU06: Keep continuation-cue listeners stable across result changes (NIT)**

The continuation-cue effect both subscribes to `scroll`/`resize` and measures
the current result layout. Because its dependency list includes workflows,
count, page, page size, and loading state, every result update tears down and
re-registers the same window listeners even though only a fresh measurement is
needed.

Suggested fix — install the event listeners in a stable mount-time effect and
trigger remeasurement from a separate effect whose dependencies follow result
changes. An `IntersectionObserver` on the existing sentinel would be another
way to keep the measurement lifecycle separate from window subscriptions.

reana-ui/src/pages/workflowList/components/WorkflowRefinementMenu.js
CameronMcClymont: New tabs and filters do not require keyboard operability in this first implementation.

reana-ui/src/pages/workflowList/workflowListQuery.js
CameronMcClymont: This interface change is intentional and the expected impact is low.

reana-ui/src/pages/workflowList/useWorkflowListQuery.test.js
CameronMcClymont: Conditioned on the controls being fixed, which PR514-01 defers (see my reply).

reana-ui/src/pages/workflowList/useWorkflowListQuery.js
CameronMcClymont: The related code is in `WorkflowList.js`, which has no page-level test harness (no redux/store mocking scaffold exists for it. Building one for two small guards looked disproportionate since the logic is a handful of lines in the component.

reana-ui/src/pages/workflowList/WorkflowList.js
CameronMcClymont: The related code is in `WorkflowList.js`, which has no page-level test harness (no redux/store mocking scaffold exists for it. Building one for two small guards looked disproportionate since the logic is a handful of lines in the component.

reana-ui/src/pages/workflowList/components/WorkflowFilters.module.scss
CameronMcClymont: This can be deferred due to minimal severity and scope of impact.

reana-ui/src/pages/workflowList/WorkflowList.js
```diff
@@ -120,6 +139,24 @@ function Workflows() {
     dispatch(fetchWorkflows({ ...apiParams, showLoader: false }));
   }, [workflowRefresh, dispatch, configLoaded]);
 
+  // A page beyond the result set — hand-edited, or left behind when workflows
+  // disappear — is replaced by the last page that still has results.
+  const lastPage = Math.max(1, Math.ceil(workflowsCount / pageSize));
+  const isPageOutOfRange = workflowsCount > 0 && page > lastPage;
+  useEffect(() => {
+    if (loading || !isPageOutOfRange) return;
+    setPage(lastPage);
```
tiborsimko: **🤖 PR514-11: Returning from workflow details resets valid list pages (MEDIUM)**

The new page-correction effect assumes that `workflowsCount` describes the
current list query:

```js
const lastPage = Math.max(1, Math.ceil(workflowsCount / pageSize));
const isPageOutOfRange = workflowsCount > 0 && page > lastPage;
useEffect(() => {
  if (loading || !isPageOutOfRange) return;
  setPage(lastPage);
}, [loading, isPageOutOfRange, lastPage, setPage]);
```

However, `fetchWorkflow()` delegates to `fetchWorkflows()` with
`workflowIdOrName`. Its response dispatches `WORKFLOWS_RECEIVED`, replacing
the same Redux `workflows` and `total` that the list reads.
`WorkflowDetails` refetches on mount with
`{ refetch: true, showLoader: false }`, so the individual workflow response
stores `total=1` when it completes.

A normal navigation sequence therefore loses the user's list position:

1. Open page 3 of a list with at least three pages.
2. Open a workflow's details and let its refresh complete successfully.
3. Press Back.

The list initially renders with `page=3`, `workflowsCount=1`, and
`loading=false`, so the effect redirects to page 1 before the page-3 response
arrives. The earlier fetch effect dispatches a loading update, but it cannot
change the values already captured by the correction effect for that render.
Returning through history from a narrower search to a valid later page can
trigger the same problem within one mount.

The shared Redux state predates this PR. The new regression is that page
correction acts on that stale state and rewrites a valid URL.

Automatic correction also needs to replace browser history entries, as
described in [PR514-02](https://github.com/reanahub/reana-ui/pull/514#discussion_r3851159068).

Suggested fix — associate the stored total with the request that produced it,
for example by including a key derived from the request parameters in
`WORKFLOWS_RECEIVED`. Correct the page only when the total belongs to a
successful list response matching the current query, ignoring totals from
individual workflow requests and superseded queries. Please include focused
regression coverage for details → Back, Back from a narrower search, and
correction after a matching response confirms that the result set has shrunk. A small test with
a mocked store is sufficient; the assertion should verify navigation
behaviour, not only the page arithmetic.

reana-ui/src/pages/workflowList/components/WorkflowRefinementMenu.js
tiborsimko: **🤖 PR514-01: New tabs and filters are not keyboard operable and expose no selected state (MEDIUM)**

Deferring this work is acceptable for this iteration, with the remaining
accessibility work now tracked in
[#528](https://github.com/reanahub/reana-ui/issues/528). This remains an
accessibility regression: the previous Dropdown and Checkbox controls
supported keyboard operation and exposed their state, whereas the new category
and refinement choices cannot be reached or activated from the keyboard.
This does not itself change the filtering results, but it prevents users who
rely on a keyboard from operating these controls.

Suggested fix — address the deferred work in that issue, covering keyboard
operation and accessible selected state for the category and
refinement controls, an accessible indication of active collapsed sections,
the refresh button's name on mobile, and the accessible-state assertions
described in [PR514-FU04](https://github.com/reanahub/reana-ui/pull/514#discussion_r3851159183).
This finding remains open until those changes are implemented.

reana-ui/src/pages/workflowList/WorkflowList.js
tiborsimko: **🤖 PR514-02: Out-of-range pages render a contradictory range and misleading pagination controls (LOW)**

The amended code suppresses the contradictory range and clamps the pagination
control's active page. The automatic correction still uses `setPage(lastPage)`,
whose search-parameter update pushes a history entry.

For example, `?page=999` with 25 workflows and a page size of 10 is corrected
to page 3, leaving the invalid URL immediately behind it. Navigating Back to
that entry triggers another correction, interfering with normal Back
navigation. This concern applies even when the count used for correction
belongs to the correct query. The stale-count defect described in PR514-11
requires a separate guard; both changes can be made in the correction effect.

Suggested fix — use `{ replace: true }` for automatic page correction, while
retaining push behaviour for deliberate page changes. A navigation option on
`setPage()`, or a separate internal correction function, would allow both.
Please cover the history behaviour in a focused regression test: after
correction, Back should reach the preceding meaningful location without
triggering another correction.

reana-ui/src/pages/workflowList/WorkflowList.module.scss
tiborsimko: Verified fixed in commit `65d877a`: the full-viewport minimum height on
`.results` has been removed, addressing the forced extra height on short
result pages.

reana-ui/src/pages/workflowList/workflowListQuery.js
tiborsimko: Verified addressed in commit `65d877a`: the query contract now explicitly
documents the intentional narrowing of `owned-by=anybody` and legacy
`shared=true` to workflows shared with the user, and the query tests assert
the shared-only request mapping. Accepting this as the documented migration
policy resolves the concern; it does not restore the former union view.

Please consider a short CHANGELOG or release-note entry explaining the change
to existing bookmarks.

reana-ui/src/pages/workflowList/components/WorkflowStatusFilter.js
tiborsimko: Verified fixed in commit `65d877a`: a selected secondary status remains
visible when the additional choices are collapsed. The added tests cover both
initial selection and pressing “Fewer statuses” with an active secondary
status.

reana-ui/src/pages/workflowList/useWorkflowListQuery.js
tiborsimko: Verified fixed in commit `65d877a`: clearing uses the normalised category,
conflicting sharing parameters are canonicalised, and the tests cover keeping
the owned view when `shared-with` takes priority over `owned-by`.

reana-ui/src/pages/workflowList/useWorkflowListQuery.js
tiborsimko: Verified fixed in commit `65d877a`: the request guard compares serialised
parameter values, so URL normalisation no longer causes an additional fetch
through that effect when the request is unchanged. Deferring a dedicated page
test for this small change is reasonable.

reana-ui/src/pages/workflowList/components/WorkflowFilters.js
tiborsimko: Verified fixed in commit `65d877a`: the recipient-list effect depends on
entering the shared-recipient mode, so changing the selected recipient no
longer refetches the options. The added test covers successive recipient
changes.

reana-ui/src/pages/workflowList/components/WorkflowStatusFilter.js
tiborsimko: Verified fixed in commit `65d877a`: secondary statuses are derived from
`NON_DELETED_STATUSES`, and the test checks that every supported non-deleted
status is offered exactly once.

reana-ui/src/pages/workflowList/components/WorkflowCategoryTabs.js
tiborsimko: Verified fixed in commit `65d877a`: the workflow list has a visually hidden
page heading, and both that heading and the document title follow the active
category.

reana-ui/src/pages/workflowList/components/WorkflowSorting.js
tiborsimko: Verified fixed in commit `65d877a`: URL parsing and the sorting dropdown
share the canonical sort options. Unsupported values parse to the default
order and retain a meaningful dropdown label.

reana-ui/src/pages/workflowList/components/WorkflowFilters.module.scss
tiborsimko: **🤖 PR514-FU02: Exercise the sidebar overflow escape on short viewports (NIT)**

The redundant media-query declaration has been removed. Deferring the
short-viewport check is reasonable for this PR. The remaining overflow
behaviour is still unverified: opening a person dropdown switches the bounded
sidebar from `overflow-y: auto` to `visible` while retaining its
`max-height`.

Suggested fix — keep this thread open until the short-viewport check is
complete. Exercise dropdown opening and closing after scrolling the sidebar
on a short desktop viewport, checking menu visibility and scroll retention.
Change the overflow approach only if that check reveals a problem.

reana-ui/src/pages/workflowList/WorkflowList.module.scss
tiborsimko: Verified fixed in commit `65d877a`: the grid uses the sidebar-width and gap
variables already used to position the loading overlay.

reana-ui/src/pages/workflowList/useWorkflowListQuery.test.js
tiborsimko: **🤖 PR514-FU04: Consolidate query tests and cover the serialiser branches (NIT)**

Verified that the hook tests are consolidated in `useWorkflowListQuery.test.js`
and that the serialiser tests cover the status/deleted combinations and the
interactive-session flag. The remaining accessible-state assertions can follow
with the keyboard and selection-semantics work from
[PR514-01](https://github.com/reanahub/reana-ui/pull/514#discussion_r3851159056).

Suggested fix — complete those assertions in
[#528](https://github.com/reanahub/reana-ui/issues/528):
exercise keyboard activation and verify the selected or checked state exposed
to users, rather than relying on Semantic UI's `.item` and `.active`
classes. The test consolidation and serialiser coverage need no further work.

reana-ui/src/pages/workflowList/components/WorkflowFilters.js
tiborsimko: Verified fixed in commit `65d877a`: the selected person is included in the
dropdown options when absent from the fetched list, keeping the filter
readable during loading and after a share is revoked. The added test covers a
selected person missing from the available options.

reana-ui/src/pages/workflowList/useWorkflowContinuationCue.js
tiborsimko: Verified fixed in commit `65d877a`: the scroll and resize listeners use a
stable callback, while result changes trigger remeasurement in a separate
effect without replacing those listeners.

reana-ui/src/pages/workflowList/components/WorkflowFilters.module.scss
CameronMcClymont: Did the short viewport check and found an issue: switching the overflow mode destroys the scroll container so the contents jump on opening and move the control that was clicked. No problems when closing.

Rendering the menu outside the scrolling container looks like the right fix, but Semantic UI React's Dropdown has no portal option, so it means replacing that control rather than adjusting the stylesheet. Since this goes beyond the scope of a NIT, opened an issue to tackle this later (#529) rather than in this PR.

reana-ui/src/pages/workflowList/WorkflowList.js
tiborsimko: Verified fixed in commit `7c1931d`: the stored total now carries its request key, and automatic page correction only uses a matching list response. Totals from individual-workflow requests and other searches cannot redirect a valid page. The new tests cover those stale totals and correction when results shrink; [CI passes all 151 tests across 16 suites](https://github.com/reanahub/reana-ui/actions/runs/34841750220/job/103968112441).

The remaining pagination interaction follow-up is tracked in [#530](https://github.com/reanahub/reana-ui/issues/530) and does not require further changes in this PR.

reana-ui/src/pages/workflowList/WorkflowList.js
tiborsimko: Verified fixed in commit `7c1931d`: automatic page correction uses `{ replace: true }`, while deliberate page changes retain normal history-push behaviour. The new history test verifies that Back skips the corrected entry.

reana-ui/src/pages/workflowList/components/WorkflowFilters.module.scss
tiborsimko: Thanks for completing the short-viewport check and documenting the scroll jump in [#529](https://github.com/reanahub/reana-ui/issues/529). The check and tracking are complete; the dropdown fix can follow in that issue. No further change is requested for this PR.

tiborsimko (approved):

PR #517, opened 2026-06-19
auth rework

### LINKED ISSUE #977: Improve SSO and use JWT for external authentication
Improve SSO allow only one provider and use JWT for external authentication.

Sohn123:
Close: reanahub/reana#977


reana-ui/src/actions.js
```diff
@@ -28,7 +27,6 @@ import client, {
   WORKFLOW_SPECIFICATION_URL,
 } from "~/client";
 import {
-  getConfig,
```
tiborsimko, line 31: **🤖 PR517-01: Interactive-session opening references a removed selector import (MEDIUM)**

The branch removes `getConfig` from the selector imports but still calls it before opening an interactive session. Clicking the action throws `ReferenceError` before making the API request, so the new per-session notebook-secret flow cannot be reached through this UI path.

The selector is absent from `reana-ui/src/actions.js:29-33` and referenced at `reana-ui/src/actions.js:463-467`.

Suggested fix — restore the selector import or pass the required inactivity configuration through another explicit dependency, and add an action-level test that executes `openInteractiveSession()`.

reana-ui/src/pages/signin/Signin.js
```diff
@@ -8,141 +8,52 @@
...
-      ? tokenIssuancePolicyRaw
-      : "manual";
-  const shouldNotifyEmailConfirmation =
-    config.userConfirmation && tokenIssuancePolicy !== "auto";
+  const bffAuth = config.auth ?? {};
+  const bffEnabled = Boolean(bffAuth.bff_enabled);
 
-  const handleClick = (ssoProvider) => {
+  const getNext = () => {
     const from = location.state?.from || {
```
tiborsimko, line 69: **🤖 PR517-02: OAuth callback error redirects are never surfaced to the user (MEDIUM)**

reana-server reports an issuer authorization failure or a provisioning failure by redirecting to `{next}?login_error=authorization` or `{next}?login_error=provisioning` without authentication cookies (`reana_server/rest/auth.py:282-288,324-333`). Nothing in the UI reads `login_error`; the protected target is redirected to `/signin`, and `Signin.js` renders only the sign-in button.

Users who cancel or are denied at the identity provider, or whose identity cannot be linked or provisioned, therefore arrive back at the same sign-in page with no explanation. In the protected-route case the original query string is retained under `location.state.from`, which means merely reading `location.search` inside `Signin` would still miss it.

Suggested fix — preserve and consume a structured login error during the protected-route redirect (or inspect `location.state.from.search` in `Signin`), render a distinct actionable message for the supported error codes, and remove the parameter after displaying it. The server should append the parameter with URL-aware query handling when `next` already contains a query string.

reana-ui/src/pages/signin/Signin.js
```diff
@@ -8,141 +8,52 @@
...
-  )
-    .trim()
-    .toLowerCase();
-  const tokenIssuancePolicy =
-    tokenIssuancePolicyRaw === "auto" || tokenIssuancePolicyRaw === "manual"
-      ? tokenIssuancePolicyRaw
-      : "manual";
-  const shouldNotifyEmailConfirmation =
-    config.userConfirmation && tokenIssuancePolicy !== "auto";
+  const bffAuth = config.auth ?? {};
```
tiborsimko, line 26: **🤖 PR517-03: The BFF flow has no deliberate "access not granted" UI state (MEDIUM)**

The server intentionally establishes a BFF session when the identity lacks the required role and says that `/api/you` plus the UI will present an "access not granted" state (`reana_server/rest/auth.py:324-329` and `reana_server/decorators.py:48-55`). The UI implements no such state. A first-time role-less identity receives a 403 from `/api/you`, which `App.js` renders as the generic full-page user-fetch error; a previously provisioned user whose role is revoked can still reach endpoints exempted by `token_required=False`, while role-gated sharing calls emit generic error notifications.

This is the UI half of PR789-02: the server distinguishes authentication from entitlement, but the client state model still treats the result as either signed in or generically broken.

Suggested fix — expose a stable machine-readable entitlement result from a narrow server endpoint (or `/api/you`) and represent it explicitly in the UI's auth state. Render an "authenticated but not authorised" page with guidance, while reserving generic error handling for actual request failures.

reana-ui/src/client.js
```diff
@@ -104,14 +108,29 @@ class Client {
...
+      ...options
+    } = {},
   ) {
+    const requestMethod = method.toLowerCase();
+    const csrfToken = !["get", "head", "options"].includes(requestMethod)
+      ? getCookieValue(CSRF_COOKIE)
+      : null;
+    const headers = {
+      ...optionHeaders,
+      ...(csrfToken ? { [CSRF_HEADER]: csrfToken } : {}),
```
tiborsimko, line 141: **🤖 PR517-04: CSRF header attachment has no test coverage (LOW)**

The client-side half of the new CSRF double-submit defence is the `X-REANA-CSRF` header that `Client._request` attaches from the `reana_csrf` cookie for every non-GET request (`src/client.js:77-78` and the `_request` wrapper). The new tests in `client.test.js` cover the session-expiry matcher, the unauthorised callback, and URL building — but not this. A regression such as reordering a headers spread so caller-supplied headers clobber the CSRF entry, or renaming the cookie on one side only, would ship silently and break every mutating action at runtime.

Suggested fix — add a test that sets `document.cookie = "reana_csrf=abc"`, issues a `post` through `_request`, and asserts the request carried `X-REANA-CSRF: abc` (and that a `get` did not).

reana-ui/src/components/WorkflowBadges.js
```diff
@@ -63,7 +97,8 @@ export default function WorkflowBadges({ workflow, badgeSize = "tiny" }) {
                 </i>
               }
               as="a"
-              href={INTERACTIVE_SESSION_URL(sessionUri, reanaToken)}
+              href={INTERACTIVE_SESSION_URL(sessionUri)}
```
tiborsimko: **🤖 PR517-05: Alternative link activation bypasses the notebook-secret fetch (LOW)**

The Notebook badge retains a real secret-less `href` while the token-fetching flow exists only in `onClick` (`src/components/WorkflowBadges.js:99-104`). A normal left-click is intercepted, fetches the per-session secret, and navigates the newly opened window. Middle-click and the browser's "open link in new tab" action follow the raw `href` instead, landing on Jupyter's token prompt even though the UI never exposes the token to the user.

Suggested fix — do not expose a directly navigable secret-less link. Render the badge as an accessible button that always runs the secret-fetch flow, or provide a BFF-protected server launch URL that performs the authorised lookup and redirects to the tokenised notebook URL for every browser activation mode.

reana-ui/src/pages/signin/Signin.js
```diff
@@ -8,141 +8,52 @@
...
+import { useSelector } from "react-redux";
+import { Button, Segment } from "semantic-ui-react";
+import { useLocation } from "react-router-dom";
 
 import { getConfig } from "~/selectors";
-import SignForm from "./components/SignForm";
 import SignContainer from "./components/SignContainer";
-import { USER_OAUTH_SIGNIN_URL } from "~/client";
-import { triggerNotification, userSignin } from "~/actions";
-import { useSubmit, useDocumentTitle } from "~/hooks";
```
tiborsimko, line 21: **🤖 PR517-06: Dead code left behind by the auth removal (NIT)**

Two leftovers from the deleted local-auth surface survive the cleanup:

- `useSubmit` (`src/hooks.js:15`) — its only consumers were the deleted `SignForm`/`Signin`/`Signup` form components; nothing imports it any more.
- `USER_ERROR.sign` (`src/errors.js:13`) — `getUserSignErrors` is gone and only `USER_ERROR.fetch` is still referenced (`src/selectors.js:28`, `src/reducers.js:206`).

Suggested fix — delete both, keeping the auth-removal diff self-contained.

reana-ui/src/pages/signin/Signin.js
```diff
@@ -8,141 +8,103 @@
...
+  const bffAuth = config.auth ?? {};
+  const bffEnabled = Boolean(bffAuth.bff_enabled);
+  const currentQuery = new URLSearchParams(location.search);
+  const fromQuery = new URLSearchParams(location.state?.from?.search ?? "");
+  const [loginError] = useState(
+    currentQuery.get("login_error") ?? fromQuery.get("login_error"),
+  );
+
+  const loginErrorMessages = {
+    authorization:
```
tiborsimko, line 35: **🤖 PR517-07: The "authorization" login-error message describes the wrong failure (LOW)**

The message for `login_error=authorization` reads (`reana-ui/src/pages/signin/Signin.js:34-36`):

```js
authorization:
  "Your identity is authenticated, but access to this REANA deployment has not been granted.",
```

but the server emits `login_error=authorization` only when the issuer returns an OAuth `error` on the callback — i.e. the user cancelled at the IdP consent screen or the issuer refused authorisation *before* any authentication completed (reana-server's `oauth_callback` in `reana_server/rest/auth.py:303-311`). The genuinely "authenticated but not entitled" case takes a different path entirely: the session is established and the UI shows `AccessNotGranted` via the `/api/you` 403. So this message is factually wrong for every path that can reach it, and it duplicates the semantics of the separate `AccessNotGranted` page. A user who clicked "Cancel" at Keycloak is told they are authenticated.

Suggested fix — reword to describe an issuer-side authorisation failure, e.g. "Sign-in was cancelled or refused by the identity provider. Please try again or contact the deployment administrator." Update the corresponding expectation in `Signin.test.js:47-49`.

reana-ui/src/pages/error/AccessNotGranted.js
```diff
@@ -0,0 +1,18 @@
...
+  This file is part of REANA.
+  Copyright (C) 2026 CERN.
+
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+import Error from "~/components/Error";
+
+export default function AccessNotGranted() {
```
tiborsimko, line 15: **🤖 PR517-08: AccessNotGranted is a terminal page with no sign-out escape hatch (LOW)**

When `/api/you` returns `access_not_granted`, `App.js:64-66` replaces the entire component tree — router, header, and the sign-out control included — with `AccessNotGranted` (`reana-ui/src/pages/error/AccessNotGranted.js:11-18`), which renders only a logo, title, and message (the shared `Error` component has no actions). Meanwhile the server keeps the BFF session alive by design ("The session is still established: /api/you will answer 403", `reana_server/rest/auth.py:346-349`).

A user who signed in with the wrong identity (personal instead of institutional account) has no in-app way to log out and retry: refreshing re-runs `loadUser` and lands on the same page, and IdP SSO means a fresh tab re-authenticates the same identity. Typing `/api/logout` into the address bar does not work either — the endpoint is POST-only and requires the CSRF header (`reana_server/rest/auth.py:374`) — so the only escape is manual cookie clearing or a hand-crafted request from the developer tools.

Suggested fix — add a "Sign out" button to `AccessNotGranted` that dispatches the existing `userSignout()` action (it already follows the server-provided `logout_url` to end the IdP session, `actions.js:171-186`); extend the `Error` component to accept children or an action slot.

reana-ui/src/actions.js
tiborsimko: Addressed in commit `b3ea218`, kept as a separate fixup on top of the original commit for review clarity. This fixup commit can be squashed into the original commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana-ui/src/pages/signin/Signin.js
tiborsimko: Addressed in commit `b3ea218`, kept as a separate fixup on top of the original commit for review clarity. This fixup commit can be squashed into the original commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana-ui/src/pages/signin/Signin.js
tiborsimko: Addressed in commit `b3ea218`, kept as a separate fixup on top of the original commit for review clarity. This fixup commit can be squashed into the original commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana-ui/src/client.js
tiborsimko: Addressed in commit `b3ea218`, kept as a separate fixup on top of the original commit for review clarity. This fixup commit can be squashed into the original commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana-ui/src/components/WorkflowBadges.js
tiborsimko: Addressed in commit `b3ea218`, kept as a separate fixup on top of the original commit for review clarity. This fixup commit can be squashed into the original commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana-ui/src/pages/signin/Signin.js
tiborsimko: Addressed in commit `b3ea218`, kept as a separate fixup on top of the original commit for review clarity. This fixup commit can be squashed into the original commit before merge, preserving the first author with `Co-authored-by` for the fixup.

reana-ui/src/pages/error/AccessNotGranted.test.js
```diff
@@ -0,0 +1,22 @@
...
+  REANA is free software; you can redistribute it and/or modify it
+  under the terms of the MIT License; see LICENSE file for more details.
+*/
+
+import { render, screen } from "@testing-library/react";
+
+import AccessNotGranted from "./AccessNotGranted";
+
+test("explains that authentication succeeded but entitlement did not", () => {
+  render(<AccessNotGranted />);
```
tiborsimko, line 27: **🤖 PR517-09: The amended authentication UI test suite is red (MEDIUM)**

At `e7cede9`, both the focused local run and the GitHub `js-tests` check fail
two of 124 tests:

- `AccessNotGranted.test.js` renders the component without a Redux
  `Provider`, but the amended component now calls `useDispatch()`;
- `Signin.test.js` expects the old text “authenticated, but access ...” for
  `login_error=authorization`, while the component now correctly renders
  “Sign-in was cancelled or not authorised by the identity provider.”

The functional changes themselves appear deliberate, but the red required
suite cannot gate later authentication regressions. The new sign-out button —
the purpose of the latest amendment — currently has no behavioural coverage.

Suggested fix — render `AccessNotGranted` with the normal test store/provider,
click the button and assert the sign-out action; update the structured
callback-error expectation to the current, semantically correct copy. Keep
both tests as behavioural guards and restore the `js-tests` check to green.

reana-ui/src/pages/error/AccessNotGranted.js
```diff
@@ -0,0 +1,23 @@
...
+import { useDispatch } from "react-redux";
+import { Button } from "semantic-ui-react";
+
+export default function AccessNotGranted() {
+  const dispatch = useDispatch();
+  return (
+    <Error
+      title="Access not granted"
+      message="You are authenticated, but your identity is not entitled to use this REANA deployment. Contact the deployment administrator to request access."
+      action={<Button onClick={() => dispatch(userSignout())}>Sign out</Button>}
```
tiborsimko: **🤖 PR517-10: Access-denied sign-out failures are invisible and network errors break the handler (LOW)**

`AccessNotGranted` now offers the necessary sign-out button, but it is rendered
outside both `BasePage` and `SignContainer`, the two layouts that mount the
global `Notification` component. If logout returns an HTTP error,
`userSignout()` dispatches `errorActionCreator`, but no notification is visible
on this terminal page and the user receives no explanation or recovery state.

On a pure network/TLS failure the handler is worse:
`errorActionCreator` destructures `error?.response` and then `data` without a
fallback. When `response` is absent it throws `TypeError` inside the `.catch`,
so even the error action is never dispatched.

Suggested fix — make `errorActionCreator` tolerate response-less failures and
give `AccessNotGranted` a local pending/error state or mount the shared
notification there. Add tests for an HTTP logout rejection and a network error,
asserting that the button remains usable and an actionable error is shown.

reana-ui/src/actions.js
tiborsimko: **🤖 PR517-01 follow-up (2026-07-28): verified fixed**

*Component:* reana-ui — interactive-session opening

The removed access-token selector is no longer referenced. Opening a running
notebook now requests the owner-only per-session secret from the server and
constructs the notebook URL from that response.

Verified fixed at `e7cede9`. No residual for the missing import/runtime crash.

reana-ui/src/pages/signin/Signin.js
tiborsimko: **🤖 PR517-02 follow-up (2026-07-28): verified fixed**

*Component:* reana-ui — BFF callback errors on the sign-in page

The sign-in route reads the structured `login_error` value from both the
current location and preserved redirect state and renders a deliberate message
for authorisation and provisioning failures. Callback errors are no longer
silently discarded.

Verified fixed at `e7cede9`. PR517-09 separately notes one stale text
expectation in the amended test.

reana-ui/src/pages/signin/Signin.js
tiborsimko: **🤖 PR517-03 follow-up (2026-07-28): verified fixed**

*Component:* reana-ui — authenticated-but-not-entitled state

`loadUser` preserves the server's status and machine-readable
`access_not_granted` code, and routing now displays a dedicated
`AccessNotGranted` page instead of treating the user as generically signed out
or operationally broken.

Verified fixed at `e7cede9`. No functional residual; PR517-09 covers the
component test harness.

reana-ui/src/client.js
tiborsimko: **🤖 PR517-04 follow-up (2026-07-28): verified fixed**

*Component:* reana-ui — CSRF header attachment

The common request wrapper reads `reana_csrf` and adds `X-REANA-CSRF` to
non-safe methods while leaving GET/HEAD/OPTIONS untouched. Tests assert both
the mutating and safe request behaviour.

Verified fixed at `e7cede9`. No residual.

reana-ui/src/components/WorkflowBadges.js
tiborsimko: **🤖 PR517-05 follow-up (2026-07-28): verified fixed**

*Component:* reana-ui — notebook launch control

The pre-secret anchor has been replaced with a button-driven launch. Every
supported activation enters the same handler, opens a blank `noopener` window,
fetches the session secret, and only then navigates it. There is no raw
`href` for middle-click or modifier-click to bypass the fetch.

Verified fixed at `e7cede9`. No residual.

reana-ui/src/pages/signin/Signin.js
tiborsimko: **🤖 PR517-06 follow-up (2026-07-28): verified fixed**

*Component:* reana-ui — obsolete local-auth code

The unused sign-up/sign-in/token-request/email-confirmation actions, client
methods, constants and associated legacy handling have been removed. The UI
now exposes the BFF identity-provider login and logout contract rather than
retaining unreachable Invenio-era branches.

Verified fixed at `e7cede9`. No residual for the reported dead code.

reana-ui/src/pages/signin/Signin.js
tiborsimko: **🤖 PR517-07 follow-up (2026-07-28): verified fixed**

*Component:* reana-ui — callback authorisation-error copy

`login_error=authorization` now explains that sign-in was cancelled or not
authorised by the identity provider. Missing REANA entitlement is handled by
the separate access-not-granted state, so the two failure meanings are no
longer conflated.

Verified fixed at `e7cede9`. The stale test expectation is tracked separately
as PR517-09.

reana-ui/src/pages/error/AccessNotGranted.js
tiborsimko: **🤖 PR517-08 follow-up (2026-07-28): verified fixed**

*Component:* reana-ui — `AccessNotGranted` recovery action

The terminal entitlement page now offers a Sign out button that dispatches the
normal BFF logout flow. The logout response's issuer URL is followed when
present, allowing a user to end the current IdP session and try another
identity.

Verified fixed at `e7cede9`. No functional residual; PR517-09 covers the
missing Redux provider in this component's test.

reana-ui/src/pages/profile/components/GitLabProjects.js
```diff
@@ -157,6 +198,48 @@ export default function GitLabProjects() {
...
+        )}
+        {webhookToken?.configured && (
+          <Message warning={webhookToken.expired} info={!webhookToken.expired}>
+            <Message.Header>
+              GitLab webhook authorization
+              {webhookToken.expired ? " has expired" : " is time-limited"}
+            </Message.Header>
+            <p>
+              {webhookToken.expired
+                ? "GitLab cannot start workflows until you renew this authorization."
```
tiborsimko, line 292: **🤖 PR517-11: Expired webhook toggles remain enabled and discard the expected 409 (LOW)**

The new status banner tells the user to renew an expired authorization, but an
unchecked project toggle remains enabled. Its expected server-side 409 follows
the existing promise catch, which throws a new error without presenting any
notification or refreshing status. The toggle simply reverts after the failed
request, leaving the user without the actionable renewal message; the same race
is possible if authorization expires after the page loaded.

Suggested fix — disable only project-enabling toggles while the loaded status is
expired (deletion should remain available), and handle a 409 by refreshing the
status and showing the renewal action. Add UI coverage for both already-expired
state and expiry between status load and toggle submission.

reana-ui/src/pages/error/AccessNotGranted.test.js
tiborsimko: **🤖 PR517-09 follow-up (2026-08-11): verified fixed at 2a44d67**

The UI tests now use the current Redux/auth mocks, cover the OIDC sign-out flow,
and assert the amended login and access-denied wording. The JavaScript suite is
green at the current head, so the auth rework is no longer landing with the
previously stale baseline tests.

reana-ui/src/pages/error/AccessNotGranted.js
tiborsimko: **🤖 PR517-10 follow-up (2026-08-11): verified fixed at 2a44d67**

The shared error action now handles failures without an Axios `response` and
falls back to a stable message. `AccessNotGranted` mounts the notification
component, and tests cover both HTTP and network failures plus retry. A failed
OIDC/config request no longer turns the access-denied page itself into an
exception or hides the recovery action.

reana-ui/src/pages/profile/components/GitLabProjects.js
Sohn123: Fixed in 73c5149: enable-toggles are now disabled while the loaded webhook status is expired (`webhookToken?.expired && hookId === null`); disabling an already-enabled project stays available. A 409 from the toggle request refetches the webhook status instead of silently reverting, so the renewal banner also covers the expire-between-load-and-submit race. New tests cover both the already-expired state and the race.

tiborsimko: **🤖 PR517-11 follow-up (2026-08-21): verified fixed at 6b0c1a1**

Project-enabling toggles are now disabled while webhook authorisation is
expired, while disabling/removal remains available. If expiry races with a
toggle request, the UI handles the server's 409, refreshes status, and presents
the renewal action instead of silently reverting the control. Tests cover both
already-expired and mid-request expiry cases. No residual concern remains for
this finding.

reana-ui/src/pages/profile/components/GitLabProjects.js
```diff
@@ -122,6 +153,23 @@ export default function GitLabProjects() {
     setSearchFilter(value);
   };
 
+  const renewWebhookToken = () => {
+    setRenewingWebhookToken(true);
+    setWebhookTokenRenewalError(null);
+    client
+      .renewGitlabWebhookToken()
+      .then(({ data }) => {
+        setWebhookToken(data);
```
tiborsimko: **🤖 PR517-12: The expired-renewal warning is discarded by its only caller (MEDIUM)**

`PUT /api/gitlab/webhook-token` now returns a `message` when the delegated
authorisation had already expired, warning that GitLab may have auto-disabled
affected hooks. The renewal handler stores the response, but the component
never renders that field:

```javascript
.then(({ data }) => {
  setWebhookToken(data);
  setWebhookTokenStatusError(null);
})
```

The response also changes `expired` to false, so the existing banner
immediately switches from "has expired" to the neutral "is time-limited"
state. The user is told that renewal succeeded but never sees the recovery
warning, even though this UI is the only production caller of the new response
contract. A GitLab-disabled hook can therefore remain silent after an
apparently successful renewal.

Suggested fix — render `webhookToken.message` as a warning in the GitLab
authorisation block and keep it visible after successful renewal. Add a
component test whose renewal response contains the warning and assert that the
manual test/re-enable guidance is shown.

This is the reana-ui half of `PR789-33` on reanahub/reana-server#789, where the
server-side `message` and its README documentation are already in place.

reana-ui/src/components/WebhookExpiryWarning.js
```diff
@@ -0,0 +1,77 @@
...
+  // Shared with GitLabProjects's profile-page view: a renewal made there
+  // dispatches to the same state, so this banner updates immediately
+  // instead of waiting for its own next fetch.
+  const status = useSelector(getGitlabWebhookToken);
+
+  useEffect(() => {
+    let active = true;
+    if (signedIn) {
+      client
+        .getGitlabWebhookToken()
```
tiborsimko: **🤖 PR517-13: The global webhook warning refetches on every page navigation (LOW)**

`WebhookExpiryWarning` is mounted inside `BasePage`, while each routed page
owns a separate `BasePage` instance. Navigating between pages therefore
unmounts and remounts the warning, and its effect performs another authenticated
`GET /api/gitlab/webhook-token` for every signed-in user. This happens even for
users who have never configured GitLab and deployments without GitLab
integration. Each request repeats JWT validation, user lookup, and a database
read solely to rediscover the same status.

The profile page independently fetches the webhook status and dispatches the
same `GITLAB_WEBHOOK_TOKEN_RECEIVED` action. Neither request path orders or
deduplicates webhook-status responses. An older in-flight response, or the
global component's error path dispatching `null`, can consequently overwrite a
newer status written after renewal. The `lastFetchRequest` guard in
`GitLabProjects` covers project-list requests only, not these status requests.

Suggested fix — fetch the status once at application level and represent
unloaded/loading separately from a loaded `null` result. Coordinate explicit
refreshes after renewal and relevant webhook operations through the same state
owner, with stale-response protection, instead of issuing an independent
request on every `BasePage` mount.

reana-ui/src/pages/profile/components/GitLabProjects.js
tiborsimko: **🤖 PR517-12 follow-up (2026-08-25): verified fixed at b84ec7b**

The user-visible recovery gap is closed through persistent profile guidance
and a proactive global warning rather than by rendering the renewal response's
`message` field. Users are warned when authorisation is expired or within the
last quarter of its lifetime, and the profile always explains how to recover a
GitLab hook which was auto-disabled before renewal. Renewal dispatches the new
status to the shared store so the warning updates immediately.

This is the UI-side completion of PR789-33 on reanahub/reana-server#789; the
server-side status, API message, documentation, and longer-term #805 tracking
are already in place. No residual concern remains for this finding.

reana-ui/src/components/WebhookExpiryWarning.js
```diff
@@ -0,0 +1,81 @@
...
+
+    if (phase === "idle") {
+      dispatch(loadGitlabWebhookTokenStatus());
+      return undefined;
+    }
+
+    let retryTimer;
+    if (phase === "error") {
+      const retryDelay = Math.max(0, retryAt - Date.now());
+      retryTimer = window.setTimeout(
```
tiborsimko, line 75: **🤖 PR517-14: The webhook-status retry repeats indefinitely (LOW)**

The shared request phase prevents navigation remounts from duplicating the
initial status request, but its error recovery is not bounded to the single
retry described by the commit:

```javascript
if (phase === "error") {
  const retryDelay = Math.max(0, retryAt - Date.now());
  retryTimer = window.setTimeout(
    () => dispatch(loadGitlabWebhookTokenStatus()),
    retryDelay,
  );
}
```

When that retry fails, `loadGitlabWebhookTokenStatus()` dispatches another
error with a fresh `retryAt = Date.now() + 60_000`. The effect runs again for
the changed deadline and schedules another attempt, repeating once per minute
for as long as the tab remains open. There is no attempt counter, terminal
error phase, or backoff. A sustained server failure or proxy rule blocking
this endpoint therefore makes every signed-in tab poll indefinitely while the
service is already degraded.

The regression test supplies a fixed error-state store followed by a
successful mocked request. It never feeds a second failure through the real
reducer, so the repeated scheduling is not exercised.

Suggested fix — make the intended retry policy explicit and bounded. For a
single retry, track the attempt in the shared slice and enter a terminal error
phase after the second failure; alternatively use capped exponential backoff
and avoid retrying non-transient 4xx responses. Add a reducer-backed test which
drives at least two consecutive failures and verifies that no further timer is
scheduled after the configured limit. The current `.catch()` discards the
error object, so the status-aware option also requires carrying the response
status into `GITLAB_WEBHOOK_TOKEN_FETCH_ERROR`.

reana-ui/src/components/WebhookExpiryWarning.js
```diff
@@ -0,0 +1,81 @@
...
+    };
+  }, [signedIn, phase, retryAt, dispatch]);
+
+  if (!signedIn || !webhookAuthorizationNeedsAttention(status)) return null;
+
+  return (
+    <Message warning>
+      <Message.Header>
+        GitLab webhook authorization{" "}
+        {status.expired ? "has expired" : "expires soon"}
```
tiborsimko: **🤖 PR517-15: Cached webhook expiry can remain stale after its deadline (LOW)**

The warning derives the start of its attention window from `expires_at`, but
the displayed expiry state still comes from the boolean captured by the last
server response:

```javascript
if (status.expired) return true;
const expiresAt = Date.parse(status.expires_at);
// ... compare expiresAt with Date.now()

{status.expired ? "has expired" : "expires soon"}
```

After an initially valid authorisation passes its `expires_at`, any ordinary
navigation or re-render makes the helper recognise that the deadline is near
or past, but the header still says “expires soon” because the cached
`status.expired` remains `false`. At that point REANA already rejects webhook
deliveries as expired, so the banner contradicts the effective server state.

There is also no time-driven transition while the user remains on one page.
Once the request reaches `phase === "ready"`, the effect schedules neither a
timer nor a successful-status revalidation. Crossing the final-quarter
warning threshold or the expiry deadline alone therefore does not cause a
render, and the banner can remain absent in a long-lived tab until some
unrelated UI activity happens.

Suggested fix — the attention helper already recognises a passed deadline
through its `expires_at` comparison; the header expression is the part which
still reads the cached `status.expired`. Derive an effective expiry once, for
example `status.expired || Date.parse(status.expires_at) <= Date.now()`, and
use it for the header text as well as the helper. Schedule a timer for the next
warning-state transition (or perform a bounded revalidation then), and cover
crossing both the final-quarter threshold and the actual expiry deadline with
fake-timer tests, including the header text after expiry.

reana-ui/src/components/WebhookExpiryWarning.js
tiborsimko: **🤖 PR517-13 follow-up (2026-08-25): partially addressed at d37c1c8**

Thanks for moving the initial request lifecycle into the shared Redux slice.
The explicit idle/loading/ready/error phases stop `BasePage` remounts from
refetching on every navigation, and request IDs correctly prevent stale
responses from the global loader from overwriting newer tagged requests.

The profile page still owns a second status value and bypasses that ordering.
Both its status fetch and renewal dispatch the shared receive action without a
request ID:

```javascript
dispatch({ type: GITLAB_WEBHOOK_TOKEN_RECEIVED, status: data });
```

The reducer rejects a response only when `action.requestId !== undefined` and
does not match `activeRequestId`, so these untagged actions are always treated
as authoritative. This is reachable after the first status has loaded: the
profile starts another status fetch after a successful enable-toggle, after a
409 from a toggle, and whenever search or pagination reloads the project list.
If the user renews while one of those requests is in flight, the older response
can land afterwards and restore the pre-renewal status in both Redux and the
component's local `webhookToken` state.

Suggested fix — route profile status reads through the tagged shared thunk and
derive the profile view from the shared status instead of maintaining an
independent copy. Give renewal a distinct, unconditionally authoritative
action (or otherwise invalidate any active read request) so a pre-renewal read
cannot overwrite it.

tiborsimko: **🤖 PR517-13 follow-up (2026-08-26): verified fixed at 2ec3ef4**

`GitLabProjects` now derives webhook status from the shared Redux slice and
routes all status reads through the request-ID-tagged
`loadGitlabWebhookTokenStatus()` thunk. The duplicate local status owner and
its untagged receive actions are gone.

Renewal uses a distinct `GITLAB_WEBHOOK_TOKEN_UPDATED` action which is
unconditionally authoritative and clears `activeRequestId`; a later response
from a read that began before renewal is therefore rejected by the reducer.
The reducer regression drives that ordering explicitly. This closes both the
duplicate-fetch/state concern and the reachable stale-response race.

reana-ui/src/components/WebhookExpiryWarning.js
tiborsimko: **🤖 PR517-15 follow-up (2026-08-26): verified fixed at 2ec3ef4**

The banner now derives effective expiry from both the server flag and the
`expires_at` deadline through `webhookAuthorizationIsExpired()`, and uses that
same result for the header text. It can no longer say “expires soon” after the
authorisation has already expired.

`nextWebhookAuthorizationTransition()` schedules a render at the warning
threshold and then at expiry, with long delays clamped to the browser's
maximum timeout and naturally rescheduled. Fake-timer tests cover both
transitions and the stale-server-boolean case. No residual concern remains.

reana-ui/src/components/WebhookExpiryWarning.js
tiborsimko: **🤖 PR517-14 follow-up (2026-08-26): verified fixed at 2ec3ef4**

The retry policy is now genuinely bounded. Only an initial transient failure
(network failure, 408, 429, or 5xx) enters `retry_wait`; the timer marks the
next request as `automaticRetry`, so a second failure sets `retryAt: null` and
enters the terminal error phase instead of scheduling again. Non-transient
4xx responses are terminal immediately.

The action now preserves the response status long enough to make this
classification, and tests cover an automatic retry failing as well as a 403
which must not be retried. No residual concern remains.

reana-ui/src/pages/profile/components/GitLabProjects.js
```diff
@@ -9,27 +9,93 @@
...
+    () => dispatch(loadGitlabWebhookTokenStatus()),
+    [dispatch],
+  );
+
+  // The webhook-token phase is shared with WebhookExpiryWarning, which
+  // normally kicks off the first fetch -- but this page must not depend on
+  // that component happening to be mounted too, so trigger it here as well
+  // if nothing has fetched yet.
+  useEffect(() => {
+    if (webhookTokenPhase === "idle") {
```
tiborsimko, line 81: **🤖 PR517-16: The profile page can fetch webhook-token status twice on mount (NIT)**

Both the global warning and the profile component independently start the same
request when the shared Redux phase captured during their initial render is
`idle`:

```javascript
if (phase === "idle") {
  dispatch(loadGitlabWebhookTokenStatus());
}
```

```javascript
if (webhookTokenPhase === "idle") {
  fetchWebhookTokenStatus();
}
```

When `GitLabProjects` is mounted inside `BasePage`, both passive effects can run
from that initial render. The first dispatch changes the store to `loading`,
but the second effect still uses its captured `idle` value. The thunk itself
does not re-check current state before calling
`client.getGitlabWebhookToken()`, so the profile-page mount can issue two
identical requests. Request IDs prevent an older response from overwriting a
newer one, but do not deduplicate the network work.

Suggested fix — centralise the initial load in one mounted owner, or have the
thunk inspect `getState()` and return immediately while the shared request is
already loading. Add a test which renders the warning and profile component
together and asserts one status request.

tiborsimko (approved):

